PR #16, opened 2021-04-27
Create the AST type definitions.

kontheocharis:
This is the frontend-agnostic AST that we will use to perform typechecking and code emitting.
Loosely modelled after the Haskell version.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1,3 +1,4 @@
+//! Frontent-agnostic Hash abstract syntax tree type definitions.
 #![allow(dead_code)]
 
 use crate::modules::ModuleIdx;
@@ -18,12 +19,14 @@ pub struct AstNode<T> {
     pub module: ModuleIdx,
 }
 
+/// [AstNode] hashes as its inner `body` type.
 impl<T: Hash> Hash for AstNode<T> {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.body.hash(state);
     }
 }
 
+/// [AstNode] dereferences to its inner `body` type.
 impl<T> Deref for AstNode<T> {
     type Target = T;
     fn deref(&self) -> &Self::Target {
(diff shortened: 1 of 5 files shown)


src/ast.rs
```diff
@@ -0,0 +1,547 @@
+//! Frontend-agnostic Hash abstract syntax tree type definitions.
+#![allow(dead_code)]
```
feds01, line 2: This is temporary?

src/ast.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// A block.
+    Block(AstNode<Block>),
+    /// An `import` call.
+    Import(AstNode<ImportPath>),
+}
+
+/// A module.
+///
+/// Represents a parsed `.hash` file.
+#[derive(Debug, Clone)]
```
feds01, line 543: What does deriving `Debug` and `Clone` do to a particular struct?

src/main.rs
```diff
@@ -1,3 +1,6 @@
+mod ast;
```
feds01, line 1: I guess we can re-organise the sources a little later once we finish up all the stuff with parsing.

feds01 (approved):
Looks really good, should we add copyright banners to the files?

src/ast.rs
kontheocharis: Yes, cause otherwise it will complain that we aren't using any of the items.

src/ast.rs
kontheocharis: https://doc.rust-lang.org/rust-by-example/trait/derive.html

src/main.rs
kontheocharis: Yeah we probably wanna create separate crates for some stuff.

feds01 (approved):

feds01:
Looks good, just got to resolve the merge conflicts and then we can merge it into main

PR #17, opened 2021-04-28
Language Grammar

feds01:
- Implemented the language grammar using pest.

Still to do:
- [ ] Tests
- [ ] Hook it up with an actual parsing mechanism, and then pipe into AST

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,5 @@ edition = "2018"
 
 [dependencies]
 clap = "3.0.0-beta.2"
+pest="2.1"
+pest_derive="2.1"
--- a/src/pest_parser/hash.pest
+++ b/src/pest_parser/hash.pest
@@ -18,10 +18,10 @@ bool  = { truel | falsel }
 
 /// Integer literal
 digit = _{ '0'..'9' }
-int   =  { digit ~ (digit | "_")* }
+int   =  @{ digit ~ (digit | "_")* }
 
 /// Float literal
-float = {
+float = @{
     int ~ "." ~ int? ~ exp? |
     int ~ exp
 }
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -8,3 +8,5 @@ edition = "2018"
 
 [dependencies]
 clap = "3.0.0-beta.2"
+pest="2.1"
+pest_derive="2.1"
```
kontheocharis, line 12: Sorry for the OCD but can you put spaces around the =

src/pest_parser/hash.pest
```diff
@@ -0,0 +1,429 @@
...
+    tuple_type       |
+    list_type        |
+    set_type         |
+    map_type         |
+    existential_type |
+    named_type
+ }
+
+/// Import statement
+import = {"import" ~ "(" ~ string ~ ")"} // @@Improvement: in the future change this to an expression
```
kontheocharis, line 243: Why change it to an expression?

src/pest_parser/hash.pest
```diff
@@ -0,0 +1,429 @@
...
+expr_inner = {
+    literal         |
+    function_call   |
+    unary_expr      |
+    intrinsic       |
+    access_name     | // variable
+    property_access |
+    block           |
+    import
+}
```
kontheocharis, line 401: Wait so do binary expressions work like this?

src/pest_parser/hash.pest
```diff
@@ -0,0 +1,429 @@
...
+    continue_statement |
+    struct_def         |
+    enum_def           |
+    trait_def          |
+    assign             |
+    expr ) ~ ";"
+}
+
+/// Module grammar
+module = _{ SOI ~ "\n"* ~ (statement ~ "\n"+) * ~ statement? ~ EOI }
```
kontheocharis, line 428: Why separated by newlines?

kontheocharis:
Looks good overall, I think we can merge this.

src/pest_parser/hash.pest
feds01: So that it in the future we may want to support dynamic imports

src/pest_parser/hash.pest
feds01: Statements could be separated some number of newlines, I think it would be better to change this to whitespace

src/pest_parser/hash.pest
feds01: Yes they do, this "approach" circumvents the grammar from using left-recursive definitions.

Essentially, this way there is an exit case if expression parsing fails

PR #19, opened 2021-04-29
Grammar

feds01:


src/error.rs
```diff
@@ -0,0 +1,31 @@
...
+pub fn report_error(err_type: ErrorType, err_msg: String) {
+    let prefix = match err_type {
+        ErrorType::ParseError => "Failed to parse\n",
+        // ErrorType::CicrularDependency => "Failed to import",
+        ErrorType::IoError => "Failed to read file",
+        ErrorType::InternalError => "Internal Panic", // @@TODO: add an internal panic function
+    };
+
+    println!("{}: {}{}", ERR, prefix, err_msg);
+    exit(-1);
```
kontheocharis, line 30: Hmm, maybe we shouldn't take the same approach we did with haskell---maybe we shouldn't immediately exit on error, but rather signal the error? Different parts of the compiler can their own error enums (with the help of `thiserror`) and return `Result`s. And then `main` can exit if it receives an error. This way we can do things like try perform some action and not necessarily exit. What do you think?

src/main.rs
```diff
@@ -28,19 +37,45 @@ struct CompilerOptions {
...
+    #[clap(short, long)]
+    debug: bool,
 }
 
 fn main() {
     let opts: CompilerOptions = CompilerOptions::parse();
 
-    println!("Stack_size is {}", opts.stack_size);
+    // print recieved cmdargs, if debug is specified
+    if opts.debug {
```
kontheocharis, line 52: Maybe we wanna use the `log` create in conjunction with some logging lib like `pretty_env_logger`? This way we can avoid always writing `if opts.debug`, among other things.

src/main.rs
```diff
@@ -28,19 +37,45 @@ struct CompilerOptions {
...
+        println!("Stack_size is {}", opts.stack_size);
 
-    for path in opts.includes.into_iter() {
-        println!("Running with {}", path);
+        for path in opts.includes.into_iter() {
```
kontheocharis, line 55: Avoid using `into_iter()` when iterating over something because it will copy each element every time you loop through it. Use `iter()` instead to get references to the elements.

src/main.rs
```diff
@@ -28,19 +37,45 @@ struct CompilerOptions {
...
+            println!("Running with {}", path);
+        }
     }
 
     match opts.execute {
-        Some(path) => println!("Are we executing -> {}", path),
+        Some(path) => match fs::canonicalize(&path) {
+            Ok(c) => {
+                // Resolve the module path
+                let contents = fs::read_to_string(c.to_str().expect("failed to convert"))
```
kontheocharis, line 64: You don't need the whole `c.to_str().expect("failed to convert")`, `&c` will suffice.

src/main.rs
```diff
@@ -28,19 +37,45 @@ struct CompilerOptions {
...
+                    .unwrap_or_else(|e| {
+                        report_error(
+                            ErrorType::IoError,
+                            format!("Couldn't read file '{}'\n{}", path, e),
+                        );
+                        exit(-1);
```
kontheocharis, line 70: Again, see above comment.

src/main.rs
```diff
@@ -28,19 +37,45 @@ struct CompilerOptions {
...
+                    });
+
+                // parse the given module
+                let _ = HashParser::parse(Rule::module, &contents).unwrap_or_else(|e| {
```
kontheocharis, line 74: We will probably wanna change this to use our AST after the "translation" is done.

kontheocharis (approved):
Looks very good!

Cargo.toml
```diff
@@ -8,5 +8,16 @@ edition = "2018"
...
+[profile.release]
+opt-level = 3
+debug = false
+debug-assertions = false
+overflow-checks = false
+lto = true
+panic = 'unwind'
+incremental = false
+codegen-units = 16
+rpath = false
```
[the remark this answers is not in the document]
feds01: only option that is changed is to enable 'lto'

src/error.rs
feds01: Yes, I think this would be a better approach. I currently set it up to just 'mimic' the Haskell version. Of course, we should move away as much as possible from the Haskell version.

src/main.rs
feds01: Ah well, I didn't know that was a thing. Ideally, we want to set up a logger for each sub-system to enable debug in specific parts of the compiler for a much better dev experience.

src/main.rs
feds01: Indeed, this was primarily set up for testing purposes.

src/main.rs
feds01: Nice, ill take note for the future.

src/main.rs
kontheocharis: Yeah, `pretty_env_logger` supports stuff like that. https://docs.rs/env_logger/0.8.3/env_logger/

feds01:
@kontheocharis Is this ready to merge?

PR #20, opened 2021-05-01
Added standard library to sources

feds01:
- Additionally, add grammatical support for hex/bin/octal literals.
- Split the sources into separate crates for parsing, type-checking and code emission/execution.


hash-parser/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+        location: Location,
+    },
+    AstGeneration {
+        rule: Rule,
+        location: Location,
+    },
+}
+
+/// Convert a [pest::error::Error] into a [ParseError]
+impl<'a> From<pest::error::Error<Rule>> for ParseError {
```
kontheocharis, line 29: Why the lifetime parameter here? Does `cargo clippy` not say anything about it?

hash-parser/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn from(pest: pest::error::Error<Rule>) -> Self {
+        match pest.variant {
+            pest::error::ErrorVariant::ParsingError {
+                ref positives,
+                ref negatives,
```
kontheocharis, line 34: Don't use `ref`, since you want to move them into `ParseError::Parsing`

hash-parser/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+            } => {
+                // @@Incomplete: Remove when we have real error formatting.
+                println!("{}: Failed to parse:\n{}", ERR, pest);
+                ParseError::Parsing {
+                    positives: positives.clone(),
+                    negatives: negatives.clone(),
```
kontheocharis, line 40: Don't use `.clone()` because that will copy the vectors for no reason.

hash-parser/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+}
+
+/// Format trait implementation for a ParseError
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
kontheocharis, line 54: We probably want to change this eventually.

hash-parser/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+//     let prefix = match err_type {
+//         ErrorType::ParseError => "Failed to parse\n",
+//         // ErrorType::CicrularDependency => "Failed to import",
+//         ErrorType::IoError => "Failed to read file",
+//         ErrorType::InternalError => "Internal Panic", // @@TODO: add an internal panic function
+//     };
+
+//     println!("{}: {}{}", ERR, prefix, err_msg);
+//     exit(-1);
+// }
```
kontheocharis, line 69: What's this?

hash-parser/src/location.rs
```diff
@@ -0,0 +1,66 @@
+//! Frontend-agnostic token/input locations utilities and definitions.
+//
+// All rights reserved 2021 (c) The Hash Language authors
+use crate::grammar::Rule;
+use std::fmt;
+
+/// Enum representing a location of a token within the source.
+#[derive(Debug, Eq, Clone, PartialEq)]
+pub enum Location {
```
kontheocharis, line 9: You can derive `Copy` for this as well, to avoid borrowing problems.

hash-parser/src/location.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+}
+
+/// Convert a [pest::Position] into a [Location]
+impl<'a> From<pest::Position<'a>> for Location {
+    fn from(pest: pest::Position<'a>) -> Location {
+        Location::Pos(pest.pos())
+    }
+}
```
kontheocharis, line 42: Does this work if you just use the implicit lifetime `'_`?

hash-parser/src/location.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Convert a [pest::Span] into a [Location]
+impl<'a> From<pest::Span<'a>> for Location {
+    fn from(pest: pest::Span<'a>) -> Location {
+        Location::Span(pest.start(), pest.end())
+    }
+}
```
kontheocharis, line 49: See comment above.

hash-parser/src/parse.rs
```diff
@@ -0,0 +1,22 @@
...
+    let result = HashParser::parse(Rule::module, source);
+
+    match result {
+        Ok(_pairs) => {
+            // here is where we do all the ast-ification
+            let vec = Vec::new();
+            Ok(Module { contents: vec })
+        }
+        Err(err) => Err(ParseError::from(err)),
+    }
```
kontheocharis, line 21: This can be shortened to
```rust
let _pairs = HashParser::parse(Rule::module, source)?;
Ok(Module { contents: vec![] })
```

hash-parser/src/location.rs
```diff
@@ -0,0 +1,66 @@
...
+
+/// Convert a [pest::error::InputLocation] into a [Location]
+impl From<pest::error::InputLocation> for Location {
+    fn from(pest: pest::error::InputLocation) -> Location {
+        match pest {
+            pest::error::InputLocation::Pos(pos) => Location::Pos(pos), // we could just use Location::Span(pos, 1) here
+            pest::error::InputLocation::Span((start, end)) => Location::Span(start, end),
+        }
+    }
+}
```
kontheocharis, line 35: Would it be a good idea to have all `pest`-related things be under the `pest_parser` module?

kontheocharis (commented):
Looks good!

hash-parser/src/error.rs
feds01: This was the old error function, i just forgot to remove it

hash-parser/src/parse.rs
feds01: well, since we will be doing stuff with the rules when we write code to convert into AST, I think it is sensible to leave it like this.

hash-parser/src/location.rs
feds01: Do you want to make multiple parsing frontends with this approach?

kontheocharis: Never mind, that's not possible for foreign trait implementations.

kontheocharis: Well I think it is generally a good idea to keep external things separate. If we ever want to make our own parser (likely) it will be very easy to do so if the usage of `pest` is isolated in the codebase.

hash-parser/src/error.rs
feds01: Could you elaborate?

hash-parser/src/error.rs
feds01: Clippy reports a few errors before it reaches this module, so it's unkown.

hash-parser/src/error.rs
feds01: Yes, it's just the beginnings of the error reporting within the parser.

hash-parser/src/error.rs
kontheocharis: `ref` essentially gets a reference to those fields, so the type of `positives`/`negatives` will be `&Vec<Rule>`. However, since you don't want to keep the old `pest::error::Error` object, you can *move* the fields `positives` and `negatives` out of it instead of getting a reference to them. Without `ref`, the type of `positives`/`negatives` is `Vec<Rule>`, which you can immediately pass into `ParseError::Parsing` without having to call `.clone()` (thereby avoiding copying all the elements).

hash-parser/src/location.rs
feds01: Yes, I think this could be moved into pest_parser or something equivalent. I think we can do this as soon as we figure out the 'ast-ification' process.

kontheocharis (approved):

PR #21, opened 2021-05-04
Added Interactive mode to one of the backends

feds01:


hash-backend/src/interpreter/command.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// Get the type of the expression
+    // @@Incomplete: This variant should store an AstNode<Expression> so that we can use the typer to get the type from the passed expression
+    Type(String),
+    /// Just prints the version of the current interactive mode
+    Version,
+}
+
+impl InteractiveCommand {
+    /// Attempt to convert a string into an interactive command
+    pub fn from(input: &str) -> Option<InteractiveCommand> {
```
kontheocharis, line 22: Maybe call this `new`?

hash-backend/src/interpreter/command.rs
```diff
@@ -0,0 +1,69 @@
...
+            _ => {
+                if input.len() > 3 {
+                    let (command, arg) = input.split_at(3);
+                    // check here if it begins with either a ':t' or a ':p' and then take an expression
+                    if command == ":t " {
+                        return Some(InteractiveCommand::Type(String::from(arg)));
+                    }
+                }
+
+                None
```
kontheocharis, line 36: Can be simplified to
```rust
_ => input.strip_prefix(":t ").map(|arg| InteractiveCommand::Type(arg.to_owned()))
```

Also, is `:p` in the comment `check here if it begins with either a ':t' or a ':p' and then take an expression` inaccurate?

hash-backend/src/interpreter/command.rs
```diff
@@ -0,0 +1,69 @@
...
+                &format!("Command '{}' does not take any arguments.", command),
+            ),
+            ":t" => {
+                match args.next() {
+                    Some(_) => unreachable!(), // if it did include a whitespace prior, it should of been caught in InteractiveCommand::from
+                    None => report_interp_error(
+                        InterpreterError::ArgumentError,
+                        &format!("Command '{}' requires one argument.", command),
+                    ),
+                }
```
kontheocharis, line 61: Shouldn't this happen in the function above, rather than here? That way `report_error` is not really needed at all.

hash-backend/src/interpreter/command.rs
kontheocharis: I think it is better to do all the command checking in here and return some sort of `Result` type?

hash-backend/src/interpreter/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+            Ok(line) => {
+                rl.add_history_entry(line.as_str());
+                execute(line.as_str());
+            }
+            Err(ReadlineError::Interrupted) | Err(ReadlineError::Eof) => {
+                println!("Exiting!");
+                break;
+            }
+            Err(err) => {
+                println!("Error: {:?}", err);
```
kontheocharis, line 46: Should we not use `report_interp_error` here?

hash-backend/src/interpreter/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+fn execute(input: &str) {
+    if input.is_empty() {
+        return;
+    }
+
+    // check here if it is an 'interactive command', as in prefixed with a ':' and then some string
+    if let Some(k) = InteractiveCommand::from(&input) {
+        match k {
+            InteractiveCommand::Quit => {
+                println!("Goodbye!");
```
kontheocharis, line 63: We probably want to factor this out in order to use it on line 42 as well.

hash-parser/src/parse.rs
```diff
@@ -13,10 +13,37 @@ pub fn module(source: &str) -> Result<Module, ParseError> {
...
             Ok(Module { contents: vec })
         }
         Err(err) => Err(ParseError::from(err)),
     }
 }
+
+/// Function to parse an individual statement. This function is primarily used for the interactive
+/// mode where only statements are accepted.
+pub fn statement(source: &str) -> Result<AstNode<Statement>, ParseError> {
+    let result = HashParser::parse(Rule::statement, source);
```
kontheocharis, line 27: You can use the question mark operator `let result = HashParser::parse(Rule::statement, source)?;` to prevent having to match.

kontheocharis (commented):
Looks good!

hash-backend/src/interpreter/command.rs
feds01: Oh and then return an `InterpreterError` if it fails, then we could get rid of the `report_error` function?

hash/Cargo.toml
```diff
@@ -10,3 +10,5 @@ clap = "3.0.0-beta.2"
 
 hash-parser = {path = "../hash-parser" } # @@TODO: remove this when the typchecker is ready since it will be the one who uses this!
```
kontheocharis, line 11: Is this accurate?

kontheocharis (approved):
Nice!

hash/Cargo.toml
feds01: I thought this would be the case since we can call the hash-transform crate to get IR and then direct it to whatever backend

kontheocharis: Yeah but surely we have to provide `hash-transform` with the AST right? It won't call the parser by itself.

feds01: Oh yes, it won't call the parser by itself, the frontend will call transform and then that calls parser, ast, type check, etc.

PR #22, opened 2021-05-05
Modify grammar to make it easier to parse into AST

kontheocharis:
Also add some scaffolding for AST translation.

### DIFF
--- a/hash-parser/src/emit.rs
+++ b/hash-parser/src/emit.rs
@@ -52,11 +52,7 @@ impl ast::IntoAstNode<ast::AccessName> for HashPair<'_> {
             Rule::access_name => AstBuilder::from_pair(&self).node(ast::AccessName {
                 names: self
                     .into_inner()
-                    .map(|p| {
-                        AstBuilder::from_pair(&p).node(ast::Name {
-                            string: p.as_str().to_owned(),
-                        })
-                    })
+                    .map(|p| p.into_ast())
                     .collect(),
             }),
             _ => unreachable!(),
@@ -78,7 +74,7 @@ fn single_access_name(ab: &AstBuilder, name: &str) -> ast::AstNode<ast::AccessNa
 impl ast::IntoAstNode<ast::Type> for HashPair<'_> {
     fn into_ast(self) -> ast::AstNode<ast::Type> {
         match self.as_rule() {
-            Rule::type_t => {
+            Rule::any_type => {
                 let ab = AstBuilder::from_pair(&self);
                 let in_type = self.into_inner().next().unwrap();
                 match in_type.as_rule() {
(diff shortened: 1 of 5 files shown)


hash-parser/src/hash.pest
```diff
@@ -2,443 +2,453 @@
...
+geq_op = { ">=" }
+leq_op = { "<=" }
+andb_op = { "&" }
+orb_op = { "|" }
+xorb_op = { "^" }
+binary_op = { 
+    triple_eq_op  |
+    double_eq_op  |
+    neq_op        |
+    double_neq_op |
```
feds01, line 211: Ah we didn't think of this before, maybe we should update the hashdocs to list all the operators so we don't forget?

hash-parser/src/hash.pest
```diff
@@ -2,443 +2,453 @@
...
+/// ---- Blocks ----
 
-/// Body block grammar, body blocks can be used within any kind
-/// of expression to denote a body of a particular expression.
-/// The expression at the end is used as a implicit 'return' statement
-/// from the given body block.
-body_block = { "{" ~  statement* ~ expr? ~ "}" }
+/// Body block
+/// Zero or one statements followed by zero or one expression.
+body_block = { curly_left ~ statement* ~ expr? ~ curly_right } // @Broken: statement parsing needs to be non-greedy
```
feds01, line 385: Yes I was hoping we could do the transformation at AST level since pest is by nature greedy

hash-parser/src/hash.pest
```diff
@@ -2,443 +2,453 @@
...
+single_expr = { subject_expr ~ (property_access | fn_args | index_arg)* }
+
+/// Unary expression
+unary_expr = { unary_op ~ unary_expr | single_expr }
+
+/// Typed expression (as)
+typed_expr = { unary_expr ~ ( as_kw ~ any_type )? }
+
+/// Binary expression
+binary_expr = { typed_expr ~ ( binary_op ~ typed_expr )* }
```
feds01, line 329: From a first glance this looks confusing, but essentially this creates precedence using pest's rules.

hash-parser/src/emit.rs
```diff
@@ -0,0 +1,191 @@
...
+fn single_access_name(ab: &AstBuilder, name: &str) -> ast::AstNode<ast::AccessName> {
+    ab.node(ast::AccessName {
+        names: vec![ab.node(ast::Name {
+            string: name.to_owned(),
+        })],
+    })
+}
+
+impl ast::IntoAstNode<ast::Type> for HashPair<'_> {
+    fn into_ast(self) -> ast::AstNode<ast::Type> {
```
feds01, line 75: I guess we will need to add implementations for each `ast::*` ?

feds01 (approved):
Looks pretty good, with some big re-organisation effort in the grammar file. nice!

hash-parser/src/hash.pest
kontheocharis: Well this doesn't actually handle precedence, I figured we can use `PrecClimber` for that.

hash-parser/src/emit.rs
kontheocharis: Yes indeed.

hash-parser/src/hash.pest
kontheocharis: Yeah, there's probably no other way to do it from within pest.

hash-parser/src/hash.pest
kontheocharis: But yes, it does handle precedence with `as` and other binary expressions.

PR #32, opened 2021-05-08
Parsing

feds01:
This patch establishes parsing within the compiler, including the ability to support future custom 'parsing backends'. 

- Grammar was finalised and tested using the `language_manifest.hash` file
- Module discovery is implemented and working
- Parallel and sequential parsing
- Full AST generation from the `pest` grammar


hash-parser/src/modules.rs
```diff
@@ -30,9 +45,134 @@ pub struct Modules {
...
 impl Modules {
     /// Get the module at the given index.
     pub fn get_module(&self, idx: ModuleIdx) -> Module<'_> {
         Module { idx, modules: self }
     }
+
+    /// Function that builds a module map of the standard library that is shipped
+    /// with the compiler distribution. Standard library modules are referenced
+    /// within imports
+    pub fn get_stdlib_modules(&self, dir: impl AsRef<Path>) -> Vec<PathBuf> {
```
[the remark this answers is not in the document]
feds01: I thought it should panic since if we aren't able to read the standard library folder, it's in a pretty bad state, and we're likely unable to continue in the pipeline.

hash-parser/src/modules.rs
```diff
@@ -30,9 +45,134 @@ pub struct Modules {
...
+        }
+
+        // otherwise, we have to resolve the module path based on the working directory
+        let work_dir = wd.canonicalize().unwrap();
+        let mut raw_path = work_dir.join(path);
+
+        // check if that path exists, if not it does return it as an error
+        if !raw_path.exists() {
+            // @@Copied
+            let path = String::from(path.to_str().unwrap());
```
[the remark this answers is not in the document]
feds01: Yes, it should be since this is the path that is given when parsing a Import statement.

hash-parser/src/modules.rs
```diff
@@ -30,9 +45,134 @@ pub struct Modules {
...
+                return Ok(raw_path);
+            }
+
+            // @@Copied
+            let path = String::from(path.to_str().unwrap());
+            Err(ParseError::ImportError {
+                import_name: path,
+                location,
+            })
+        } else {
```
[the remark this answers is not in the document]
feds01: well the same route follows, I mean if it was a symbolic link, `is_dir` catches this case and we attempt to follow the same logical route.

hash-parser/src/ast.rs
```diff
@@ -540,7 +540,7 @@ pub enum Expression {
     /// A property access.
     PropertyAccess(PropertyAccessExpr),
     /// A literal.
-    LiteralExpr(Literal),
+    LiteralExpr(AstNode<Literal>),
```
kontheocharis, line 543: Why should there be another AstNode here?

hash-parser/src/emit.rs
```diff
@@ -156,8 +156,8 @@ impl ast::IntoAstNode<ast::Type> for HashPair<'_> {
     }
 }
 
-impl ast::IntoAstNode<ast::Literal> for HashPair<'_> {
-    fn into_ast(self) -> ast::AstNode<ast::Literal> {
+impl IntoAstNode<Literal> for HashPair<'_> {
+    fn into_ast(self) -> AstNode<Literal> {
         match self.as_rule() {
             Rule::literal_expr => {
                 let ab = AstBuilder::from_pair(&self);
```
kontheocharis, line 163: Did we decide on whether we should have non-float exponents? My vote would be no.

hash-parser/src/emit.rs
```diff
@@ -95,23 +101,287 @@ impl ast::IntoAstNode<ast::Type> for HashPair<'_> {
...
+                                ab.node(Literal::Int(val))
+                            }
+                            Rule::octal_literal => {
+                                let val =
+                                    BigInt::parse_bytes(inner.as_str().as_bytes(), 8).unwrap();
+                                ab.node(Literal::Int(val))
+                            }
+                            Rule::bin_literal => {
+                                let val =
+                                    BigInt::parse_bytes(inner.as_str().as_bytes(), 2).unwrap();
```
kontheocharis, line 199: Do we always unwrap for all BigInt parsing? Are we always sure it will succeed?

hash-parser/src/emit.rs
```diff
@@ -95,23 +101,287 @@ impl ast::IntoAstNode<ast::Type> for HashPair<'_> {
...
+impl IntoAstNode<Expression> for HashPair<'_> {
+    fn into_ast(self) -> AstNode<Expression> {
+        match self.as_rule() {
+            Rule::expr => {
+                let ab = AstBuilder::from_pair(&self);
+                let expr = self.into_inner().next().unwrap();
+
+                match expr.as_rule() {
+                    Rule::block => ab.node(Expression::Block(expr.into_ast())),
+                    Rule::struct_literal => ab.node(Expression::LiteralExpr(expr.into_ast())),
```
kontheocharis, line 327: Could do `ab.node(Expression::LiteralExpr(expr.into_ast().body))` to avoid extra AstNode, but I think this is fine as well.

feds01 (commented):
> Did we decide on whether we should have non-float exponents? My vote would be no.

I think that the way it is done now is not in the best way. What I would like to see is that float fractional parts should be optional. However, this would create an ambiguity between integer and float tokenisation.

kontheocharis:
> > Did we decide on whether we should have non-float exponents? My vote would be no.
> 
> I think that the way it is done now is not in the best way. What I would like to see is that float fractional parts should be optional. However, this would create an ambiguity between integer and float tokenisation.

@feds01  I think we can implement that eventually, it will just require some black magic during typechecking.

kontheocharis:
> What is the alternative if it does succeed, error report?

@feds01  We probably want `into_ast` to return some sort of `Result` instead of just a plain AstNode. And yeah, we'd report an error.

feds01:
> > What is the alternative if it does succeed, error report?
> 
> @feds01 We probably want `into_ast` to return some sort of `Result` instead of just a plain AstNode. And yeah, we'd report an error.

this will be useful since we might want to do error reporting of this kind as early as possible.

hash-parser/src/emit.rs
feds01: yep i removed this ability

PR #41, opened 2021-07-01
Visualising AST

feds01:
This PR essentially brings an AST visualiser that can display AST nodes into a tree format. This visualiser can be invoked via the interactive console by specifying the command `:d` and then visualising the following expression.

### DIFF
--- a/compiler/hash-ast/src/count.rs
+++ b/compiler/hash-ast/src/count.rs
@@ -158,7 +158,7 @@ impl NodeCount for Pattern {
             Pattern::Namespace(pat) => pat.patterns.iter().map(|p| p.node_count()).sum(),
             Pattern::Tuple(pat) => pat.elements.iter().map(|e| e.node_count()).sum(),
             Pattern::Literal(_) => 1,
-            Pattern::Or(pat) => pat.a.node_count() + pat.b.node_count(),
+            Pattern::Or(pat) => pat.variants.iter().map(|e| e.node_count()).sum(),
             Pattern::If(pat) => {
                 let count = pat.pattern.node_count();
                 count + pat.condition.node_count()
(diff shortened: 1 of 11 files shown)


compiler/hash-ast/src/visualise.rs
```diff
@@ -65,7 +121,7 @@ where
         write!(
             f,
             "{}",
-            self.node_display(0)
+            self.node_display()
                 .into_iter()
                 .map(|mut line| {
```
kontheocharis, line 126: `self.node_display().join("\n")`

compiler/hash-ast/src/visualise.rs
```diff
@@ -131,133 +223,676 @@ impl NodeDisplay for Literal {
...
-                lines.push("ret".to_string());
-
-                // if a return statement has a lin
-                if let Some(ret_expr) = expr {
-                    next_lines.push(format!(
-                        "{}{}",
-                        END_PIPE,
-                        ret_expr.node_display(next_indent).join("\n")
-                    ));
+        let child_lines: Vec<String> = match &self {
```
kontheocharis, line 394: Return tuple of `(node_name, child_lines)` from each branch, instead of having two separate `match`es

compiler/hash-ast/src/count.rs
```diff
@@ -164,7 +158,7 @@ impl NodeCount for Pattern {
             Pattern::Namespace(pat) => pat.patterns.iter().map(|p| p.node_count()).sum(),
             Pattern::Tuple(pat) => pat.elements.iter().map(|e| e.node_count()).sum(),
             Pattern::Literal(_) => 1,
```
kontheocharis, line 160: Use 0 instead of 1

compiler/hash-ast/src/parse.rs
```diff
@@ -328,6 +328,9 @@ fn parse_file(
         |elapsed| debug!("ast: {:.2?}", elapsed),
     )?;
 
+    // @@Temp: remove me
+    println!("{}", module);
```
kontheocharis, line 332: Remove it?

kontheocharis (approved):

Issue #46, opened 2021-07-05
Patterns are uneccesarilly wrapped in `Or` pattern variant.


feds01:
The given expression `let (x, y) = get_coords()` should yield a `tuple` pattern with two binds for `x` and `y`. However, it currently produces binds that are wrapped in an `Or` variant, which can be visualised like so:

```
>>> :d let (x, y) = get_coords();
└─let
  ├─pattern
  │ └─tuple
  │   ├─pattern
  │   │ └─or
  │   │   └─pattern
  │   │     └─bind "x"
  │   └─pattern
  │     └─or
  │       └─pattern
  │         └─bind "y"
  └─value
    └─function_call
      └─subject
        └─ident "get_coords"
```

PR #47, opened 2021-07-05
Fix '#46'

feds01:


compiler/hash-pest-parser/src/translate.rs
```diff
@@ -907,9 +907,25 @@ where
...
+                // variant since this is redundant
+                match pairs.next() {
+                    None => Ok(first),
+                    Some(pat) => {
+                        let second = self.transform_pattern(pat)?;
+
+                        // collect any remaining patterns in the or secquence
+                        let variants = vec![first, second]
+                            .into_iter()
+                            .chain(ab.try_collect(pairs.map(|p| self.transform_pattern(p)))?)
```
kontheocharis, line 923: Do you need to `try_collect` here?

kontheocharis (approved):
Looks good

compiler/hash-pest-parser/src/translate.rs
feds01: Yes, since the patterns after the first and second are still to be transformed and there may be many patterns. I think the best way is to use a `try_collect`, unless there is a better way to unwrap a `Vec` of `ParseResult` into `AstNode<Pattern>` with the `chain` call?

Issue #44, opened 2021-07-04
Move errors to hash-reporting crate.


kontheocharis:
Remove all the dump errors from being all over the place.

PR #48, opened 2021-07-05
Re organise crates

feds01:
- This moves all of the related code to interactive (REPL) into its own crate `hash-interactive`
- Removes the un-used `hash-backend` crate which will later be renamed and split into multiple crates.
- Moves CompilerError and InteractiveCommandError into `hash-reporting` crate. (#44)


compiler/hash/src/main.rs
```diff
@@ -125,7 +124,7 @@ fn main() {
                 Ok(())
             }
             None => {
-                interactive::init()?;
+                init()?;
```
kontheocharis, line 127: Maybe namespace it?

kontheocharis (approved):

PR #49, opened 2021-07-06
Use identifier caching to prevent overhead from string operations.

kontheocharis:
Before, each identifier in the AST (like in Name and AccessName) was a string (Cow, granted), which wasn't ideal. This PR makes the following change:

Identifiers are kept track of using a static map structure that assigns IDs to each string identifier. Then, these IDs can be used in the AST nodes instead of strings, to represent identifiers. This will also be useful during program analysis, when checking for identifier equality.

Furthermore, this PR removes type_id from the AstNode struct and moves it to Expression, which is the only node that actually requires it.


compiler/hash-pest-parser/src/translate.rs
```diff
@@ -147,15 +141,17 @@ impl NodeBuilder {
...
-                type_args: vec![],
-            })),
-            args: self.node(FunctionCallArgs {
-                entries: vec![lhs, rhs],
-            }),
-        }));
+        let fn_call = self.node(Expression::new(ExpressionKind::FunctionCall(
+            FunctionCallExpr {
+                subject: self.node(Expression::new(ExpressionKind::Variable(VariableExpr {
+                    name: self.make_single_access_name(AstString::Borrowed("ord")),
```
feds01, line 147: In the future, we should put common symbols into the hashmap before we run to avoid that additional first time adding cost.

compiler/hash-pest-parser/src/translate.rs
```diff
@@ -1007,26 +1012,31 @@ where
...
-                            Deref => {
-                                Ok(ab.node(Expression::Deref(self.transform_expression(operand)?)))
+                                Ok(ab.node(Expression::new(ExpressionKind::FunctionCall(
+                                    FunctionCallExpr {
+                                        subject: ab.node(Expression::new(
+                                            ExpressionKind::Variable(VariableExpr {
+                                                name: ab.make_single_access_name(
+                                                    AstString::Borrowed(fn_call),
+                                                ),
+                                                // name: ab.node(AccessName {
```
feds01, line 1022: Should this comment still be here?

feds01 (approved):
Looks good to me!

compiler/hash-pest-parser/src/translate.rs
kontheocharis: Yes, probably a file alongside `ident.rs` that injects them at startup.

feds01 (approved):

PR #50, opened 2021-07-06
Added a compiler panic handler with formatting and backtracing.

feds01:
This pull request adds a `Haskell` compatible internal panic display message which collects useful information about the location and nature of the internal panic.

Also, this PR adds some dummy files that are in preparation for the testing sub-system in `hash-ast`.


kontheocharis (approved):
LGTM

PR #51, opened 2021-07-09
Integrated logging via the '-d/--debug' cmdline options.

feds01:
The library 'pretty-env-logger' doesn't support variable logging levels during runtime. To workaround
this limitation, we now implement a `log::Log` which does allow for this behaviour. The logger is initialised
in the same way as before, but if 'debug' is specified via cmdline args, the level is set to `debug`.


compiler/hash-ast/src/parse.rs
```diff
@@ -67,6 +67,10 @@ where
     }
 
     pub fn new_with_workers(backend: B, worker_count: usize) -> Self {
+        if worker_count == 0 {
+            panic!("Cannot spawn a parallel parser with 0 threads");
+        }
```
kontheocharis, line 72: What if we use `NonZeroUsize`?

kontheocharis (approved):

PR #54, opened 2021-07-15
Refactor the parser

kontheocharis:
The previous parsing approach was to have a thread pool and a channel. As module imports were encountered jobs were spawned on the pool, which communicated back to the main thread using the channel (to add module info into the Modules map). However, this proved to be very cumbersome from a code quality perspective, and also introduced the unnecessary overhead of having a channel.

The new approach makes `Modules` concurrent (using `dashmap`) so that it can be used by all the worker threads. Furthermore, each worker thread is now responsible for spawning child threads (on nested imports). This removes the need for a channel.
Error handling occurs through a concurrent queue of Errors which is passed into each `ParModuleResolver`. Once an error is encountered, it is added to the queue. At the end of parsing, all errors should be returned back to the caller.


compiler/hash-ast/src/error.rs
```diff
@@ -33,3 +33,15 @@ impl From<(io::Error, PathBuf)> for ParseError {
...
+    fn from(_: ParseIntError) -> Self {
+        todo!()
+    }
+}
+
+impl From<ParseFloatError> for ParseError {
+    fn from(_: ParseFloatError) -> Self {
+        todo!()
+    }
+}
```
feds01, line 47: So I guess here we need a location and source of the file?

compiler/hash-ast/src/resolve.rs
```diff
@@ -0,0 +1,122 @@
...
+pub trait ModuleResolver {
+    fn module_source(&self) -> Option<&str>;
+    fn add_module(
+        &mut self,
+        import_path: impl AsRef<Path>,
+        location: Option<SourceLocation>,
+    ) -> ParseResult<ModuleIdx>;
+}
+
+#[derive(Debug, Copy, Clone, Constructor, Getters)]
```
feds01, line 28: What is `Getters`, are they just derived functions to get a specific field, why is it better over just getting the field?

compiler/hash-pest-parser/src/backend.rs
```diff
@@ -0,0 +1,66 @@
...
+    parse::ParserBackend,
+    resolve::ModuleResolver,
+};
+use hash_utils::timed;
+use pest::Parser;
+use std::path::Path;
+
+pub struct PestBackend;
+
+impl ParserBackend for PestBackend {
```
feds01, line 22: I am having to do the same pattern here for within `hash-parser` on the [tokeniser](https://github.com/hash-org/lang/tree/tokeniser) branch. Is there a way of getting rid of the `impl ParserBackend for ...` as it seems to be redundant.

compiler/hash-pest-parser/src/translate.rs
```diff
@@ -586,23 +587,22 @@ where
...
                         let item = inner.into_inner().next().unwrap();
-                        let val = BigInt::parse_bytes(item.as_str().as_bytes(), 8).unwrap();
+                        let val = u64::from_str_radix(item.as_str(), 8)?;
                         Ok(ab.node(Literal::Int(val)))
                     }
                     Rule::bin_literal => {
                         let item = inner.into_inner().next().unwrap();
-                        let val = BigInt::parse_bytes(item.as_str().as_bytes(), 2).unwrap();
+                        let val = u64::from_str_radix(item.as_str(), 2)?;
                         Ok(ab.node(Literal::Int(val)))
```
feds01, line 606: Once we merge the `tokeniser` stuff and decide on a formal spec for float and int literals, we should move away from relying on letting Rust parse these for us. Since we want to control the behaviour and not let it be guided by an external format.

feds01 (commented):
Looks really good, and definitely a huge clean up of the previous parsing stuff.

compiler/hash-ast/src/error.rs
kontheocharis: Wtf, can you believe that I didn't actually add these intentionally... I must have accidentally pressed some language server code action.

compiler/hash-ast/src/resolve.rs
kontheocharis: Yes. Better because when dealing with value types, they prevent you from modifying the field, and allow only read access through a reference. Technically not necessary, but without them I'd also have to make the struct fields public.

compiler/hash-pest-parser/src/backend.rs
kontheocharis: What do you mean? Which part is redundant?

compiler/hash-pest-parser/src/translate.rs
kontheocharis: Agreed.

compiler/hash-pest-parser/src/backend.rs
feds01: Since `PestBackend` and `HashBackend` carry no data on them, is there some way to just have `impl ParserBackend` for the crate. Then, by the magic of compile-time feature flags, we can just get the relevant crate and get the implementation we want?

compiler/hash-ast/src/error.rs
feds01: I see, I guess that's why they teach you to code review at computer school.

compiler/hash-ast/src/resolve.rs
feds01: Well, it seems to me that one way or another, they'll have to be 'public'. I just feel like it's unnecessary bloat that Java introduces to create 'encapsulation'.

compiler/hash-ast/src/error.rs
kontheocharis: Yep 😅

compiler/hash-ast/src/resolve.rs
kontheocharis: To be a bit pragmatic, I actually think there is some merit to this 'encapsulation' pattern, since it restricts the API of the type, which if not done might lead to unwanted consequences elsewhere in the codebase. However, since these types are localised to this crate, I think they would be fine just being plain structs.

compiler/hash-pest-parser/src/backend.rs
kontheocharis: Well it doesn't work like that. Traits have to be implemented by types, there are no "free traits" like there are free functions. So you need a unit struct to implement the trait. This isn't considered an anti-pattern, and is used pretty commonly AFAIK.

compiler/hash-ast/src/error.rs
kontheocharis: Never mind, it turns out that these are necessary to implement the modified int literal parsing. I think you must have added them?

compiler/hash-pest-parser/src/backend.rs
feds01: That's fair enough, I guess the unit struct won't even mean anything at runtime.

feds01 (approved):
I think once all of this is documented, we can merge it in!

PR #55, opened 2021-07-21
Implementation for the `hash-parser` Tokeniser

feds01:
## Overview
Implemented a fully-fledged lexer that can handle the whole range of hash source file inputs.

The process that `hash-parser` utilises is that it will first convert the input source into a stream of `Token`s, ' which can then be transformed into `AstNode`s. This is a more efficient process to using grammar because there is no notion of `recursive decent` and thus parsing rules are generally far simpler. This is also beneficial because the source file is more compactly represented in memory (in comparison to the `pest` alternative). This saves on resources and time, due to memory allocations taking some time to occur. Additionally, we can hook in the `Wall` allocator to even further benefit from fine-tuned memory management because this is written by us.

Furthermore, during the lexing process, we are able to capture string literals and identifiers, placing them in the `IDENTIFIER_MAP` and the corresponding string literal map. This means that we don't have to waste time re-allocating components when working with just tokens instead of having to allocate twice with `pest`.

Without the allocator, there are some very promising results:
```
examples/lang_manifest.hash (pest) (599.447µs)
examples/lang_manifest.hash (self) (172.642µs)
examples/ignored/100k_printing.hash (pest) (257.431599ms)
examples/ignored/100k_printing.hash (self) (78.939833ms)
```

### Quality of life improvements:
- Panic handler is now thread-safe and collects info about which thread panics.
- Panic handler can now handle messages that have `Arguments` (from `std::fmt`), however, this adds *another nightly flag*.

### House Keeping:
- Added a `.mailmap` to avoid bogus contributor statistics and attributions.
- Added a special rule to `cargo fmt` so that `single-line-if-else` is allowed to be  the same length as `max-width`
- Compiling the project is now possible with two parsing backends. It can be specified which backend to use (by default it is `hash-parser`).


compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+use crate::token::TokenErrorKind;
+use crate::token::TokenKind;
+use crate::token::TokenResult;
+use crate::utils::*;
+use std::cell::Cell;
+use std::iter;
+
+/// Representing the end of stream, or the initial character that is set as 'prev' in
+/// a [Lexer] since there is no character before the start.
+const EOF_CHAR: char = '\0';
```
kontheocharis, line 21: What if there is a null byte in the input stream?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+
+    /// Peeks the second symbol from the input stream without consuming it.
+    pub(crate) fn peek_second(&self) -> char {
+        self.nth_char(1)
+    }
+
+    // /// Returns nth character relative to the current position.
+    // /// If requested position doesn't exist, `EOF_CHAR` is returned.
+    // /// However, getting `EOF_CHAR` doesn't always mean actual end of file,
+    // /// it should be checked with `is_eof` method.
```
kontheocharis, line 67: Double commenting

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    /// Moves to the next character.
+    pub(crate) fn next(&self) -> Option<char> {
+        let offset = self.offset.get();
+
+        // @@Safety: We rely that the byte offset is correctly computed when stepping over the
+        //           characters in the iterator.
+        let slice = unsafe { std::str::from_utf8_unchecked(&self.contents.as_bytes()[offset..]) };
+        let ch = slice.chars().next()?;
```
kontheocharis, line 85: Shouldn't prev be set to EOF if this returns None?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+
+        // only increment the offset by one if there is a next character
+        self.prev.set(ch);
+
+        self.offset.set(self.offset.get() + ch.len_utf8());
```
kontheocharis, line 90: No need to get again, there is the local `offset`

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+        Some(ch)
+    }
+
+    /// Checks if there is nothing more to consume.
+    pub(crate) fn is_eof(&self) -> bool {
+        self.contents.len() == self.len_consumed()
```
kontheocharis, line 96: Will offset ever reach `self.contents.len()`? Won't it always be `<= self.contents.len() - 1`?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    /// Parses a token from the input string.
+    pub(crate) fn advance_token(&self) -> Option<Token> {
+        let offset = self.offset.get();
+        let token_kind = loop {
```
kontheocharis, line 102: This is only a loop because of the whitespace?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+
+                // Identifier (this should be checked after other variant that can
+                // start as identifier).
+                c if is_id_start(c) => break self.ident(),
+
+                // Numeric literal.
+                '0'..='9' => {
+                    break self
+                        .number()
+                        .unwrap_or_else(|e| panic!("error: {:#?}", e.message))
```
kontheocharis, line 165: TODO: error handling

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                // character literal.
+                '\'' => {
+                    break self
+                        .char()
+                        .unwrap_or_else(|e| panic!("error: {:#?}", e.message)); // @@ErrorReporting: this is where we hook into error reporting to print the result
+                                                                                // and display it to the user
+                }
+
+                // String literal.
+                '"' => break self.string().unwrap_or_else(|e| panic!("error: {:?}", e)),
```
kontheocharis, line 176: What about character literals?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+
+                _ => break TokenKind::Unexpected,
```
kontheocharis, line 178: There should be an unexpected token kind?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+            }
+        };
+
+        let location = Location::span(offset, self.len_consumed());
+        Some(Token::new(token_kind, location))
+    }
+
+    /// Consume an identifier, at this stage keywords are also considered to be identfiers
+    pub(crate) fn ident(&self) -> TokenKind {
```
kontheocharis, line 187: Maybe it should be documented that this expects the first character to have already been consumed?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+        let suffix = self.eat_while(is_id_continue);
+        let name: String = iter::once(first).chain(suffix).collect();
+
+        // create the identifier here from the created map
+        let ident = IDENTIFIER_MAP.create_ident(AstString::Owned(name.as_str().to_owned()));
+        TokenKind::Ident(ident)
+    }
+
+    /// Consume a number literal, either float or integer
+    pub(crate) fn number(&self) -> TokenResult<TokenKind> {
```
kontheocharis, line 200: See ident comment

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+        // record the start location of the literal
+        let start = self.offset.get() - 1;
+
+        // firstly, figure out if this literal has a base, if so then we need to perform
+        // some magic at the end to cast it into the correct base...
+        if prev == '0' {
+            let maybe_radix = match self.peek() {
+                'b' => Some(2),
+                'o' => Some(8),
+                'x' => Some(16),
```
kontheocharis, line 213: What about `'d'` for decimal (which should be the default as well)?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+            // if this does have a radix then we need to handle the radix
+            if let Some(radix) = maybe_radix {
+                self.next(); // accounting for the radix
+
+                let chars = self.eat_while(|c| c.is_digit(radix)).collect::<String>();
+                let value = u64::from_str_radix(chars.as_str(), radix);
+
+                // @@ErrorHandling: We shouldn't error here, this should be handeled by the SmallVec<..> change to integers
+                if value.is_err() {
+                    println!("value={}, radix={}", chars, radix);
```
kontheocharis, line 226: Should be turned to debug?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                    return Err(TokenError::new(
+                        Some("Integer literal too large".to_string()),
+                        TokenErrorKind::MalformedNumericalLiteral,
+                        Location::span(start, self.offset.get()),
+                    ));
+                }
+
+                return Ok(TokenKind::IntLiteral(value.unwrap()));
```
kontheocharis, line 234: Use match to avoid unwrapping.

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                'x' => Some(16),
+                _ => None,
+            };
+
+            // if this does have a radix then we need to handle the radix
+            if let Some(radix) = maybe_radix {
+                self.next(); // accounting for the radix
+
+                let chars = self.eat_while(|c| c.is_digit(radix)).collect::<String>();
+                let value = u64::from_str_radix(chars.as_str(), radix);
```
kontheocharis, line 222: What about separators? How come the logic below isn't used for this?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+        }
+
+        let pre_digits = iter::once(self.prev.get())
+            .chain(self.eat_decimal_digits())
+            .collect::<String>();
+
+        // peek next to check if this is an actual float literal...
+        match self.peek() {
+            '.' => {
+                // Only eat the charach
```
kontheocharis, line 245: charach

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                self.next();
+
+                let after_digits = self.eat_decimal_digits().collect::<String>();
```
kontheocharis, line 248: Is it necessary to collect here?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                }
+
+                // if an exponent was speified, as in it is non-zero, we need to apply the exponent to
+                // the float literal.
+                let value = if exp != 0 { num.unwrap() * 10f64.powi(exp) } else { num.unwrap() };
+
+                Ok(TokenKind::FloatLiteral(value))
+            }
+            // Imediate exponent
+            'e' | 'E' => {
```
kontheocharis, line 278: Repeated logic for exponents, maybe there is a way to factor it out?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+
+    /// Transform an ordinary character into a well known escape sequence specified by the
+    /// escape literal rules. More information about the escape sequences can be found at
+    /// [escape sequences](https://hash-org.github.io/lang/basics/intro.html)
+    fn char_from_escape_seq(&self) -> TokenResult<char> {
+        debug_assert!(self.prev.get() == '\\');
+
+        // @@Incomplete: come up with a better algorithm to transform escaped literals, rather than manual
+        // transformations!
```
kontheocharis, line 363: What do you have in mind?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                        Some("Expected '{' after a '\\u' escape sequence".to_string()),
+                        TokenErrorKind::BadEscapeSequence,
+                        Location::span(start, self.offset.get()),
+                    ));
+                }
+
+                self.next();
+
+                // here we expect up to 6 hex digits, which is finally closed by a '}'
+                let chars: String = self.eat_while(|c| c.is_ascii_hexdigit()).collect();
```
kontheocharis, line 387: Is it necessary to collect here? Maybe could use the slice from the input?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                            Some("ASCII escape code must only contain hex digits".to_string()),
+                            TokenErrorKind::Unexpected(c),
+                            Location::span(start, self.offset.get()),
+                        )),
+                    })
+                    .collect();
+
+                // check that getting the 2 characters was ok, if not then propagate the error
+                if let Err(e) = chars {
+                    return Err(e);
```
kontheocharis, line 431: Could use `?`

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                    let ch = self.char_from_escape_seq()?;
+                    value.push(ch);
+                }
+                ch => value.push(ch),
+            }
+        }
+
+        // Essentially we put the string into the literal map and get an id out which we use for the
+        // actual representation in the token
+        let id = STRING_LITERAL_MAP.create_string(AstString::Owned(value.as_str().to_owned()));
```
kontheocharis, line 530: Don't have to use as_str and then to_owned, value is already owned.

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,618 @@
...
+                return self.next();
+            }
+
+            None
+        })
+    }
+}
+
+/// Function to tokenise an input string. Resulting in an iterator of [Token]s
+pub fn tokenise(input: &str) -> impl Iterator<Item = Token> + '_ {
```
kontheocharis, line 608: Don't need `'_`

compiler/hash/src/crash_handler.rs
```diff
@@ -0,0 +1,60 @@
...
+// All rights reserved 2021 (c) The Hash Language authors
+
+use backtrace::Backtrace;
+use std::panic::PanicInfo;
+use std::process::exit;
+use std::{io::Write, sync::atomic, thread};
+
+pub(crate) fn panic_handler(info: &PanicInfo) {
+    // keep track to ensure that we only panic once and multiple threads can exit gracefully!
+    static PANIC_ONCE: atomic::AtomicBool = atomic::AtomicBool::new(false);
```
kontheocharis, line 12: So, if two threads panic at the same time, one of them will not be shown. Is that a good idea?

kontheocharis (commented):
Wow, this looks really good!

compiler/hash-parser/src/lexer.rs
feds01: Yes, it could be factored out to just check if it is a white space and eat it.

compiler/hash-parser/src/lexer.rs
feds01: I think the next big step is to sort out `hash-reporting` and plug this in with it.

compiler/hash-parser/src/lexer.rs
feds01: It's added to enhance error handling and reporting capabilities. When these tokens are converted into AST nodes, we might want to report the error and then try and recover to see if there are more errors. A good example of where this behaviour could be useful in handling the `Greek question-mark tragedy`. There might be a greek question mark instead of a semi-colon, which is unexpected and thus we might want to report it with additional context.

compiler/hash-parser/src/lexer.rs
feds01: This is a bogus `println!`, should be removed.

compiler/hash-parser/src/lexer.rs
feds01: Old comment before the method was implemented.

compiler/hash-parser/src/lexer.rs
feds01: That is matched on `'`, and then `chars()` is called

compiler/hash-parser/src/lexer.rs
feds01: In any case, it still has to be collected into a string since the `parse` method expects one.

compiler/hash-parser/src/lexer.rs
feds01: Yes because if it is not collected, the iterator won't actually advance the character stream and hence the offset would be incorrect.

compiler/hash-parser/src/lexer.rs
feds01: Yep, this should be supported! Radix integer literals should also support digit separators.

compiler/hash-parser/src/lexer.rs
feds01: It doesn't really make a difference since all of the look-ahead functions check if EOF has been reached.

compiler/hash-parser/src/lexer.rs
kontheocharis: Im just blind, dw.

compiler/hash/src/crash_handler.rs
feds01: I think it is quite unlikely that two threads will panic at the same time, and if they do, what are the chances that they panic on entirely different things.

compiler/hash-ast/src/keyword.rs
```diff
@@ -0,0 +1,65 @@
...
+impl Keyword {
+    pub fn iterator() -> Iter<'static, Keyword> {
+        #[rustfmt::skip]
+        static DIRECTIONS: [Keyword; 16] = [
+            Keyword::Let, Keyword::For, Keyword::While, Keyword::Loop,
+            Keyword::If, Keyword::Else, Keyword::Match, Keyword::As,
+            Keyword::In, Keyword::Where, Keyword::Trait, Keyword::Enum,
+            Keyword::Struct, Keyword::Continue, Keyword::Break, Keyword::Return,
+        ];
+        DIRECTIONS.iter()
```
kontheocharis, line 36: There won't be any error if a new keyword gets added to the enum but not this list. There's a crate https://crates.io/crates/strum which might be useful here, which would autogenerate this method (and to_str for that matter). What do you think about using that?

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,631 @@
...
+        // > ':' => match self.peek() {
+        // >     ':' => {
+        // >         self.next();
+        // >         break TokenKind::NameAccess
+        // >     }
+        // >     _ => break TokenKind::Colon
+        // > },
+        //
+        // could work here, but however what about if there was a space or a comment between the colons, this might be
+        // problematic. Essentially, we pass the responsobility of forming more compound tokens to AST gen rather than here.
```
kontheocharis, line 130: Woah, very nice.

compiler/hash-parser/src/lexer.rs
```diff
@@ -0,0 +1,631 @@
...
+            ':' => TokenKind::Colon,
+            ';' => TokenKind::Semi,
+            ',' => TokenKind::Comma,
+            '.' => TokenKind::Dot,
+
+            // Consume a token tree, which is a starting delimiter, followed by a an arbitrary number of tokens and closed
+            // by a followiing delimiter...
+            ch @ ('(' | '{' | '[') => self
+                .eat_token_tree(Delimiter::from_left(ch).unwrap())
+                .unwrap(),
```
kontheocharis, line 154: Neat implementation.

kontheocharis (approved):
Looks great!

compiler/hash-ast/src/keyword.rs
feds01: I was looking for something like that, but I couldn't find anything particularly good. I'll check it out and have a look at it. Also, is there a way to create the hashmap at compile time to insert the keywords?

kontheocharis: Unfortunately not. The map has to be created at runtime. However, it can be written as a lazy static. (See `lazy_static` crate).

compiler/hash-parser/src/lexer.rs
kontheocharis: Just realised you got a typo `responsobility`

PR #60, opened 2021-08-06
Integrate hash-alloc with the rest of the compiler

kontheocharis:
Generally, this is done by replacing Box with Brick, and Vec with Row.


compiler/hash-pest-parser/src/translate.rs
```diff
@@ -1718,13 +1736,13 @@ mod tests {
...
-    // pub(crate) fn parse_input<T>(rule: Rule, input: &str) -> AstNode<T>
+    // pub(crate) fn parse_input<T>(rule: Rule, input: &str) -> AstNode<'c, T<'c>>
     // where
-    //     for<'a> HashPair<'a>: IntoAstNode<T>,
+    //     for<'a> HashPair<'a>: IntoAstNode<'c, T<'c>>,
     // {
     //     let resolver = SeqModuleResolver::new();
     //     let mut result = grammar::HashGrammar::parse(rule, input).unwrap();
-    //     let parsed: AstNode<T> = result.next().unwrap().into_ast(&resolver);
+    //     let parsed: AstNode<'c, T<'c>> = result.next().unwrap().into_ast(&resolver);
```
feds01, line 1745: Maybe we should make testing one of our next priorities

compiler/hash-parser/src/backend.rs
```diff
@@ -4,28 +4,42 @@
...
+
         let tokens = timed(
-            || tokenise(contents).collect::<Vec<_>>(),
+            || Lexer::new(contents, &wall).tokenise(),
             log::Level::Debug,
             |elapsed| println!("tokenise:    {:?}", elapsed),
         );
 
-        let mut gen = AstGen::new(tokens, resolver);
+        println!("Tokens bro: {:#?}", tokens);
```
feds01, line 40: Rogue print statement.

compiler/hash-ast/src/visualise.rs
```diff
@@ -160,7 +161,7 @@ impl NodeDisplay for Type {
             Type::Named(named_ty) => {
                 let mut components = vec![];
 
-                // add a mid connector for the lhs section, and then add vertical pipe
+                // add a mid connector for the<'_> lhs section, and then add vertical pipe
```
feds01, line 164: See above comment

compiler/hash-ast/src/literal.rs
```diff
@@ -28,23 +26,25 @@ lazy_static! {
...
             *key
         } else {
-            let ident = StringIdentifier::new();
+            let ident = StringLiteral::new();
 
             // copy over the string so that we can insert it into the reverse lookup table
-            let value_copy = value.clone();
+            // let wall = STATIC_CASTLE.wall();
+            // let value_copy = BrickString::new(value, &wall);
+            let value_copy = Box::leak(value.to_owned().into_boxed_str());
```
feds01, line 38: Is this still copying over the string?

compiler/hash-alloc/src/lib.rs
```diff
@@ -0,0 +1,509 @@
+//! Arena allocator implementation for use within the Hash compiler sources.
+//!
+//! All rights reserved 2021 (c) The Hash Language authors
+
+#![allow(dead_code)] // @@REMOVE
```
feds01, line 5: Remove ?

feds01 (commented):
Looks really good overall, just a few thoughts:

- I think we should document the allocator stuff before merging it into `main`.
- It will also be useful to merge `ast-gen` into this branch before merging into main given how the two branches have already diverged. It might be harder in the future to merge the two as development efforts progress.

kontheocharis:
Yes, I agree that it should be documented first. I can tackle the first two bullet points. As for the segfault, I still cannot reproduce it on my machine, can you?

compiler/hash-ast/src/visualise.rs
```diff
@@ -84,7 +85,7 @@ fn child_branch(lines: &[String]) -> Vec<String> {
     draw_branches_for_lines(lines, END_PIPE, PADDING)
 }
 
-/// Function to draw branches for a Vector of lines, connecting each line with the appropriate
+/// Function to draw branches for a<'_> Vector of lines, connecting each line with the appropriate
```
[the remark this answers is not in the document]
kontheocharis: Regex  ftw

compiler/hash-ast/src/visualise.rs
kontheocharis: I wonder why this was replaced...

compiler/hash-ast/src/literal.rs
kontheocharis: Yes, which isn't necessary. I am also not sure as to whether to keep this as Box::leak or use a static castle. In essence, both leak memory, but the latter requires a murex lock every time.

feds01:
Should we run a docker container for linux to see if it happens?

compiler/hash-ast/src/literal.rs
feds01: Keeping it as a `Box::leak` might be better if it avoids the mutex. Maybe it's better to put a `@@Cleanup` on this and leave it for another fight another day?

kontheocharis:
Closed in favour of #61.

PR #61, opened 2021-08-25
Second revision of hash_alloc, using `bumpalo` under the hood

kontheocharis:
This is the second revision of the arena allocator crate. The custom implementation is currently put aside in favour of a `bumpalo`-based implementation. This is because of safety issues that are rather difficult to debug (see #60).

This revision is complete with documentation and moderate testing.


compiler/hash-alloc/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "hash-alloc"
+version = "0.1.0"
+authors = ["The Hash Language authors"]
+edition = "2018"
+
+[dependencies]
+rayon = "1.5"
+dashmap = "4.0"
+parking_lot = "0.11"
```
feds01: What's this dependency for?

compiler/hash-alloc/src/brick.rs
```diff
@@ -0,0 +1,153 @@
...
+
+impl<T> Borrow<T> for Brick<'_, T> {
+    fn borrow(&self) -> &T {
+        self.deref()
+    }
+}
+
+impl<T> Drop for Brick<'_, T> {
+    fn drop(&mut self) {
+        // Safety: destructor only runs once.
```
feds01, line 102: Should this be prefixed with a `@@`?

compiler/hash-ast/src/count.rs
```diff
@@ -213,7 +213,7 @@ impl NodeCount for Type {
                 ty.name.node_count() + arg_count
             }
 
-            // TypeVar variant just counts for one node since it just wrapper for Name,
+            // TypeVar variant just counts for one<'_> node since it just wrapper for Name<'_>,
```
feds01, line 216: Funky Comment

compiler/hash-ast/src/visualise.rs
```diff
@@ -84,7 +85,7 @@ fn child_branch(lines: &[String]) -> Vec<String> {
     draw_branches_for_lines(lines, END_PIPE, PADDING)
 }
 
-/// Function to draw branches for a Vector of lines, connecting each line with the appropriate
+/// Function to draw branches for a<'_> Vector of lines, connecting each line with the appropriate
```
feds01, line 88: Funky comment

compiler/hash-ast/src/visualise.rs
```diff
@@ -412,7 +413,7 @@ impl NodeDisplay for Statement {
                 )
             }
             Statement::Assign(decl) => {
-                // add a mid connector for the lhs section, and then add vertical pipe
+                // add a mid connector for the<'_> lhs section, and then add vertical pipe
```
feds01, line 416: See above

compiler/hash-ast/src/visualise.rs
```diff
@@ -630,7 +634,7 @@ impl NodeDisplay for Expression {
             }
             ExpressionKind::Variable(var) => {
                 // check if the length of type_args to this ident, if not
-                // we don't produce any children nodes for it
+                // we don't produce any children nodes for it<'_>
```
feds01, line 637: See above

feds01 (changes requested):
This looks good to me, I think and would like `ast-gen` merged into this branch before this merged into `main`. 

This is because the new stuff in the tokeniser has already diverged from `ast-gen` and I don't think we should let it diverge more. Additionally, identifier path representations have also changed how they are stored, so we should fast forward on that. Finally, there has also been a lot of code added in `hash-parser/src/gen.rs`, which will depend on the new allocator API. What do you think?

compiler/hash-alloc/Cargo.toml
kontheocharis: Removed extraneous dependencies in the latest commit.

compiler/hash-alloc/src/brick.rs
kontheocharis: Well it could be, but it's not something that one would `grep` for. I just put this type of comment above every `unsafe` block as a hard rule. One would search for `unsafe` instead. But if you think it is better I can add `@@`.

kontheocharis:
Will merge ast-gen into this branch.

compiler/hash-alloc/src/brick.rs
feds01: Maybe we could have a different prefix like ':' or '::' to begin comments with a particular topic?

PR #67, opened 2021-10-29
AST: Complete implementation for AST generation from the parser.

feds01:


compiler/hash-ast/src/parse.rs
```diff
@@ -77,31 +80,47 @@ enum EntryPoint<'a> {
...
         Self {
             worker_count,
             backend,
+            visualise,
         }
     }
 
+    pub fn set_visualisation(&mut self, visualise: bool) {
+        self.visualise = visualise;
+    }
```
kontheocharis, line 113: Is this needed? Why would one want to change the visualisation of an already constructed parser?

compiler/hash-ast/src/parse.rs
```diff
@@ -178,22 +197,49 @@ where
...
+                    //            At the moment, this isn't the cleanest way of going about the problem, we're
+                    //            overwriting the previous content of the interactive session with the current
+                    //            input. This is incorrect because we want to preserve the lines of interactive
+                    //            that are considered to be valid (this is a much deeper problem actually.)
+                    //
+                    //            So, in @@Future, we shouldn't even insert the contents here since it needs
+                    //            to pass the typechecking stage and then be inserted into the line, I guess typechecking
+                    //            can *evict* the current source line if it is invalid?
+                    //
+                    //            @@Copying: we're also having to copy the source because we get a &str instead of a String.
```
kontheocharis, line 210: I wonder if EntryPoint::Interactive should contain a String instead of a &str.

compiler/hash-ast/src/parse.rs
```diff
@@ -178,22 +197,49 @@ where
                     );
                     let entry_resolver = ParModuleResolver::new(ctx, entry_module_ctx, scope);
 
+                    // @@Cleanup: we need to insert the contents of interactive into the module builder...
+                    //            At the moment, this isn't the cleanest way of going about the problem, we're
+                    //            overwriting the previous content of the interactive session with the current
+                    //            input. This is incorrect because we want to preserve the lines of interactive
+                    //            that are considered to be valid (this is a much deeper problem actually.)
```
kontheocharis, line 204: Why would we want to preserve the previous lines?

compiler/hash-ast/src/parse.rs
```diff
@@ -206,23 +252,51 @@ where
...
+
+        // If we need to visualise the file... then do so after the parser has finished
+        // the whole tree.
+        if self.visualise {
+            for module in modules.iter() {
+                println!(
+                    "file \"{}\":\n{}",
+                    module.filename().display(),
+                    module.ast()
+                );
```
kontheocharis, line 274: Maybe this should use `log` instead of print?

compiler/hash-ast/src/parse.rs
```diff
@@ -206,23 +252,51 @@ where
...
         let entry = EntryPoint::Interactive { contents };
-        let (interactive, modules) = self.parse_main(entry, directory)?;
-        Ok((interactive.unwrap(), modules))
+        let (result, modules) = self.parse_main(entry, directory);
+
+        // Ensure that this method always returns a body block or an
+        // error since it can never be `None`.
+        match result {
+            Ok(Some(block)) => (Ok(block), modules),
+            Ok(None) => unreachable!(),
```
kontheocharis, line 297: How do we know that it will always return a body block?

compiler/hash-ast/src/resolve.rs
```diff
@@ -118,15 +120,14 @@ where
                     },
                     Level::Debug,
                     |elapsed| println!("ast: {:.2?}", elapsed),
-                )?;
+                );
+
+                // @@Hack: we still need to add the contents of the file into the map
```
kontheocharis, line 125: Why is it necessary to split contents from node?

compiler/hash-ast/src/visualise.rs
```diff
@@ -177,7 +177,7 @@ impl NodeDisplay for Type<'_> {
             }
             Type::Ref(ref_ty) => child_branch(&ref_ty.node_display()),
             Type::RawRef(ref_ty) => child_branch(&ref_ty.node_display()),
-            Type::TypeVar(var) => child_branch(&[format!("var \"{}\"", var.name)]),
+            Type::TypeVar(var) => child_branch(&[format!("var {}", var.name)]),
```
kontheocharis, line 180: The visualiser could be hooked up to use `hash_utils::tree_writing`.

compiler/hash-parser/src/gen.rs
```diff
@@ -133,13 +131,20 @@ where
...
+
     /// Function that increases the offset of the next token
     pub(crate) fn next_token(&self) -> Option<&Token<'c>> {
         let value = self.stream.get(self.offset.get());
 
         if value.is_some() {
-            // @@Dumbness: because Cell::update is unsafe
-            self.offset.set(self.offset.get() + 1);
+            // @@UnsafeLibUsage
+            self.offset.update::<_>(|x| x + 1);
```
kontheocharis, line 147: Is this really faster than the way it was before?  Surely not, right?

compiler/hash-parser/src/gen.rs
```diff
@@ -177,24 +183,75 @@ where
...
+        self.error("Unexpectedly reached the end of file.")
+    }
+
+    /// Create a generalised "Reached end of file..." error.
+    pub(crate) fn unexpected_eof_with_ctx<T>(&self, ctx: impl Into<String>) -> ParseResult<T> {
+        self.error(format!(
+            "{}: but unexpectedly reached the end of file.",
+            ctx.into()
+        ))
+    }
```
kontheocharis, line 247: I wonder if these messages should be in a centralised error place (maybe in reporting?). Maybe the parsing should produce errors in the form of ADTs (enum variants with data).

compiler/hash-parser/src/gen.rs
```diff
@@ -278,151 +412,821 @@ where
...
+        // the order is (Lt, Eq, Gt)
+        let mut branches = match fn_ty {
+            CompoundFn::Leq => {
+                row![&self.wall; self.node(MatchCase {
+                    pattern: self.node(Pattern::Or(OrPattern {
+                        variants: row![&self.wall;
+                            self.node(Pattern::Enum(EnumPattern {
+                                name: self.make_access_name_from_str("Lt", location),
+                                args: row![&self.wall],
+                            })),
```
kontheocharis, line 659: Could possibly factor this into a `make_...` function because it is repeated a lot in this function.

compiler/hash-parser/src/gen.rs
```diff
@@ -1214,40 +2390,47 @@ where
...
-                Some(op) => {
+                // check if the operator here is re-assignable, as in '+=', '/=', if so then we need to stop
+                // parsing onwards because this might be an assignable expression...
+                // Only perform this check if know prior that the expression is not made of compounded components.
+                Some(op) if !op.assignable => {
                     // consume the number of tokens eaten whilst getting the operator...
-                    (0..consumed_tokens).for_each(|_| {
-                        self.next_token();
-                    });
+                    self.offset.update(|x| x + consumed_tokens as usize);
```
kontheocharis, line 2398: Why is consumed_tokens not a usize already?

compiler/hash-parser/src/gen.rs
```diff
@@ -1362,14 +2615,83 @@ where
...
+                    // This must be a map
+                    Some(token) if token.has_atom(TokenAtom::Colon) => {
+                        gen.skip_token();
+
+                        let rhs_type = gen.parse_type()?;
+
+                        // @@CopyPasta
+                        if gen.has_token() {
+                            gen.expected_eof()?;
+                        }
```
kontheocharis, line 2637: Why is this pattern necessary?

compiler/hash-parser/src/operator.rs
```diff
@@ -47,121 +56,217 @@ impl Operator {
...
-                }
-                _ => (Some(Operator::Lt), 1),
-            },
-            TokenKind::Atom(TokenAtom::Gt) => match gen.peek_second() {
-                Some(token) if token.kind == TokenKind::Atom(TokenAtom::Eq) => {
-                    (Some(Operator::GtEq), 2)
-                }
-                Some(token) if token.kind == TokenKind::Atom(TokenAtom::Gt) => {
-                    (Some(Operator::Shr), 2)
+        let (op, mut consumed): (_, u8) = match &(token.unwrap()).kind {
```
kontheocharis, line 59: See comment above about usize return type

compiler/hash-parser/src/operator.rs
```diff
@@ -47,121 +56,217 @@ impl Operator {
...
+                | OperatorKind::Shl
+                | OperatorKind::Add
+                | OperatorKind::Sub
+                | OperatorKind::Mul
+                | OperatorKind::Div
+                | OperatorKind::Mod
+        )
+    }
+
+    pub fn as_str(&self) -> &'static str {
```
kontheocharis, line 194: Maybe this isn't clear enough of a name. Maybe it should be called `descriptive_name` or something, because otherwise someone might think that .as_str() returns the actual operator (like the fmt::Display impl).

compiler/hash-parser/src/operator.rs
```diff
@@ -47,121 +56,217 @@ impl Operator {
...
+    fn from(op: Operator) -> Self {
+        let Operator { kind, assignable } = op;
 
-pub enum OperatorFn {
-    Named { name: &'static str, assigning: bool },
-    LazyNamed { name: &'static str, assigning: bool },
-    Compound { name: CompoundFn, assigning: bool },
-}
+        match kind {
+            OperatorKind::As => panic!("Cannot convert 'as' into a function call."),
```
kontheocharis, line 242: When we have types as values, this will probably be possible :).

compiler/hash-parser/src/token.rs
```diff
@@ -183,10 +306,15 @@ pub enum TokenAtom {
 
     Keyword(Keyword),
 
-    /// Delimiter: '(' '{', '[' and right hand-side variants, useful for error reporting and messages
+    /// General classification of an identifier
+    // @@Cleanup: find a better way to describe when we expect an identifier
+    GenericIdent,
```
kontheocharis, line 311: Why is this called GenericIdent?

compiler/hash-reporting/src/errors.rs
```diff
@@ -2,30 +2,87 @@
...
+            } => {
+                builder
+                    .add_element(ReportElement::CodeBlock(ReportCodeBlock::new(
+                        src,
+                        "Unknown path",
+                    )))
+                    .add_element(ReportElement::Note(ReportNote::new(
+                        "note",
+                        "Couldn't import this file path.",
+                    )));
```
kontheocharis, line 78: Cool!

compiler/hash-reporting/src/highlight.rs
```diff
@@ -0,0 +1,97 @@
+use std::ops::BitOr;
```
kontheocharis, line 1: How come the reporting stuff is here? Didn't we merge it to main before?

kontheocharis (commented):
Big PR! Very nice work.

compiler/hash-ast/src/ast.rs
```diff
@@ -579,8 +594,8 @@ pub enum ExpressionKind<'c> {
     Variable(VariableExpr<'c>),
     /// A property access.
     PropertyAccess(PropertyAccessExpr<'c>),
-    /// A reference expression.
-    Ref(AstNode<'c, Expression<'c>>),
+    /// A reference expression with a flag denoting whether it is a raw ref or not
+    Ref(AstNode<'c, Expression<'c>>, bool),
```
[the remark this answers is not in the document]
feds01: Yes this could be changed, but i think there might be alot of code duplication in some parts when dealing with refs. Admittedly, it might be better to distinguish the two when it comes to later stages where the difference does actually matter.

compiler/hash-ast/src/parse.rs
```diff
@@ -77,31 +80,47 @@ enum EntryPoint<'a> {
 }
 
 pub struct ParParser<B> {
+    /// Number of workers the parser should use to parse a job.
     worker_count: NonZeroUsize,
+    /// Whether or not to visualise the generated AST
+    visualise: bool,
+    /// What backend to use for parsing, whether PEST or self hosted.
```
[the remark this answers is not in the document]
feds01: Yeah good catch!

compiler/hash-ast/src/parse.rs
feds01: This comment is probably not well formed. It's not so much that we should store previous lines, but rather the generated AST from previous statements needs to be stored somewhere. Or it might not be the AST that is stored, but rather the typchecked results.

compiler/hash-ast/src/parse.rs
feds01: I tried changing this before but I ran into some problems, could be worth taking another look.

compiler/hash-reporting/src/highlight.rs
feds01: No, I don't believe it was merged into main, only into `ast-gen`

compiler/hash-parser/src/operator.rs
feds01: indeed

compiler/hash-parser/src/operator.rs
feds01: Or we could go the route to actually implement the trait `fmt::Debug` and use that to convert into a descriptive name?

compiler/hash-ast/src/module.rs
```diff
@@ -34,27 +35,32 @@ impl<'c> ModuleBuilder<'c> {
         Self::default()
     }
 
-    pub(crate) fn add_module_at(
+    pub fn add_module_at(
```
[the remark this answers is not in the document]
feds01: It was due to the fact that once the `add_module_at` is called, the contents of the `source` are moved and later accesses would require copying. This was a hack to get the reporting stuff to work (since it needs both the module index and the source available). We should take a look at fixing up this hack.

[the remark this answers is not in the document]
feds01: Actually this was pretty easy to fix.

compiler/hash-ast/src/resolve.rs
feds01: Yeah this was just a silly workaround whilst hooking up the reporting logic.

compiler/hash-ast/src/parse.rs
feds01: If it is an entry point to interactive job, it guarantees that a block will be returned from `parse_main`

compiler/hash-ast/src/ast.rs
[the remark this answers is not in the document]
kontheocharis: Then we might want a `RefType` enum at least, because we will need more than raw/non-raw references (e.g. weak references, mutable references in the future).

compiler/hash-ast/src/parse.rs
kontheocharis: I imagine we need to keep track of TypecheckCtx, but not much else.

compiler/hash-ast/src/parse.rs
kontheocharis: Fair enough; it might be worth it to add a message to unreachable, like `"Non-body block parsed node for interactive input"`.

compiler/hash-parser/src/token.rs
feds01: It's meant to represent a Identifier that starts with a capitol letter.

compiler/hash-parser/src/gen.rs
feds01: It's a `u8` because the number of chars can't be larger than a `u8`.

compiler/hash-parser/src/operator.rs
feds01: Given that how many times this function will be called, i think it's reasonable to keep the return type size as a u8.

compiler/hash-parser/src/gen.rs
feds01: Yes i think this can be done a little later once the parser gets expanded error handling. This should probably happen in a different PR.

compiler/hash-parser/src/gen.rs
feds01: This is just more ergonomic than getting, then setting.

compiler/hash-parser/src/gen.rs
feds01: Essentially assert the the inner generator has no more tokens to consume. If it does then it is an error and the parser couldn't eat the entire generator.

compiler/hash-ast/src/visualise.rs
feds01: work begun on `ast-treewriter`

kontheocharis (approved):
💥💥💥🎉🎉🎉

PR #68, opened 2021-11-21
Add a generic tree writer in `hash_utils`.

kontheocharis:
This can be used to print any sort of tree-like structure in various styles.


feds01 (approved):
LGTM

PR #69, opened 2021-11-27
Fix 66

feds01:
This PR fixes an issue with `parse_expression` and `parse_singular_expression`. If the expression is checking that there is a struct literal, and if so it shouldn't skip a token until it is verified that this matches a struct literal.

This PR also adds fixes the panic where if a file was incorrect and the reporting attempts to access file contents before they are added into the module builder.


compiler/hash-ast/src/module.rs
```diff
@@ -35,16 +35,7 @@ impl<'c> ModuleBuilder<'c> {
...
-        &self,
-        index: ModuleIdx,
-        path: PathBuf,
-        contents: String,
-        node: ast::Module<'c>,
-    ) {
-        self.path_to_index.insert(path.clone(), index);
-        self.filenames_by_index.insert(index, path);
-        self.contents_by_index.insert(index, contents);
+    pub fn add_module_at(&self, index: ModuleIdx, node: ast::Module<'c>) {
```
kontheocharis, line 38: See #70.

compiler/hash-parser/src/gen.rs
```diff
@@ -1628,7 +1628,7 @@ where
             ))?,
             // @@ErrorReporting
             None => self
-                .error("Expected block body, which begins with a '{{', but reached end of input")?,
+                .error("Expected block body, which begins with a '{', but reached end of input")?,
```
kontheocharis, line 1631: See #71.

kontheocharis (approved):

Issue #72, opened 2021-11-27
Improper error message when parsing invalid expressions:


feds01:
The following code is incorrect and shouldn't parse:

```rust
while i > {
}
```

However, the error that is printed by the parser is not correct:
```
 --> /Users/x/Documents/hash-org/lang/examples/visual.hash:1:11
1 |   while i > {
  |             ^
2 |   }
  |   ^ here
  = note: Expected block body, which begins with a '{{', but reached end of input
  ```

A more proper error message should be something like:
```
 --> /Users/x/Documents/hash-org/lang/examples/visual.hash:1:11
1 |   while i > {
  |             ^
  |             here
2 |   } 
  |
  = note: Expected expression after the operator '>'
  ```

feds01:
However, it is not known whether the user meant to add the block there or if the user meant to start a body block. For now, we can improve the error  message to expect a body block after the current token.

PR #73, opened 2021-11-28
fix: #72

feds01:
This PR improves a particular parser error messaging case where there is an expected block.  The message has been changed to more accurately reflect the location of the error rather than using the current token for the error span. The error location has been changed to use the next token after the current.


compiler/hash-parser/src/gen.rs
```diff
@@ -173,6 +173,18 @@ where
         }
     }
 
+    /// Get the next location of the token, if there is no token after, we use the
```
kontheocharis, line 176: Maybe "get the location of the next token", rather than "get the next location of the token"?

kontheocharis (approved):

PR #83, opened 2021-12-19
Implementation of AST Visitor

kontheocharis:
This pull request adds an implementation of the visitor pattern for the AST node structure. This is done to avoid repetition when recursing into children, for example in typechecking and visualisation.

The `AstVisitor` trait contains methods which can be implemented for each node in the tree. This is to be used in conjunction with functions inside `visitor::walk::*`, which traverse into the children and return the visitor results.

Each node has its own "return type" within the visitor, which allows for arbitrarily complicated traversal of the tree. Furthermore, enum types within the AST have `*_same_children` helper functions within `visitor::walk`, which automatically matches on the variants and traverses the variant enum members, as long as all variants have the same return type in the visitor (this is done using a type bound for associated types).

Things still to do before the PR can be merged:
- [x] Fix compile errors relating to slight reorganisation of some AST enum types.
- [x] Add any remaining implementations in `walk::`, specifically for statement types.
- [x] Add documentation to the `visitor` module.
- [x] Finish writing `ast::tree`, which is a proof-of-concept usage of this visitor that replaces the current `visualise` module.

Things to do after the PR is merged:
- [ ] Use the visitor to make typechecking more concise.
- [ ] Use the visitor for node counting.
- [ ] Investigate if there is a way to make a generic proc-macro which can generate a visitor from a tree description. This is most likely a longer-term goal but might come in handy because we will have other sorts of trees within the compiler (in the typechecker for example).

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -67,7 +67,7 @@ impl<'c, T> AstNode<'c, T> {
         self.body
     }
 
-    pub fn ast_ref<'s>(&'s self) -> AstNodeRef<'s, T> {
+    pub fn ast_ref(&self) -> AstNodeRef<T> {
         AstNodeRef {
             body: self.body.as_ref(),
             location: self.location,
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -67,6 +67,22 @@ impl<'c, T> AstNode<'c, T> {
         self.body
     }
 
+    pub fn ast_ref<'s>(&'s self) -> AstNodeRef<'s, T> {
+        AstNodeRef {
+            body: self.body.as_ref(),
+            location: self.location,
+            id: self.id,
+        }
+    }
+
+    pub fn with_body<'u, U>(&self, body: &'u U) -> AstNodeRef<'u, U> {
+        AstNodeRef {
+            body,
+            location: self.location,
+            id: self.id,
+        }
+    }
+
     /// Get the location of this node in the input.
     pub fn location(&self) -> Location {
         self.location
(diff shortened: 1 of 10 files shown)


compiler/hash-ast/src/tree.rs
```diff
@@ -0,0 +1,584 @@
+use hash_utils::tree_writing::TreeNode;
```
feds01, line 1: Wow holy shit, yeah we will need to look into having a macro solution for all of this code.

compiler/hash-ast/src/ast.rs
```diff
@@ -149,21 +205,31 @@ pub enum RefKind {
...
+#[derive(Debug, PartialEq)]
+pub struct RawRefType<'c>(pub AstNode<'c, Type<'c>>);
+
+/// The existential type (`?`).
+#[derive(Debug, PartialEq)]
+pub struct ExistentialType;
+
+/// The type infer operator.
+#[derive(Debug, PartialEq)]
+pub struct InferType;
```
feds01, line 222: Do these become zero sized?

compiler/hash-ast/src/ast.rs
```diff
@@ -149,21 +205,31 @@ pub enum RefKind {
...
-    /// A raw reference type
-    RawRef(AstNode<'c, Type<'c>>),
-    /// A type variable.
+    Ref(RefType<'c>),
+    RawRef(RawRefType<'c>),
     TypeVar(TypeVar<'c>),
-    /// The existential type (`?`).
-    Existential,
-    /// The type infer operator.
-    Infer,
```
feds01, line 166: Did you mean to remove comments on these variants?

feds01 (commented):
These are some pretty big changes, but I think we're definitely moving in the right direction.

feds01 (commented):
Once this PR is ready to merge, ill go in and snip the history and it will remove all the previous commits.

compiler/hash-ast/src/tree.rs
kontheocharis: Yes, although that would be for visitor.rs, not tree.rs.

compiler/hash-ast/src/ast.rs
kontheocharis: Yes.

compiler/hash-ast/src/ast.rs
kontheocharis: I did, because they all have a single member and the members have the same docs too. Otherwise docs would be duplicated for no real reason (in cargo-doc you can click on the member to view docs for it).

compiler/hash-ast/src/visitor.rs
```diff
@@ -0,0 +1,1579 @@
...
+            ast::ExpressionKind::Typed(inner) => {
+                Expression::Typed(visitor.visit_typed_expr(node.with_body(inner)))
+            }
+            ast::ExpressionKind::Block(inner) => {
+                Expression::Block(visitor.visit_block_expr(node.with_body(inner)))
+            }
+            ast::ExpressionKind::Import(inner) => {
+                Expression::Import(visitor.visit_import_expr(node.with_body(inner)))
+            }
+        }
```
feds01, line 517: It would be nice if u could do something like:
```rs
ExpressionKind:{variant}(inner) =>
        Expression::{variant}(...)
```

feds01 (approved):
LGTM, the weird history is fixed too!

compiler/hash-ast/src/visitor.rs
kontheocharis: Yeah, that would be a macro.

PR #85, opened 2021-12-22
Base tests for parser correctness

feds01:
This PR adds a set of basic tests to verify that the parser is able to correctly handle various edge cases, states and forms. This is likely not a complete set of feature tests and more will be added in the future. However, this set of tests is good enough to now be merged into mainstream, and be used by commit to verify the correctness of the parser and eliminate regression. Additionally, this PR drops support for the PEST parser as it's considered that the self-hosted parser is now fully-fledged, more accurate, faster and has a considerably lower memory footprint.

During the test writing, various bugs were discovered and fixed which are detailed below:
- better error messages for unclosed token trees.
- fix panic when a particular token tree was unclosed and at the end of the file.
- fix: don't put keywords in the identifier map as it increases the amount of work done by the hashmap lookups (saves about ~10ms avg).
- fix panic that occurs when reporting crate tries to print an error for module contents that don't exist. This mainly occurred to an edge case of an early failing condition and using default `Location` when converting an `ImportError` into a `ParseError`.
- fix very confusing error messages occurring inside of token trees that were being printed from the parent's context whereas they should have been referencing the child context.
- fix set and map literal parsing confusion with block statements.
- fix weirdness with struct literals and for-loops
- fix weirdness with re-assignment operators not being allowed in some cases for blocks.
- add support for type arguments when a function call is written in infix style.
- fix unclosed string literals not being shown as errors in the tokeniser.
- fix float literals not being able to be parsed with '_' in the decimal part.
- fix enforce Unicode escape literals have a maximum of 6 hex chars.
- fix support re-assignment in parentheses and other locations.
- fix dangling token trees allowed if the previous token tree was of the same delimiter type.

Miscellaneous improvements:
- Use Rust 2021
- Upgrade all project dependencies
- Added support struct literal field name shortcut de-sugaring (ecd5527aa8a2d9344e5bba6f9fcc1e393ddf2509)
- standardise the logging label style so that it conforms to the rest of the project style.
- Add `SliceDisplay` in `hash-utils` (moved from the `hash-parser` crate) for pretty-printing vectors in an English-like form.


compiler/hash-ast/src/resolve.rs
```diff
@@ -129,20 +129,20 @@ where
...
+                //         contents of the file for reporting purposes.
+                ctx.module_builder
+                    .add_contents(import_index, resolved_import_path, import_source);
+
+                match import_node {
+                    Ok(module) => {
+                        ctx.module_builder.add_module_at(import_index, module); // Add the import to modules
+                        Ok(())
+                    }
+                    Err(err) => Err(err),
```
kontheocharis, line 144: ? syntax could be used here.

compiler/hash-utils/src/printing.rs
```diff
@@ -0,0 +1,39 @@
+use std::fmt;
+
+/// Type wrapper around a slice of items.
+pub struct SliceDisplay<'a, T: 'a>(pub &'a [T]);
```
kontheocharis, line 4: Given the docs below, maybe it would be better to call this "SequenceDisplay" or something, not the programming-specific "SliceDisplay". Also since this is the purpose of this struct, maybe the doc comment should be on the struct itself and not on the fmt::Display implementation.

tests/parser/benches/benches.rs
```diff
@@ -0,0 +1,40 @@
+#![feature(test)]
+extern crate test;
+
+use hash_alloc::Castle;
+use hash_ast::module::ModuleIdx;
+use hash_parser::lexer::Lexer;
+use test::{black_box, Bencher};
+
+static STRINGS: &str = r#""tree" "to" "a" "graph" "that can" "more adequately represent" "loops and arbitrary state jumps" "with\"\"\"out" "the\n\n\n\n\n" "expl\"\"\"osive" "nature\"""of trying to build up all possible permutations in a tree." "tree" "to" "a" "graph" "that can" "more adequately represent" "loops and arbitrary state jumps" "with\"\"\"out" "the\n\n\n\n\n" "expl\"\"\"osive" "nature\"""of trying to build up all possible permutations in a tree." "tree" "to" "a" "graph" "that can" "more adequately represent" "loops and arbitrary state jumps" "with\"\"\"out" "the\n\n\n\n\n" "expl\"\"\"osive" "nature\"""of trying to build up all possible permutations in a tree." "tree" "to" "a" "graph" "that can" "more adequately represent" "loops and arbitrary state jumps" "with\"\"\"out" "the\n\n\n\n\n" "expl\"\"\"osive" "nature\"""of trying to build up all possible permutations in a tree.""#;
```
kontheocharis, line 9: Wow

tests/parser/tests/cases/should_fail/destructure_onto_void/case.hash
```diff
@@ -0,0 +1,2 @@
+// This doesn't make much sense semantically either.
+let (,) = k;
```
kontheocharis, line 2: Why should this fail? Can we not match on empty tuples?

tests/parser/tests/cases/should_fail/integer_literal_too_large/case.hash
```diff
@@ -0,0 +1 @@
+let k = 23384912398412397432849123413027948091234809213480923409823231498;
```
kontheocharis, line 1: What is the maximum integer literal supported?

kontheocharis: Also, what about negative literals?

tests/parser/tests/cases/should_fail/non_hex_unicode_char/case.hash
```diff
@@ -0,0 +1 @@
+let k = '\u{g00000}';
```
kontheocharis, line 1: Everybody gangsta until g shows up in hex

tests/parser/tests/cases/should_pass/compount_pattern_matching_with_structs/case.hash
```diff
@@ -0,0 +1,6 @@
+let data = match some_struct {
+    Dog { name, age = 6 } | Dog { name, age = 5 }  => do_something_with_name();
+    Dog { name, age = 13 } | Dog { name, age } if age % 2 == 0  => do_something_with_name();
+    Dog { name, age } => do_something_with_name();
+    _ => #unreachble();
+};
```
kontheocharis, line 6: This is neat! (though you misspelled unreachable)

tests/parser/tests/cases/should_pass/if_branches/case.hash
```diff
@@ -0,0 +1,7 @@
+if b == 2 {
+     print("b is 2")
+} else if b == 3 {
+     print("b is 3")
+} else {
+    print("b isn't 2 or 3 ")
+}
```
kontheocharis, line 7: Is this meant to be an expression? How does it pass without semis?

tests/parser/tests/cases/should_pass/inline_comment/case.hash
```diff
@@ -0,0 +1,8 @@
+if b == 2 {
+     print("b is 2")
+} else if b == 3 {
+     // print("b is 3")
+     a
+} else {
+    print("b isn't 2 or 3 ")
+}
```
kontheocharis, line 8: See previous.

tests/parser/tests/cases/should_pass/nested_re_assignment/case.hash
```diff
@@ -0,0 +1,2 @@
+// This technically should parse, but it wouldn't make sense since the '+=' operator has a void return type
```
kontheocharis, line 1: Is that a part of the trait definition?

tests/parser/tests/cases/should_pass/parenthesised_for_loop/case.hash
```diff
@@ -0,0 +1,3 @@
+(for name in names.iter() {
```
kontheocharis, line 1: Why should this pass?

tests/parser/tests/cases/should_pass/parenthesised_while_loop/case.hash
```diff
@@ -0,0 +1,6 @@
+(
```
kontheocharis, line 1: See for loop comment.

tests/parser/tests/cases/should_pass/pattern_matching_let/case.hash
```diff
@@ -0,0 +1,6 @@
+// struct desturcturing
+let Dog { name } = dog;
+let Dog { name = better_name } = dog;
+
+// namespace destructuring
+let { name: even_better_name} = dog;
```
kontheocharis, line 6: What do you think about having "=" for namespace destructuring, same as structs?

tests/parser/tests/cases/should_pass/refutable_pattern_destructuring/case.hash
```diff
@@ -0,0 +1,6 @@
+// Well technically this pattern is 'irrefutable', even if we know that this is true, the compiler
+// cannot sensibly reason about it much further than parsing...
```
kontheocharis, line 2: Unless we have literal types :)

tests/parser/tests/cases/should_pass/struct_pattern_matching/case.hash
```diff
@@ -0,0 +1,5 @@
+let data = match some_struct {
+    Dog { name, age = 6 } => do_something_with_name();
+    Dog { name, age } => do_something_with_name();
+    _ => #unreachble();
```
kontheocharis, line 4: Same typo

tests/parser/tests/cases/should_pass/unary_ops/case.hash
```diff
@@ -0,0 +1,3 @@
+let k = ~a;
+let k = ---a;
```
kontheocharis, line 2: In the future, we probably want a lint to say "Hey, --a does not mean to decrement a!".

kontheocharis (approved):
incroyable!

tests/parser/tests/cases/should_fail/destructure_onto_void/case.hash
feds01: Yes, this can be added as support in the parser although it doesn't make much sense to do this.

tests/parser/tests/cases/should_fail/integer_literal_too_large/case.hash
feds01: The maximum integer supported is a `u64`, however, it maybe problematic with signed literals as then it becomes 2^6 3

tests/parser/tests/cases/should_pass/refutable_pattern_destructuring/case.hash
feds01: yeees

tests/parser/tests/cases/should_pass/unary_ops/case.hash
feds01: that's a good point, it should be explicitly stated in the docs that it doesn't do what you think it does.

tests/parser/tests/cases/should_pass/if_branches/case.hash
feds01: well because it is treated as a block and thus it becomes an expression in the block. It might also be the case that it is at the end of the file.

tests/parser/tests/cases/should_pass/nested_re_assignment/case.hash
feds01: No i don't think it should be, ill remove this comment.

tests/parser/tests/cases/should_fail/integer_literal_too_large/case.hash
feds01: We could get around this by not converting to negative straight away and using the `neg` trait?

compiler/hash-ast/src/resolve.rs
feds01: Well unfortunately the call still has to be made to add the contents of the file into the module builder:
```rs
 ctx.module_builder.add_contents(import_index, resolved_import_path, import_source);
```

If the call is made before, the string is moved and the borrow checking complains about this.

tests/parser/tests/cases/should_pass/parenthesised_for_loop/case.hash
feds01: Treated as an expression at the end of the module.

tests/parser/tests/cases/should_pass/pattern_matching_let/case.hash
feds01: I think it would be a lot better to standardise syntax and make the parsing code a little cleaner.

compiler/hash-ast/src/resolve.rs
kontheocharis:
```rs
import_node.map(|module| ctx.module_builder.add_module_at(import_index, module))
```
Doesn't this do the same thing?

tests/parser/tests/cases/should_fail/destructure_onto_void/case.hash
kontheocharis: Well, you're right it doesn't make sense now, but it might make sense if we ever have type unions. I guess for now it can be left like this.

tests/parser/tests/cases/should_fail/integer_literal_too_large/case.hash
kontheocharis: hmm, true. Maybe that is indeed a good idea. However, negative integer literals won't be able to be represented up to -(2^64-1) anyway in the language (it will be -(2^63-1) again).

tests/parser/tests/cases/should_pass/if_branches/case.hash
kontheocharis: But doesn't an expression in statement position (since this is a module [right?]) require a semicolon at the end?

tests/parser/tests/cases/should_pass/parenthesised_for_loop/case.hash
kontheocharis: Okay, resolving this, see previous response.

tests/parser/tests/cases/should_pass/pattern_matching_let/case.hash
kontheocharis: Yeah I agree.

tests/parser/tests/cases/should_pass/unary_ops/case.hash
kontheocharis: Maybe an issue can be created for that.

tests/parser/tests/cases/should_pass/pattern_matching_let/case.hash
feds01: Ill make an issue after this is merged.

tests/parser/tests/cases/should_pass/if_branches/case.hash
feds01: Not if it's a block

compiler/hash-ast/src/resolve.rs
feds01: `add_module_at` doesn't add the contents of the source for the very reason.

PR #92, opened 2021-12-25
Wrap `ast::Module` nodes inside `ast::AstNode` in `Modules`

kontheocharis:
This is done for the sake of consistency with `parse_interactive`, and provides more information about a module (i.e. the span of the file that contains it, as well as the filename of the file).

This is also needed to be able to pass a module into an `ast::visitor::AstVisitor`.


feds01 (approved):
LGTM

PR #93, opened 2021-12-25
Lexer: Partially flatten tokens

feds01:
After some experimentation, it was evident that avoiding using inlined 'token trees' could yield better performance since the memory use of the TokenKind would be reduced by 40%. So this patch stores `token_trees` in a separate variable which is represented as a row of rows of tokens, which can then be later indexed by the `TokenKind::Tree(delim, token_tree_index)` token kind.

At a later stage, the lexer will entirely move away from token_trees to further improve lexer performance despite the parsing logic having to be slightly more complicated when not working around the trees.


kontheocharis (approved):
I'm guessing that the rest of the flattening will occur in a separate PR?

Issue #96, opened 2021-12-31
Switch to using a channel/queue system for creating identifiers across parsing threads


feds01:
After some profiling efforts, a realisation came to light that about 15% of time parsing/AST generation was spent just waiting on threads to acquire the identifier/string hash map mutex and insert identifiers/strings into the map. 

After some consideration, we can attack the problem in two distinct vectors:

- Reduce waiting on hashing and lookups in the hashmap.
- Avoid waiting on acquiring the mutex and getting the result for the map inseration.

The first point can be addressed by using the crate `slotmap` which promises the very efficient (`O(1)`) insertion/lookup times without the need for hashing. The official documentation and paper for the CPP version is available [here](https://github.com/orlp/slotmap). By using `slotmap` we don't have to depend on hashing, and incrementing an atomic counter for the identifiers.

Secondly, we can use a channel to in order to have the `SlotMap` live in the main thread, and worker threads send tasks (add an identifier) via the queue (possibly using `crossbeam`) to maximise efficiency.

PR #103, opened 2022-01-06
Parser: Improve and refactor `Identifier` handling

feds01:
This PR improves the handling of identifiers in the parser and removes a bunch of redundant behaviour in the `IdentifierMap` type. The changes can be summarised by the following points:
- Use `ThreadLocal` to keep a wall on each thread in order for faster allocations when creating identifiers and inserting them into the identifier map.
- Don't even try to allocate before looking up if an identifier exists in the map.
- Better naming convention in the `IdentifierMap` type:
  - `get_ident` gets an identifier 
  - `create_ident` creates an identifier :))
- Introduced a concept of `CORE_IDENTIFIERS` which is a wrapper around `IDENTIFIER_MAP` that contains inserted core identifiers like the primitives and the `_` identifier.
- Removed dependency on `parking_lot` as we no longer need to lock the wall per each thread.
- Implementation for `From<u32> for Identifier` and `From<Identifier> for u32`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -319,7 +319,6 @@ dependencies = [
  "lazy_static",
  "log",
  "num_cpus",
- "parking_lot",
  "rayon",
  "strum",
  "strum_macros",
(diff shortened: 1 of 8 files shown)


feds01:
closes #96

kontheocharis (approved):

PR #104, opened 2022-01-11
Reporting: Add basic reporting functionality to typechecking

feds01:
This PR implements basic reporting functionality to the type checker. The error variants have all been verified to carry type information in one way or another.

It was decided that type definitions (structs and enums) should store the `SourceLocation` with the definition and types of expressions should be stored in a separate location map that might or might not have type information on a particular type.


compiler/hash-ast/src/module.rs
```diff
@@ -97,6 +89,22 @@ impl<'c> Modules<'c> {
...
+            // @@Copying
+            map.insert(
+                *key.0,
+                SourceModule::new(key.1.to_owned(), contents.to_owned()),
+            );
+        });
+
+        SourceMap::new(map)
+    }
+
```
kontheocharis, line 107: Is it worth it to do all of this copying? Can we not have SourceMap be a trait instead?

compiler/hash-source/src/module.rs
```diff
@@ -0,0 +1,47 @@
...
+
+#[derive(Debug)]
+pub struct SourceMap {
+    map: HashMap<ModuleIdx, SourceModule>,
+}
+
+impl SourceMap {
+    pub fn new(map: HashMap<ModuleIdx, SourceModule>) -> Self {
+        Self { map }
+    }
```
kontheocharis, line 43: Maybe this should take an iterator instead, it is kind of leaking private implementation otherwise.

compiler/hash-typecheck/src/error.rs
```diff
@@ -1,25 +1,374 @@
...
+                    )))
+                    .add_element(ReportElement::Note(ReportNote::new(
+                        "note",
+                        format!("The field `{}` doesn't exist on type `{}`.", name, ty_name),
+                    )));
+            }
+        }
+
+        // @@ErrorReporting: we might want to properly handle incomplete reports?
+        builder.build().unwrap()
```
kontheocharis, line 372: When would such a thing happen?

compiler/hash-typecheck/src/traverse.rs
```diff
@@ -101,6 +103,22 @@ impl<'c, 'w, 'm, 'g, 'i> ModuleTypechecker<'c, 'w, 'm, 'g, 'i> {
...
+        match self.module_or_interactive {
+            ModuleOrInteractive::Module(idx) => idx,
+            ModuleOrInteractive::Interactive(_) => ModuleIdx(0),
+        }
+    }
+
+    pub fn source_location(&self, location: Location) -> SourceLocation {
+        SourceLocation {
+            location,
+            module_index: self.current_module(),
```
kontheocharis, line 118: To prevent the hack above, maybe this should use ModuleOrInteractive?

kontheocharis:
Great work!

compiler/hash-typecheck/src/traverse.rs
```diff
@@ -976,7 +1117,14 @@ impl<'c, 'w, 'm, 'g, 'i> visitor::AstVisitor<'c> for ModuleTypechecker<'c, 'w, '
                     .iter()
                     .map(|&ty| match self.types().get(ty) {
                         TypeValue::Var(var) => Ok(*var),
-                        _ => Err(TypecheckError::BoundRequiresStrictlyTypeVars),
+                        _ => {
+                            // We need to get the location of the bound for error reporting
+                            let bound_location = node.bound.as_ref().unwrap().location();
```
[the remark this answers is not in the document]
feds01: Yes because this error can only occur if a bound was present on struct definition right. Since the previous item maps `bound.as_ref().map(|bound| {...})` and thus we know that the bound exists here.

compiler/hash-typecheck/src/error.rs
feds01: This was just a copied over a comment from the `ParseError` implementation.

PR #106, opened 2022-01-11
Typechecking: A first attempt at trait resolution

kontheocharis:
This PR implements a minimum version of trait resolution, albeit not functional yet. It is being re-integrated into `typechecking` because it also contains other useful additions, including more typechecking errors, folding and mapping over types, as well as typechecking trait definitions and implementations. This PR also changes `AstNodes` to be a named struct rather than a type alias, adding a convenience method to get the span location of the inner nodes. Furthermore, the `ast_nodes` macro is created to simplify the creation of this struct and generally aid in refactoring.


compiler/hash-ast/src/ast.rs
```diff
@@ -145,12 +145,61 @@ impl<T> Deref for AstNodeRef<'_, T> {
     }
 }
 
-// @@TODO: Ideally we want some useful methods on this and potentially a span that represents
-//         all the nodes inside it. Why is this useful? Well because sometimes we want to represent
-//         a collection ob sub nodes that might have a larger parent span like for example arguments
-//         in a function def. The entire AstNodes might include the span of the parenthesees but each
-//         argument might just have the span of the argument name and type.
-pub type AstNodes<'c, T> = Row<'c, AstNode<'c, T>>;
```
feds01, line 153: I would prefer to keep this comment or make a gh issue about parent span

compiler/hash-typecheck/src/traits.rs
```diff
@@ -185,10 +160,136 @@ pub struct CoreTraits {
...
+    }
+
+    pub fn match_trait_impl(
+        &mut self,
+        trait_impl: &TraitImpl,
+        _trait_args: &[TypeId],
+    ) -> TypecheckResult<Substitution> {
+        let trt = self.global_storage.traits.get(trait_impl.trait_id);
+
+        // @@Ambiguity: for now let's assume all type variables in here are new
```
feds01, line 261: Is this an issue with scoped type variables?

compiler/hash-typecheck/src/traverse.rs
```diff
@@ -1032,35 +1109,75 @@ impl<'c, 'w, 'm, 'g, 'i> visitor::AstVisitor<'c> for ModuleTypechecker<'c, 'w, '
         ctx: &Self::Ctx,
         node: ast::AstNodeRef<ast::LetStatement<'c>>,
     ) -> Result<Self::LetStatementRet, Self::Error> {
```
feds01, line 1034: Maybe split this in two functions to handle bounds separately?

feds01 (commented):
One step closer to full tc

compiler/hash-ast/src/ast.rs
kontheocharis: See #107.

compiler/hash-typecheck/src/traits.rs
kontheocharis: No, this will actually change but for now I left it to instantiate all type variables. It is not really ambiguous, more like ambiguous to the programmer what it means. Like ```let foo<$T> = () => { let bar<$T> = () => { ..}; };` is the inner `$T` a new variable, or the same as the outer `$T`?

compiler/hash-typecheck/src/traverse.rs
kontheocharis: Yeah, I will do this once I also add support for trait bounds.

compiler/hash-typecheck/src/traits.rs
feds01: Ah right ok

Issue #99, opened 2022-01-06
Typechecking: Abstract the typechecker behind a trait similarly to parser


kontheocharis:
We want to do this because we don't want to couple the rest of the compiler to the current typechecking implementation. Having a trait similar to Parser but for typechecking will allow us to write code that typechecks opaquely and is thus more reusable.

kontheocharis:
Done in https://github.com/hash-org/lang/tree/compiler-pipeline

PR #111, opened 2022-01-14
Pipeline: A new way of compiler bootstrapping

kontheocharis:
This PR introduces a series of restructuring operations on the core compiler workings:

- `hash_ast` no longer contains any parsing-related code; this is up to each parser to implement.
- `hash_pipeline` is a new crate which contains bootstrapping logic for the compiler. For one, this includes traits `Parser` and `Checker` for implementations of parsing and typechecking. These are currently implemented by `hash_parser` and `hash_typecheck` respectively. Finally, the `Compiler` type manages `Parser` and `Checker`, and is able to keep checker state between interactive invocations.
- The `Modules` type has been removed in favour of `Sources`. `Sources` does not allow concurrent access, and as such performance should be improved. Furthermore, `Sources` keeps track of both interactive blocks as well as modules, and keeps the sources, AST, and filenames for each one of these.
- `hash_parser` now contains a new implementation of parallel parsing, using a channel and `rayon` threadpool to delegate parsing jobs. This design is much simpler and most likely more performant (although benchmarks are required to verify this).


kontheocharis:
Closes #99.

feds01 (approved):

PR #115, opened 2022-01-19
Various compiler Improvements

### LINKED ISSUE #107: AST: Keep track of a span in `AstNodes`
Instead of using the location of the children, `AstNodes` should be given a location on construction, in case there are no children. This would also simplify the `.location()` method.

### LINKED ISSUE #112: Improve error message for binary exprs
The current error message for unclosed binary expressions in blocks is not informative enough, we should make it better. Here is an example of the case:
```
>>> let f = () => { l < };
error[0001]: Failed to parse
 --> <interactive>:1:19
1 |   let f = () => { l < };
  |                     ^ here
  = note: Unexpectedly reached the end of input.
 ```

### LINKED ISSUE #113: Parser: Malformed error report:
Code snippet:
```rs
let viktor = Dog { 
    name = "Viktor", +
};
```

Error report:
```
error: Failed to parse
 --> /Users/alex/Documents/hash-org/lang/examples/test.hash:9:20
8 |   let viktor = Dog { 
9 |       name = "Viktor", +
  |                      ^ here
10 |       // age = 12, 
  = note: Expected an expression here, however received a '+'. Consider adding 
```

### LINKED ISSUE #114: Incorrect span for operator
The span for the operator is invalid.
```
error[0006]: Symbol `add` is not defined in the current scope.
 --> <interactive>:1:5
1 |   x + 2
  |       ^ not found in this scope
  ```

### PR DESCRIPTION
**Pull request by feds01**
- Added `hash-error-codes` to host general error codes that can occur at the type checking stage.
- remove `hash-ast/visualise.rs` as it's about to be replaced by the new tree writer.
- Fixed some various parsing errors and updated documentation in various locations. 
- Added parent span to `AstNodes`
- moved operator definitions from `hash-parser` into `hash-ast`
- added typechecking error reporting for an invalid number of arguments to a function.
- cleanup how pipeline compiler is created (move it out of `hash-interactive`).

closes #113 
closes #114 
closes #112 
closes #107

### CODE REVIEW

File: compiler/hash-ast/src/lib.rs [hunk 1]
```diff
@@ -1,9 +1,17 @@
...
 #![feature(iter_intersperse)]
 
+use hash_alloc::Castle;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    /// AST static allocator
+    pub static ref STATIC_CASTLE: Castle = Castle::new();
+}
+
```
kontheocharis, line 14: Eventually I expect we want to remove this?

compiler/hash-ast/src/operator.rs
```diff
@@ -1,37 +1,201 @@
...
+    Mod,
 
-// Types of unary operators that might need to be transformed.
-pub enum UnaryOpType {
-    FnCall(&'static str),
-    Ref,
-    Deref,
+    /// 'as'
+    // @@Cleanup: This is a special operator since it performs a higher order operation compared
+    //            to all other members of the Operator enum, maybe it should be in it's own enum??
```
kontheocharis, line 61: Good point, what does "higher order" mean. You mean like it is not just between expressions?

compiler/hash-ast/src/operator.rs
```diff
@@ -1,37 +1,201 @@
...
-    Leq,
-    Geq,
-    Lt,
-    Gt,
+/// This implementation will be used for error messages. The debug formatter is
+/// mainly used in error reports whereas the display formatter is used when
+/// printing code from tokens.
+impl std::fmt::Debug for OperatorKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
```
kontheocharis, line 98: Isn't this the same as the to_string implementation? Maybe this should be derived instead to show the actual enum variant?

compiler/hash-ast/src/operator.rs
```diff
@@ -1,37 +1,201 @@
...
+
+    /// Compound functions that use multiple function calls when they are transformed.
+    pub fn is_compound(&self) -> bool {
+        matches!(
+            self,
+            OperatorKind::Lt | OperatorKind::Gt | OperatorKind::LtEq | OperatorKind::GtEq
+        )
+    }
+
+    pub fn to_string(&self) -> &'static str {
```
kontheocharis, line 160: Possibly better to call this `to_str` because `to_string` is also a trait method of ToString/Display and since this type implements Display that can get confusing

compiler/hash-error-codes/src/error_codes.rs
```diff
@@ -0,0 +1,28 @@
...
+    ExpectingBooleanInCondition = 15,
+    MissingStructField = 16,
+    BoundRequiresStrictlyTypeVars = 17,
+    ExpectingBindingForTraitImpl = 18,
+    TraitDefinitionNotFound = 19,
+    TypeAnnotationNotAllowedInTraitImpl = 20,
+    TypeArgumentLengthMismatch = 21,
+    NoMatchingTraitImplementations = 22,
+    FunctionArgumentLengthMismatch = 23,
+}
```
kontheocharis, line 28: We probably also want to find a way of erroring at compile time if a number is used twice..

compiler/hash-pipeline/src/lib.rs
```diff
@@ -1,3 +1,15 @@
+//! Hash Compiler pipeline implementation. The pipeline is a abstract representation
+//! of the compiler flow managing the compiling steps like parsing, typechecking, optimisation
+//! passes, etc. The pipeline is used to abstract away the idea of depending on specific
+//! implementations of the parser or typechecker and just use a common trait
+//! interface that can be used. This file also has definitions for how to access
+//! sources whether module or interactive.
+//!
+//! @@Cleanup: Probably move all the source map stuff somewhere else?
```
kontheocharis, line 8: Maybe it would make sense to put it in hash-source?

compiler/hash-typecheck/src/error.rs
```diff
@@ -463,6 +510,60 @@ impl TypecheckError {
...
+                            "note",
+                            "consider removing the last argument.".to_string(),
+                        )));
+                    } else {
+                        builder.add_element(ReportElement::Note(ReportNote::new(
+                            "note",
+                            format!("consider removing the last {} arguments.", diff),
+                        )));
+                    }
+                }
```
kontheocharis, line 565: Very nice!

compiler/hash-ast/src/operator.rs
feds01: Yeah I think this was just copied over from the old sources.

compiler/hash-error-codes/src/error_codes.rs
feds01: That would be neat.

compiler/hash-ast/src/lib.rs
feds01: Yes definitely

compiler/hash-pipeline/src/lib.rs
feds01: The only problem is that `Module` depends on `hash-ast` and `hash-source` cannot depend on it since that would create a cyclical dependency.

compiler/hash-ast/src/operator.rs
feds01: Yeah, it's kind of strange because it's like on an operator but it receives special handling when parsing. Take the code snippet below:
https://github.com/hash-org/lang/blob/79961772ac1854b626d0650a981474dcbdb3a664/compiler/hash-parser/src/gen.rs#L2660-L2673

All other operators go through a particular transformation where they're transpiled, whereas for the `as` operator, the expression gets converted into an `ExpressionKind::Typed`.  This is fine, however, it could somehow be improved to not contain such special handling, especially if we add operators of a similar nature.

kontheocharis: Fair enough, I guess it is because it always has the lowest precedence and also takes a type and an expression? Maybe that's what should be stated.

compiler/hash-pipeline/src/lib.rs
kontheocharis: I see. Then maybe it is fine where it is? Maybe it should be placed in its own module like `sources`?

feds01: Yeah I think so

compiler/hash-ast/src/operator.rs
feds01: Yeah

PR #117, opened 2022-01-21
Typechecking: Minimum traits

kontheocharis:
This PR implements an MVP version of trait resolution and inference, including inferring trait implementations from function call arguments and surrounding context.

In order to achieve this feat, the nature of unknown types has changed. Each unknown type carries an ID (`UnknownTypeId`, which points to a slot with an optional "resolved" type). It is no longer the case that `TypeId`s themselves can have their type swapped; this is only true for unknown types now. This allows types to be entirely deep-copied without losing the identity of each unknown type within them. This theoretically improves type inference across the board, not just for traits.

There are a few more things to tackle after this PR regarding traits and generic resolution:
- Implement support for trait bounds.
- Find a way to not have super deep nested unknown types, which happens with many unifications.
- Perhaps rename "unknown types" to something a bit more fitting.


feds01 (approved):
LGTM

PR #118, opened 2022-01-21
Typechecker: Implement tuple types

feds01:
This patch implements tuple types in the typechecker by modifying the way they are represented in the AST. 

Additionally:
- This also adds some minor improvements to AST visualisation. 
- Reporting boundary between line numbers and code is highlighted in blue to give a clear boundary between both elements.


compiler/hash-interactive/src/lib.rs
```diff
@@ -134,7 +134,13 @@ where
                         .visit_body_block(&(), block.node())
                         .unwrap();
 
-                    println!("{}", TreeWriter::new(&tree));
+                    // Don't use any spaces between the label and child...
+                    let config = TreeWriterConfig {
```
kontheocharis, line 138: Maybe this should be the default behaviour?

kontheocharis (approved):

PR #122, opened 2022-01-22
AST: use special variant for function types.

### LINKED ISSUE #119: Create an AST function type instead of using named type "Function". This is to circumvent the hack in traversing function types that is currently in place in the typechecker.

feds01:
Similarly to tuple types, the current way of handling function types
was to check if an identifier was being used to denote the function
as 'Function' which was very hacky. This patch removes this behaviour
and adds a new variant to the `Type` variant called `Fn`.

This makes it considerably easier to work within the typechecker
and the visualiser.

closes #119


kontheocharis (approved):
LGTM

PR #126, opened 2022-01-23
Typecheck: Implement patterns and match blocks

### LINKED ISSUE #123: Implement typechecking patterns
Patterns include:
- [x] Structs
- [x] Enums
- [x] Tuples
- [x] binding patterns
- [x] if patterns
- [x] or patterns
- [x] Literals (string, char, float, int)
- [ ] namespaces

### LINKED ISSUE #124: Implement match cases

### LINKED ISSUE #125: Implement assignment statements

feds01:
This PR brings functionality to the typechecker for typechecking patterns and match blocks. The current typechecking strategy for match blocks is to take the matched expression and verify that all patterns unify to the same type, if they do not that is invalid. The second check that is performed ensures that the return type of all branches in a match statement is of the same type (otherwise this would be incorrect behaviour).

The PR can be summarised by the following changes:
- Implements typechecking for the match block that.
- Implement typechecking for struct patterns.
- Implement typechecking for enum patterns.
- Implement assignment statements.
- Refactor some typechecking errors to have unified logic in regards to argument length mismatching.

closes #125
closes #124 
closes #123


compiler/hash-typecheck/src/error.rs
```diff
@@ -336,13 +396,51 @@ impl TypecheckError {
...
+                }
+
+                builder
+                    .add_element(ReportElement::CodeBlock(ReportCodeBlock::new(
+                        location, "here",
+                    )))
+                    .add_element(ReportElement::Note(ReportNote::new(
+                        ReportNoteKind::Note,
+                        format!(
+                            "This variant `{}` has an invalid number of arguments.",
```
kontheocharis, line 435: Maybe s/invalid/incorrect/ ?
Also it might be useful to print the actual numbers.

compiler/hash-typecheck/src/traverse.rs
```diff
@@ -1477,19 +1473,139 @@ impl<'c, 'w, 'g, 'src> visitor::AstVisitor<'c> for SourceTypechecker<'c, 'w, 'g,
...
+                match variant_type {
+                    TypeValue::Fn(FnType {
+                        args: expected,
+                        return_ty,
+                    }) => {
+                        // let received_ty =  TypeValue::Fn(FnType {
+                        //     args,
+                        //     ret: variant_id,
+                        // });
+
```
kontheocharis, line 1510: Delete?

compiler/hash-typecheck/src/error.rs
```diff
@@ -36,6 +36,43 @@ impl Symbol {
...
+
+            format!("consider adding {} argument(s).", diff)
+        } else {
+            let diff = given - wanted;
+
+            if diff == 1 {
+                "consider removing the last argument.".to_string() // @@Reporting: suggestions!
+            } else {
+                format!("consider removing the last {} arguments.", diff)
+            }
```
kontheocharis, line 71: The wrong argument(s) aren't always going to be the last. Maybe it's better to only say this if the types agree too?

kontheocharis (approved):

compiler/hash-typecheck/src/error.rs
feds01: I guess we can determine this when we make the error?

compiler/hash-typecheck/src/error.rs
```diff
@@ -432,13 +452,14 @@ impl TypecheckError {
                     .add_element(ReportElement::Note(ReportNote::new(
                         ReportNoteKind::Note,
                         format!(
-                            "This variant `{}` has an invalid number of arguments.",
-                            name
+                            "This variant `{}` has an invalid number of arguments. The variant {}.",
```
kontheocharis, line 455: Still says "invalid".

PR #129, opened 2022-02-22
typechecker: implement refutability check.

### LINKED ISSUE #127: Add pattern refutability check
This is a very crude implementation of refutability and is only used as a prototype before the real refutability checking is added via pattern usefulness checking algorithm. 

Some useful links:
- The algorithm is described in this paper [here](http://moscova.inria.fr/~maranget/papers/warn/index.html).
- A variant of the algorithm was implemented in the Rust compiler [here](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_mir_build/thir/pattern/usefulness/index.html).

Our goal with this task is to implement a very fast and dirty version of the algorithm that can handle *most* refutability checks. In cases where it cannot be proven that the pattern is irrefutable, the algorithm will throw its hands up and give up.

feds01:
closes #127

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -443,7 +443,7 @@ pub struct EnumPattern<'c> {
     /// The name of the enum variant.
     pub name: AstNode<'c, AccessName<'c>>,
     /// The arguments of the enum variant as patterns.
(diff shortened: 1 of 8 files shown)


kontheocharis (approved):
Yes.

PR #135, opened 2022-02-22
Typecheck namespace patterns

kontheocharis:
This PR adds support for namespace patterns, which can be used to extract contents from modules.


compiler/hash-typecheck/src/traverse.rs
```diff
@@ -1619,10 +1630,74 @@ impl<'c, 'w, 'g, 'src> visitor::AstVisitor<'c> for SourceTypechecker<'c, 'w, 'g,
...
+                                variable_type_id,
+                                UnifyStrategy::ModifyTarget,
+                            )?;
+                        }
+                        Some(symbol_type) => match field.pattern.body() {
+                            ast::Pattern::Binding(BindingPattern(binding)) => {
+                                self.scopes().add_symbol(binding.ident, symbol_type);
+                            }
+                            _ => {
+                                return Err(TypecheckError::DisallowedPatternNonVariable(
```
feds01, line 1671: Maybe add a note here to why this happens

feds01 (approved):

PR #177, opened 2022-05-06
Compiler: split parser into lexer and parser

feds01:
This patch extracts the lexer component of the parser from `hash-parser` into `hash-lexer`. Additionally, the token definitions have been moved into `hash-token` so that a third party can also access `Token` definitions.

Things still to do:
- [ ] If it's better to convert to using `Report` instead of passing around `ParseError` and `LexerError`
- [ ] How to make the lexing and parsing stages more specific parts within `hash-pipeline`


compiler/hash-lexer/src/error.rs
```diff
@@ -0,0 +1,92 @@
...
+#[error("{kind} {}", .message.as_ref().unwrap_or(&String::from("")))]
+pub struct LexerError {
+    pub(crate) message: Option<String>,
+    kind: LexerErrorKind,
+    pub location: Location,
+}
+
+/// This implementation exists since we can't use tuples that are un-named
+/// with foreign module types.
+pub struct LexerErrorWrapper(pub SourceId, pub LexerError);
```
kontheocharis, line 26: Maybe it would be better to call this `LexerErrorAtSource` or equivalent?

kontheocharis (approved):

PR #185, opened 2022-05-09
parser: Implement spread patterns, within tuples and lists

feds01:
This patch implements `spread` patterns within the parser. This patch allows spread patterns to only occur within list and tuple patterns (although this is subject to change). Additionally, this patch fixes a bug within the pattern parsing that allows invalid patterns that follow within a pattern collection to be ignored by the parser because the generator was not being checked to be empty.

Items that are still outstanding:
- [x] Write tests for list patterns, and write tests for spread patterns within tuples and lists.


compiler/hash-parser/src/parser/pattern.rs
```diff
@@ -230,6 +243,14 @@ impl<'c, 'stream, 'resolver> AstGen<'c, 'stream, 'resolver> {
             }
         }
 
+        // @@ErrorReporting: So here, there is a problem because we do actually want to report
+        //                   that this should of been the end of the pattern but because in some
```
kontheocharis, line 247: "should have"

compiler/hash-parser/src/parser/pattern.rs
```diff
@@ -230,6 +243,14 @@ impl<'c, 'stream, 'resolver> AstGen<'c, 'stream, 'resolver> {
...
+        //                   contexts the function is being peaked and the error is being ignored,
```
kontheocharis, line 248: peeked.

kontheocharis (approved):

PR #187, opened 2022-05-19
Parser: New struct enum syntax

feds01:
This patch changes the parser to support the new syntax for `struct` and `enum` definitions. This means that some components of the parser had to significantly change and some parts were disabled whilst the new syntax is being implemented.


kontheocharis (approved):
I guess it is failing cause of tests.

PR #188, opened 2022-05-19
Parser: convert all statements into expression variants

feds01:
This patch converts all of the variants that were in `Statement` into `ExpressionKind` equivalents and re-organises some of the parsing code.

fixes #188


compiler/hash-ast/src/ast.rs
```diff
@@ -912,33 +857,53 @@ pub enum ExpressionKind<'c> {
     Typed(TypedExpr<'c>),
     Block(BlockExpr<'c>),
     Import(ImportExpr<'c>),
+    StructDef(StructDef<'c>),
+    EnumDef(EnumDef<'c>),
+    Bound(Bound<'c>),
```
kontheocharis, line 862: Can this be renamed to `TypeFunctionDef`? Also, maybe `FunctionDef` should be here too, instead of `LiteralExpr::Function`?

compiler/hash-ast/src/ast.rs
```diff
@@ -912,33 +857,53 @@ pub enum ExpressionKind<'c> {
...
     pub fn kind(&self) -> &ExpressionKind<'c> {
         &self.kind
     }
+
+    pub fn no_continuation(&self) -> bool {
+        matches!(
+            self.kind(),
+            ExpressionKind::Import(_) | ExpressionKind::Break(_) | ExpressionKind::Continue(_)
+        )
+    }
```
kontheocharis, line 899: Is this really necessary?

kontheocharis: If so, maybe a comment as to why would be useful?

compiler/hash-parser/src/parser/definitions.rs
```diff
@@ -0,0 +1,151 @@
...
+use crate::parser::error::TyArgumentKind;
+
+use super::{error::AstGenErrorKind, AstGen, AstGenResult};
+
+impl<'c, 'stream, 'resolver> AstGen<'c, 'stream, 'resolver> {
+    /// Parse a [StructDefinition] which includes the name of the struct with a
+    /// ForAll to specify any bounds or and generic arguments to the struct, with
+    /// zero or more struct fields. An example for a struct would be:
+    ///
+    /// Name := <T,Q> where eq<T> => struct( ... );
```
kontheocharis, line 18: This will need to change eventually to be more like function defintions.

kontheocharis (approved):

compiler/hash-ast/src/ast.rs
feds01: Yeah I think this was a remnant of the old syntax, I will now remove it.

kontheocharis (approved):

PR #205, opened 2022-05-23
Language: Support new trait syntax and type functions defs

feds01:
This patch supports the new `trait` block syntax and the new syntax for specifying arguments within type function definitions. This patch also drops support for the old `where ...` type bounds syntax.


compiler/hash-ast/src/ast.rs
```diff
@@ -657,15 +667,10 @@ pub struct EnumDef<'c> {
     pub entries: AstNodes<'c, EnumDefEntry<'c>>,
 }
 
-/// A trait definition, e.g. `trait add = <T> => (T, T) => T;`.
+/// A trait definition, e.g. `add := <T> => trait { add: (T, T) => T; }`.
```
kontheocharis, line 670: It should be `add: (T, T) -> T` (single arrow)

kontheocharis (approved):

PR #213, opened 2022-05-25
Parser: Refine type parsing

feds01:
This patch implements the slightly amended type parsing and adds support for the new specified type [grammar](https://hash-org.github.io/lang/features/types.html).

 In summary, this patch changes/adds the following:
- Remove `TypeVar`, `Infer` and `Existential` variants from`Type`
- `Set`, `Map` and `List` are now variants of `Type`
- `NamedType` no longer holds `type_args` as that is structured as a `TypeFunctionCall`
- Added `MergedType` which allows you to specify the intersection of types  (primarily used in type bounds).
- Added support for parenthesising a `Type` which makes it a `GroupedType`. The difference between a `TupleType` and `GroupedType` is that `TupleType`s must have multiple members or single element `TupleType`s should have a trailing comma at the end of the parentheses.
- Added support for `TypeFunction` and `TypeFunctionCall` variants which allows to specify the `type` of a `TypeFunction`.
- Use a `Vector` instead of re-allocating `row` each time for `AstNodes`. Only create a row in the end when the whole `Vector` has been constructed.
- rename `parse_token_atom` to`parse_token`
- rename `parse_token_atom_fast` to`parse_token_fast`

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -315,6 +315,25 @@ pub struct TupleType<'c> {
     pub entries: AstNodes<'c, NamedFieldTypeEntry<'c>>,
 }
 
+/// The list type, , e.g. `{str}`.
+#[derive(Debug, PartialEq)]
+pub struct ListType<'c> {
+    pub inner: AstNode<'c, Type<'c>>,
+}
+
+/// The set type, , e.g. `{str}`.
+#[derive(Debug, PartialEq)]
+pub struct SetType<'c> {
+    pub key: AstNode<'c, Type<'c>>,
+}
+
+/// The map type, e.g. `{str: u32}`.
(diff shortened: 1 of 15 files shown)


compiler/hash-ast/src/ast.rs
```diff
@@ -315,24 +300,82 @@ pub struct TupleType<'c> {
...
 /// A type.
 #[derive(Debug, PartialEq)]
 pub enum Type<'c> {
     Tuple(TupleType<'c>),
+    List(ListType<'c>),
+    Set(SetType<'c>),
+    Map(MapType<'c>),
     Fn(FnType<'c>),
     Named(NamedType<'c>),
+    Grouped(GroupedType<'c>),
```
kontheocharis, line 373: Not necessary

kontheocharis (approved):
Very nice sir

PR #219, opened 2022-05-27
Parser: Use `as` syntax in namespace patterns

feds01:


docs/src/features/patterns.md
```diff
@@ -259,21 +259,21 @@ or_pattern = ( single_pattern "|" )+ single_pattern
 
 binding_pattern = identifier
 
-tuple_pattern_member = identifier | ( identifier "=" single_pattern )
+tuple_pattern_member = identifier | ( identifier "=" identifier )
```
kontheocharis, line 262: Why is this changed?

feds01: I thought that it doesn't make sense to attach anything but a bind to the range pattern

PR #226, opened 2022-05-31
Parser: Speed improvement of parsing declarations

feds01:
This patch adds/fixes the following:

-  Improve(speed) parsing of decls/expressions to determine if the parser should try to parse a declaration or an expression.
- `MergeDeclaration` lhs should be an `Expression` and not a `Pattern`.

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -652,7 +652,7 @@ pub struct Declaration<'c> {
 #[derive(Debug, PartialEq)]
 pub struct MergeDeclaration<'c> {
     /// The pattern to bind the right-hand side to.
-    pub pattern: AstNode<'c, Pattern<'c>>,
+    pub decl: AstNode<'c, Expression<'c>>,
 
     /// Any value that is assigned to the binding, simply
     /// an expression.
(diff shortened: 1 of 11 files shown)


compiler/hash-ast/src/ast.rs
```diff
@@ -652,7 +652,7 @@ pub struct Declaration<'c> {
 #[derive(Debug, PartialEq)]
 pub struct MergeDeclaration<'c> {
     /// The pattern to bind the right-hand side to.
```
kontheocharis, line 654: Maybe change the comment too?

stdlib/num.hash
```diff
@@ -8,59 +8,55 @@
...
-trait asinh = <T> => (T) => T;
-trait acosh = <T> => (T) => T;
-trait atanh = <T> => (T) => T;
+Pow ~= impl float {
+    pow := (base, exponent) => intrinsic_pow(base, exponent);
+};
+
+Pow ~= impl int {
+    pow := (base, exponent) => intrinsic_pow(base, exponent);
+};
```
kontheocharis, line 27: These should be the other way around -- float ~= impl Pow, etc

kontheocharis (approved):

PR #231, opened 2022-06-13
Add scaffolding for the new typechecker

kontheocharis:
The recent language changes have warranted a re-write of the typechecking logic, due to the difference in which traits, tuples and functions work, as well as the introduction of type functions. This PR adds the fundamental building blocks for the new typechecker. These include:
- Definitions for the primitives of the typechecker: types, values (which are now kept track of if they are compile-time), as well as module and nominal (struct and enum) definitions.
- A formatter for types that is more sophisticated than the old one, with support for grouping complex expressions, as well as using defined names in the output instead of the fully expanded form.
- A builder (specifically `PrimitiveBuilder`) that allows for painless creation of complex nested types without having to manually construct types in the stores. Ideally, the stores should never be directly accessed in AST traversal code, only higher level builders like `PrimitiveBuilder`.
- A module structure that separates inferring logic, type evaluation/substitutions/unifications, type storage, and errors/formatting.

Lots of time has been spent making the type system consistent, and even though this is not necessarily the case yet, it certainly appears so without having written any of the AST traversal logic. Docs about how types work can be found in `storage::primitives`. The notion of the "level" of a type is introduced, which separates traits, types, and values. The following diagram illustrates this:
```
Traits (Level 2)
    ^
    |
    | instance of
    |
 Types (Level 1)
    ^
    |
    | instance of
    |
 Values (Level 0)
```
This allows having types and values of each level, and all of these interoperating together in the program, which mirrors the recent changes in the language. Type functions and type function application are their own citizens (as well as type merging), whose levels depend on their (eventual) return type. In other words, the level of `<A, B, C, ...> -> D` is the same as the level of `D`. Equivalently, for types `A`, `B` and `C` of level N, `A ~ B ~ C` has level N.

Subsequent PRs will implement more of the unification logic (which is currently WIP), as well as the AST traversal and adding the types to AST nodes. After all this is completed, the monomorphisation process can begin.


kontheocharis:
Closes #231

kontheocharis:
Need to integrate with the pipeline before merge.

compiler/hash-typecheck/src/fmt.rs
```diff
@@ -0,0 +1,440 @@
...
+            }) => {
+                write!(f, "{}", name)
+            }
+            NominalDef::Struct(_) => {
+                write!(f, "struct(..)")
+            }
+            NominalDef::Enum(_) => {
+                write!(f, "enum(..)")
+            }
+        }
```
feds01, line 241: So, im guessing later these will have impls for fields of these types

compiler/hash-typecheck/src/storage/primitives.rs
```diff
@@ -0,0 +1,514 @@
...
+/// then combined.
+///
+/// The `general_return_ty` field is always a supertype of the return type of each case.
+#[derive(Debug, Clone)]
+pub struct TyFnValue {
+    /// An optional name for the type function, if it is directly assigned to a binding.
+    pub name: Option<Identifier>,
+    pub general_params: Params,
+    pub general_return_ty: TyId,
+    pub cases: Vec<TyFnCase>,
```
feds01, line 291: This is a pretty interesting way to store what are essentially implementations of this type function

compiler/hash-typecheck/src/storage/primitives.rs
```diff
@@ -0,0 +1,514 @@
...
+    /// Has a level 2 type.
+    Trt(TrtDefId),
+    /// A value returned by the `type` keyword.
+    ///
+    /// Has a level 1 type.
+    Ty(TyId),
+    /// A runtime value.
+    ///
+    /// Has a level 0 type.
+    Rt,
```
feds01, line 392: what would be a runtime value?

compiler/hash-typecheck/src/storage/scope.rs
```diff
@@ -0,0 +1,125 @@
...
+/// If the kind is `Constant`, then all the members have:
+/// - Any `visibility`
+/// - `mutability = Immutable`
+///
+/// If the kind is `Variable`, then all members have:
+/// - `visibility = Private`
+/// - Any `mutability`
+///
+/// @@Future: Maybe we should store the origin of each scope in some place, for better error
+/// messages.
```
feds01, line 34: Could add a span to it indeed

feds01 (approved):

compiler/hash-ast/src/ident.rs
```diff
@@ -7,7 +7,7 @@ use fnv::FnvBuildHasher;
 use hash_alloc::{collections::string::BrickString, Castle, Wall};
 use hash_utils::counter;
 use lazy_static::lazy_static;
-use std::{borrow::Borrow, thread_local};
+use std::{borrow::Borrow, fmt::Display, thread_local};
 
 counter! {
```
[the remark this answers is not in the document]
kontheocharis: Where do you think it should be?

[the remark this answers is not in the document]
feds01: Could put it into `hash-source`

compiler/hash-typecheck/src/fmt.rs
feds01: Could make a `verbose` mode or something of that nature

kontheocharis: Yeah, I'll add a @@Future comment for it.

compiler/hash-ast/src/ident.rs
```diff
@@ -16,6 +16,20 @@ counter! {
...
+    }
+}
+
+/// Create an identifier from a string.
+impl From<&str> for Identifier {
+    fn from(name: &str) -> Self {
+        IDENTIFIER_MAP.create_ident(name)
+    }
+}
+
```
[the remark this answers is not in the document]
kontheocharis: Opened #234.

compiler/hash-ast/src/ident.rs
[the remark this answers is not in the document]
kontheocharis: Opened #234.

PR #251, opened 2022-06-15
Pipeline revamp

### LINKED ISSUE #239: LM: Rework the compiler pipeline to be more agnostic to the specific stages

feds01:
closes #239


compiler/hash-ast-desugaring/src/lib.rs
```diff
@@ -17,49 +17,82 @@ use hash_ast::{
...
+            for (id, module) in sources.iter_mut_modules() {
+                // Skip any modules that have already been de-sugared
+                if state.contains(&SourceId::Module(id)) {
+                    continue;
+                }
+
+                for expr in module.node_mut().contents.iter_mut() {
+                    scope.spawn(|_| {
+                        AstDesugaring
+                            .visit_expression(&(), expr.ast_ref_mut())
```
kontheocharis, line 83: Would be nice if the pool was a context to the visitor so that each expression could be queued rather than the top-level ones. Though I guess that might result in other bottlenecking related to queuing costs..

compiler/hash-ast-desugaring/src/lib.rs
```diff
@@ -17,49 +17,82 @@ use hash_ast::{
...
+                            .unwrap()
+                    })
+                }
+            }
+        });
+
+        // Add all of the id's into the cache
```
kontheocharis, line 90: `id's` -> `ids`

kontheocharis (approved):

compiler/hash-ast-desugaring/src/lib.rs
feds01: Yes, that would be really nice, it could basically find body blocks and just queue all of the children to be "evaluated" and later processed.

I will add a note about this.

PR #253, opened 2022-06-15
parser: introduce `set` and `map` syntax

### LINKED ISSUE #252: Introduce `map` and `set` syntax

feds01:
This patch adds the `set!` and `map!` syntax to denote literals rather than dealing with the `set/map/block` ambiguity.

In the future, these should become macros rather than being keywords with a hard-coded lexer skip when it encounters `!`

This also cleans up a lot of the token tree parsing by introducing the `parse_delim_tree` function, which is general-purpose for parsing the next token and producing an `AstGen` generator.

closes #252


tests/parser/tests/cases/should_pass/map_literals/case.hash
```diff
@@ -1,6 +1,7 @@
-my_map := {:};
-my_map := {"a": 1 };
-my_map := {"a": 1, };
-my_map := {"a": 1 + 2, };
-my_map := {"a": 1, "b": 2};
-my_map := {"a": 1, "b": 2,};
+my_map := map!{};
```
kontheocharis, line 1: Do you reccomend keeping a space between `!` and `{` or not?

PR #281, opened 2022-06-18
semantics: implement warnings about useless expressions

### LINKED ISSUE #280: Warn about using int/str/float/tuple-containing-literals literals in statement position (will have no effect)

feds01:
Note: pr #279 should be approved and merged first!

This PR expands the semantic analysis stage to generate warnings about the parsed tree. It is in it's very early stages and only currently looks for 'useless' expressions: those that are constant and are in the statement positions within a body block, e.g. the following source:
```
main := () => {
   r := input<str>();
   "a";
    
   print("Hello " + r + "!");
};
```
would generate the error:
```rust
warn: this expression is useless
 --> /Users/alex/Documents/hash-org/lang/examples/weird.hash:3:4
2 |      r := input<str>();
3 |      "a";
  |      ^^^ here
4 |       
  = note: A constant expression in a body block that has no side-effects is useless
```

closes #280


compiler/hash-ast-passes/src/analysis/warning.rs
```diff
@@ -0,0 +1,53 @@
...
+            AnalysisWarningKind::UselessExpression => {
+                builder
+                    .with_message("this expression is useless")
+                    .add_element(ReportElement::CodeBlock(ReportCodeBlock::new(
+                        warning.location,
+                        "here",
+                    )))
+                    .add_element(ReportElement::Note(ReportNote::new(
+                        ReportNoteKind::Note,
+                        "A constant expression in a body block that has no side-effects is useless",
```
kontheocharis, line 46: Maybe "redundant" rather than "useless"?

kontheocharis (approved):

Issue #284, opened 2022-06-18
TC: Implement general term validation


kontheocharis:
See comments in `validate.rs` file in the typechecker `ops` module for information about what this would entail.

PR #286, opened 2022-06-18
TC: Implement getting the type of a term

### LINKED ISSUE #285: TC: Implement getting the type of a term

kontheocharis:
This PR implements resolving the type of any given term, to produce another term. This is very useful for type inference and needed for unification. This operation lives in `Typer`. Also, this PR implements struct field accessing for runtime terms (which was missed in the last simplification PR #282), and also adds scaffolding for term validation in `Validator`. 

Closes #285.

The next step after this PR is to tackle #284, namely term validation, which should be relatively quick. After that, unification can be tackled directly.

### DIFF
--- a/compiler/hash-typecheck/src/ops/simplify.rs
+++ b/compiler/hash-typecheck/src/ops/simplify.rs
@@ -3,8 +3,8 @@
 // @@Remove
 #![allow(unused)]
 use super::{
-    params::pair_args_with_params, substitute::Substituter, unify::Unifier, validate::TermLevel,
-    AccessToOps, AccessToOpsMut,
+    building::PrimitiveBuilder, params::pair_args_with_params, substitute::Substituter,
+    unify::Unifier, validate::TermLevel, AccessToOps, AccessToOpsMut,
 };
 use crate::{
     error::{TcError, TcResult},
@@ -42,7 +42,7 @@ impl<'gs, 'ls, 'cd> AccessToStorageMut for Simplifier<'gs, 'ls, 'cd> {
 fn does_not_support_access<T>(access_term: &AccessTerm) -> TcResult<T> {
     Err(TcError::UnsupportedPropertyAccess {
         name: access_term.name,
-        value: access_term.subject_id,
+        value: access_term.subject,
     })
 }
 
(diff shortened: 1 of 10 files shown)


compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -33,31 +35,136 @@ impl<'gs, 'ls, 'cd> AccessToStorageMut for Validator<'gs, 'ls, 'cd> {
...
+            Term::TyFn(_) => todo!(),
+            Term::TyFnTy(_) => todo!(),
+            Term::AppTyFn(_) => todo!(),
+            Term::AppSub(_) => todo!(),
+            Term::Unresolved(_) => todo!(),
+            Term::Level3(_) => todo!(),
+            Term::Level2(_) => todo!(),
+            Term::Level1(_) => todo!(),
+            Term::Level0(_) => todo!(),
+        }
```
feds01, line 157: How would you determine that a term is constant?

feds01 (approved):

compiler/hash-typecheck/src/ops/simplify.rs
```diff
@@ -783,3 +850,82 @@ impl<'gs, 'ls, 'cd> Simplifier<'gs, 'ls, 'cd> {
...
+
+    use crate::{
+        fmt::PrepareForFormatting,
+        storage::{core::CoreDefs, primitives::ModDefOrigin, GlobalStorage, LocalStorage},
+    };
+
+    use super::*;
+
+    #[test]
+    fn test_simplify() {
```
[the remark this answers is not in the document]
kontheocharis: I'm planning on adding such tests once unification is implemented, cause now most of them would just panic.

compiler/hash-typecheck/src/ops/validate.rs
kontheocharis: Basically any expression that isn't a struct, type function, enum, trait, mod/impl, type function application, access, enum variant, or a function literal, is not constant.

feds01: So what's a straight forward example of a non-constant term, I am guessing a runtime value?

Also, could this be useful for other applications when lowering?

kontheocharis: `land_on_the_moon()` is not constant. Neither is `3`. As far as the typechecker is concerned they are both runtime values. In the future this can be relaxed if literal types are implemented.

kontheocharis: This could be useful for lowering, but it would have to be a bit more sophisticated to actually be useful, like e.g. to detect pure functions Vs side effects.

feds01: Yes I see now, this makes more sense (and it could definitely become more sophisticated when literal types are implemented).

Issue #295, opened 2022-06-19
TC: Implement usage of default arguments


kontheocharis:
Currently, default parameter values are not implemented in the typechecker. There are @@Todo items where this needs to be added.

PR #298, opened 2022-06-19
TC: Implement term unification

### LINKED ISSUE #232: LM: Implement the rest of the type unification logic

### LINKED ISSUE #283: tc: check that all named function call args come after positional args

### LINKED ISSUE #297: TC: Add positional argument check back to pair_.. function

kontheocharis:
This PR implements unification of terms, completing the core of the typechecker (along with simplification, substitutions, and typing). Unification takes two terms A and B, and ensures that A is a sub-term or equal term to B, that is, any instance of A can be used where B is required. This asymmetry only applies to instantiable terms (level 1 and above); for uninstantiable terms, unification is simply equality checking. Unification produces a substitution as a result, which can be applied to surrounding terms to infer more information about them (this is how type inference works).

The remaining tasks in the typechecker are:

- Implement further term validation (see `validate.rs`) (see #284)
- Write tests for simplification, unification, substitutions, and typing (see #296).
- Implement default variables (see #295)

These should be done in separate PRs, in order to keep the changes manageable.

Closes #232, closes #283, closes #297.

### DIFF
--- a/compiler/hash-typecheck/src/ops/validate.rs
+++ b/compiler/hash-typecheck/src/ops/validate.rs
@@ -171,7 +171,38 @@ impl<'gs, 'ls, 'cd> Validator<'gs, 'ls, 'cd> {
     /// Determine if the two given substitutions are equivalent.
     ///
     /// That is, if for any term X, they produce the same result when applied to X
-    pub fn subs_are_equivalent(&mut self, _s0: &Sub, _s1: &Sub) -> TcResult<bool> {
-        todo!()
+    ///
+    /// @@Correctness: This is not based on any accepted algorithm, and requires testing to ensure
+    /// its correctness.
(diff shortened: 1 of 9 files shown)


compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -167,4 +167,42 @@ impl<'gs, 'ls, 'cd> Validator<'gs, 'ls, 'cd> {
...
         }
     }
+
+    /// Determine if the two given substitutions are equivalent.
+    ///
+    /// That is, if for any term X, they produce the same result when applied to X
+    ///
+    /// @@Correctness: This is not based on any accepted algorithm, and requires testing to ensure
+    /// its correctness.
+    pub fn subs_are_equivalent(&mut self, s0: &Sub, s1: &Sub) -> bool {
```
feds01, line 177: This is pretty cool

feds01 (approved):

PR #301, opened 2022-06-20
TC: Minor improvements to simplification and adding Root term

kontheocharis:
This PR includes a set of minor improvements to term simplification, notably flattening nested `Term::Merge`s. Furthermore, the `Root` term is introduced, which is used as the type of `TraitKind` and `TyFnTy`. This is needed because typing `TyFnTy` as `TraitKind` is not consistent (since it is not a trait). The `Root` term is also typed as `Root`, and is the only term to do so. This PR is a precursor to tackling #284.

### DIFF
--- a/compiler/hash-typecheck/src/error.rs
+++ b/compiler/hash-typecheck/src/error.rs
@@ -67,6 +67,14 @@ pub enum TcError {
         args: Args,
         unification_errors: Vec<TcError>,
     },
+    InvalidElementOfMerge {
+        term: TermId,
+    },
+    MergeShouldOnlyContainOneNominal {
+        merge_term: TermId,
+        nominal_term: TermId,
+        second_nominal_term: TermId,
+    },
     UnsupportedTypeFunctionApplication {
         subject_id: TermId,
     },
--- a/compiler/hash-typecheck/src/ops/validate.rs
+++ b/compiler/hash-typecheck/src/ops/validate.rs
@@ -1,6 +1,6 @@
 //! Contains utilities to validate terms.
 use crate::{
-    error::TcResult,
+    error::{TcError, TcResult},
     storage::{
         primitives::{FnTy, Level1Term, ModDefId, NominalDefId, Sub, Term, TermId, TrtDefId},
         AccessToStorage, AccessToStorageMut, StorageRefMut,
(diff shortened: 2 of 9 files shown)


compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -50,16 +50,21 @@ impl<'gs, 'ls, 'cd> Validator<'gs, 'ls, 'cd> {
...
     pub fn validate_mod_def(&mut self, _mod_def_id: ModDefId) -> TcResult<()> {
+        // Ensure if it is a trait impl it implements all the trait members.
         todo!()
     }
 
     /// Validate the trait definition of the given [TrtDefId]
     pub fn validate_trt_def(&mut self, _trt_def_id: TrtDefId) -> TcResult<()> {
+        // Ensure Self exists?
+        // @@Design: do we allow traits without self?
         todo!()
```
feds01, line 61: This would be a pretty weird thing to do, but is there something that prevents someone from doing it?

compiler/hash-typecheck/src/error.rs
```diff
@@ -67,6 +67,14 @@ pub enum TcError {
...
         unification_errors: Vec<TcError>,
     },
+    InvalidElementOfMerge {
+        term: TermId,
+    },
+    MergeShouldOnlyContainOneNominal {
+        merge_term: TermId,
+        nominal_term: TermId,
+        second_nominal_term: TermId,
+    },
```
feds01, line 77: I wonder how these will become human-readable messages :)

feds01 (approved):

compiler/hash-typecheck/src/ops/validate.rs
kontheocharis: Not really, it's just a question of when does Self get introduced. Like, if someone types a function as taking Self, does Self automatically get injected implicitly into the impl? Or does it always have to be explicitly stated?

compiler/hash-typecheck/src/error.rs
kontheocharis: If the concept of a merge is known by the user, it shouldn't be difficult to generate understandable error messages for this.

PR #303, opened 2022-06-21
reporting: split logic into modules and fix span highlighting bug

### LINKED ISSUE #299: reporting: Start span start line mis-calculation
In the following error report:
image

The highlighter mistakingly highlights the `9` line number. The span of the expression does not actually begin on the 9th line. This might be a mistake within the lexing implementation or a miscalculation within the report offset to `(col, row)` calculation. It's clear that the line of the span is for some reason being recorded as the end of line 9 where there is a `\n` character at the end. 

Further investigation yielded that the actual span of expression (as printed from the raw source is):
```
loc=134:195, actual="IoError = struct(
    error: IoErrorType,
    message: str,
);"
```

Ideally, the report should not highlight the line number 9, like so:

image

This could be fixed by changing `offset_col_row` by not counting the `\n` in the case of the initial span offset, an implementation could be:
```rust
pub(crate) fn offset_col_row(offset: usize, source: &str, non_inclusive: bool) -> (/* col: */ usize, /* row: */ usize) {
    let source_lines = source.split('\n');

    let mut bytes_skipped = 0;
    let mut total_lines: usize = 0;
    let mut last_line_len = 0;

    let mut line_index = None;
    for (line_idx, line) in source_lines.enumerate() {
        // One byte for the newline
        let skip_width = line.len() + 1;


        // Here, we don't want an inclusive range because we don't want to get the last byte because
        // that will always point 
        let range = if non_inclusive {
            bytes_skipped..bytes_skipped + skip_width
        } else {
            bytes_skipped..bytes_skipped + skip_width + 1
        };

        if range.contains(&offset) {
            line_index = Some((offset - bytes_skipped, line_idx));
            break;
        }

        bytes_skipped += skip_width;
        total_lines += 1;
        last_line_len = line.len();
    }

    line_index.unwrap_or((
        last_line_len.saturating_sub(1),
        total_lines.saturating_sub(1),
    ))
}
```

feds01:
This PR is a minor refactor of the `hash_reporting` crate, primarily designed to split the logic of the reporting into separate and more manageable modules. This PR also fixes a bug where line numbers of the erroneous span would sometimes be highlighted incorrectly. 

Further improvements to the `hash_reporting` crate such as splitting the definition of the various reporting components and the logic to render the reporting components will also be added (but in a separate PR).

closes #299

### DIFF
--- a/compiler/hash-reporting/src/report.rs
+++ b/compiler/hash-reporting/src/report.rs
@@ -13,13 +13,13 @@ use crate::{
     writer::{offset_col_row, ReportCodeBlockInfo, ERROR_MARKING_CHAR},
 };
 
-/// Enumeration describing the kind of [Report]; either being a warning, info or a
+/// Enumeration describing the kind of [Report]; either being a warning, info or an
 /// error.
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq)]
 pub enum ReportKind {
     /// The report is an error.
     Error,
-    /// The report is an informational diagnostic(likely for internal purposes).
+    /// The report is an informational diagnostic (likely for internal purposes).
     Info,
     /// The report is a warning.
     Warning,
@@ -134,9 +134,9 @@ impl ReportCodeBlock {
 
                 let location = self.source_location.span;
 
-                let (start_col, start_row) = offset_col_row(location.start(), source);
-                let (end_col, end_row) = offset_col_row(location.end(), source);
-                let (_, last_row) = offset_col_row(source.len(), source);
+                let (start_col, start_row) = offset_col_row(location.start(), source, true);
+                let (end_col, end_row) = offset_col_row(location.end(), source, false);
+                let (_, last_row) = offset_col_row(source.len(), source, false);
 
                 let (top_buf, bottom_buf) = compute_buffers(start_row, end_row);
                 let indent_width = (start_row.saturating_sub(top_buf) + 1)
--- a/compiler/hash-reporting/src/report.rs
+++ b/compiler/hash-reporting/src/report.rs
@@ -1,28 +1,34 @@
-//! Hash Compiler error and warning reporting module.
-use crate::highlight::{highlight, Colour, Modifier};
-use core::fmt;
-use hash_error_codes::error_codes::HashErrorCode;
-use hash_source::{location::SourceLocation, SourceMap};
-use hash_utils::path::adjust_canonicalization;
 use std::{
     cell::Cell,
+    fmt,
     iter::{once, repeat},
 };
 
-const ERROR_MARKING_CHAR: &str = "^";
+use hash_error_codes::error_codes::HashErrorCode;
+use hash_source::{location::SourceLocation, SourceMap};
+use hash_utils::path::adjust_canonicalization;
 
-/// Enumeration describing the type of report; either being a warning, info or a
+use crate::{
+    highlight::{highlight, Colour, Modifier},
+    writer::{offset_col_row, ReportCodeBlockInfo, ERROR_MARKING_CHAR},
+};
+
+/// Enumeration describing the kind of [Report]; either being a warning, info or a
 /// error.
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq)]
 pub enum ReportKind {
+    /// The report is an error.
     Error,
+    /// The report is an informational diagnostic(likely for internal purposes).
     Info,
+    /// The report is a warning.
     Warning,
+    // @@TODO: Add `Internal` variant: This is an internal compiler error.
 }
 
 impl ReportKind {
     /// Get the [Colour] of the label associated with the [ReportKind].
-    fn as_colour(&self) -> Colour {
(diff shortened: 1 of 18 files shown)


compiler/hash-reporting/src/report.rs
```diff
@@ -1,28 +1,34 @@
...
+use hash_source::{location::SourceLocation, SourceMap};
+use hash_utils::path::adjust_canonicalization;
+
+use crate::{
+    highlight::{highlight, Colour, Modifier},
+    writer::{offset_col_row, ReportCodeBlockInfo, ERROR_MARKING_CHAR},
+};
 
-/// Enumeration describing the type of report; either being a warning, info or a
+/// Enumeration describing the kind of [Report]; either being a warning, info or a
```
kontheocharis, line 16: nitpick `a error` -> `an error`

compiler/hash-reporting/src/report.rs
```diff
@@ -1,28 +1,34 @@
...
 /// error.
 #[derive(Debug, Copy, Clone, Hash, Eq, PartialEq)]
 pub enum ReportKind {
+    /// The report is an error.
     Error,
+    /// The report is an informational diagnostic(likely for internal purposes).
```
kontheocharis, line 22: space before parentheses

compiler/hash-reporting/src/report.rs
```diff
@@ -1,28 +1,34 @@
...
     Info,
+    /// The report is a warning.
     Warning,
+    // @@TODO: Add `Internal` variant: This is an internal compiler error.
```
kontheocharis, line 26: Does this mean that panics essentially create Internal reports and print them when they happen?

compiler/hash-reporting/src/report.rs
```diff
@@ -161,18 +125,20 @@ impl ReportCodeBlock {
...
                 let source = modules.contents_by_id(source_id);
 
                 let location = self.source_location.span;
+                // println!("loc={location}, actual=\"{}\"", &source[location.start()..=location.end()]);
 
-                let (start_col, start_row) = offset_col_row(location.start(), source);
-                let (end_col, end_row) = offset_col_row(location.end(), source);
-                let (_, last_row) = offset_col_row(source.len(), source);
+                let (start_col, start_row) = offset_col_row(location.start(), source, true);
+                // println!("start_row={start_row}");
```
kontheocharis, line 139: Remove

compiler/hash-reporting/src/report.rs
```diff
@@ -161,18 +125,20 @@ impl ReportCodeBlock {
...
-    fn info<T: SourceMap>(&self, modules: &T) -> ReportCodeBlockInfo {
+    pub(crate) fn info<T: SourceMap>(&self, modules: &T) -> ReportCodeBlockInfo {
         match self.info.get() {
             Some(info) => info,
             None => {
                 let source_id = self.source_location.source_id;
                 let source = modules.contents_by_id(source_id);
 
                 let location = self.source_location.span;
+                // println!("loc={location}, actual=\"{}\"", &source[location.start()..=location.end()]);
```
kontheocharis, line 136: Remove

compiler/hash-reporting/src/report.rs
```diff
@@ -252,10 +218,11 @@ impl ReportCodeBlock {
                 index + 1,
                 pad_width = longest_indent_width
             );
+
             writeln!(
                 f,
                 "{} {}   {}",
-                if (start_row..=end_row).contains(&index) {
+                if (start_row..=end_row).contains(&index) && !line.is_empty() {
```
kontheocharis, line 225: Will this not skip over some lines in the file?

compiler/hash-reporting/src/writer.rs
```diff
@@ -0,0 +1,146 @@
...
+    let mut total_lines: usize = 0;
+    let mut last_line_len = 0;
+
+    let mut line_index = None;
+    for (line_idx, line) in source_lines.enumerate() {
+        // One byte for the newline
+        let skip_width = line.len() + 1;
+
+        // Here, we don't want an inclusive range because we don't want to get the last byte because
+        // that will always point
```
kontheocharis, line 49: Will always point to what?

kontheocharis (changes requested):

compiler/hash-reporting/src/report.rs
feds01: No since this is also occurring on the print below as well

compiler/hash-reporting/src/report.rs
kontheocharis: But is the reporter responsible for the abort? Surely it would make more sense if as far as the reporter is concerned Internal  is just another "color mode" (anyway, a cosmetic setting); and the actual abort logic is basically just the panic handler creating an Internal report, printing it, and then aborting right there and then (so maybe the panic handler should add the stack trace as part of the report).

feds01: Yes that is what I was intending, it's purely cosmetic and the panic should occur in the call site.

kontheocharis (approved):

PR #309, opened 2022-06-24
reporting: implement `block` mode for larger spans

### LINKED ISSUE #290: reporting: display large spans in a more elegant way.
Currently, the reporting of errors does not consider that some spans that are highlighted might be large or span multiple lines. IN the event of spanning multiple lines, we want to create a better span annotation in order to reduce the reported span being an eye sore. There are two things that can be done:

- Annotate multi-line spans using a wrapping arrow, here is an example from rust:
```rust
error[E0308]: `match` arms have incompatible types
  --> src/main.rs:8:13
   |
4  |       let _ = match some_moon_phase {
   |  _____________-
5  | |         [1, .., k] => println!("k is {k}"),
   | |                       -------------------- this is found to be of type `()`
6  | |         _ => {
7  | |             println!("ok");
8  | |             2
   | |             ^ expected `()`, found integer
9  | |         }
10 | |     };
   | |_____- `match` arms have incompatible types

```

- Secondly, for large spans, we can skip rendering lines by using `...`, for example:
```rust
error[E0308]: `match` arms have incompatible types
 --> src/main.rs:9:13
  |
4 |     let _ = match some_moon_phase {
  |             --------------------- `match` arms have incompatible types
5 |         [1, .., k] => {
6 |             println!("k is {k}");
  |             -------------------- this is found to be of type `()`
...
9 |             2
  |             ^ expected `()`, found integer
```

feds01:
This patch adds a separation of logic for rendering reports and the actual definitions of the data structures. This patch also introduces modes for printing reports, specifically `block` and `line` modes. The `line` mode is self-explanatory and is used to render reports with single line spans, whereas the `block` mode is used for larger spans.

closes #290 

There is still some work to be done when shortening even larger spans by omitting line spans and replacing the lines with `..` as detailed in #290

### DIFF
--- a/compiler/hash-reporting/src/render.rs
+++ b/compiler/hash-reporting/src/render.rs
@@ -1,11 +1,20 @@
-//! Hash diagnostic report rendering logic.
+//! Hash diagnostic report rendering logic. Within this renderer,
+//! there are currently two modes: `line` and `block`. Within this
+//! module, the modes are often referred to either being line or block.
+//! The difference between the two modes is how they are displayed
+//! when rendered. The `line` mode is designed to highlight spans that
+//! are contained to a single line, whereas `block` mode is designed to
+//! for bulkier spans that might take up many lines. More specific examples
+//! of both modes are located in the documentation of the functions
+//! that are responsible for rendering either mode.
 
 use std::{
     fmt,
-    iter::{once, repeat},
+    iter::{once, repeat, Enumerate, Skip, Take},
+    str::Split,
 };
 
-use hash_source::SourceMap;
+use hash_source::{location::SourceLocation, SourceMap};
 use hash_utils::path::adjust_canonicalization;
 
 use crate::{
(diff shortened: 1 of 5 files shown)


compiler/hash-reporting/src/render.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// Function to compute a row and column number from a given source string
+/// and an offset within the source. This will take into account the number
+/// of encountered newlines and characters per line in order to compute
+/// precise row and column numbers of the span.
+pub(crate) fn offset_col_row(
+    offset: usize,
+    source: &str,
+    non_inclusive: bool,
+) -> (/* col: */ usize, /* row: */ usize) {
```
kontheocharis, line 42: Maybe it might be worth it to make a `ColRowOffset` struct with fields `col` and `row`?

compiler/hash-reporting/src/render.rs
```diff
@@ -0,0 +1,464 @@
...
+            }
+        }
+    }
+
+    /// Function to extract the block of the code that will be used to display the span
+    /// of the diagnostic.
+    fn get_source_view<'a, T: SourceMap>(
+        &self,
+        modules: &'a T,
+    ) -> Take<Skip<Enumerate<Split<'a, char>>>> {
```
kontheocharis, line 149: Is it possible to just write `impl Iterator<Item=&str>` here?

kontheocharis (approved):

compiler/hash-reporting/src/render.rs
feds01: apparently you can't

feds01: Also, it would be `char` not `str` right

PR #310, opened 2022-06-25
Add rustfmt options to group imports and format comments

kontheocharis:
This should hopefully lead to a more consistent code appearance, and shorter import lists.


rustfmt.toml
```diff
@@ -1 +1,3 @@
-single_line_if_else_max_width = 92
```
feds01, line 1: I am a fan of single line if statements, I would personally keep this rule.

feds01 (changes requested):

feds01 (approved):

PR #321, opened 2022-06-28
AST+TC: fix todos within the AST and add basic location information during tc traversal

feds01:
This patch fixes the two outstanding todos within `ast.rs`:
- `VariableExpr` should not have `type_args` on it and any expression which is an `AccessName` followed by type arguments should be parsed as a `TypeFunctionCall`.
- rename `FunctionDefArg` and friends to `FunctionDefParam`

This patch also adds basic information about term locations during traversal, which has led to the first error report within typechecking to be rendered with locations!

### DIFF
--- a/compiler/hash-typecheck/src/ops/building.rs
+++ b/compiler/hash-typecheck/src/ops/building.rs
@@ -10,7 +10,7 @@ use crate::storage::{
     },
     GlobalStorage,
 };
-use hash_source::identifier::Identifier;
+use hash_source::{identifier::Identifier, location::SourceLocation};
 use std::cell::{Cell, RefCell};
 
 /// Helper to create various primitive constructions (from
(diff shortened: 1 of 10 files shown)


compiler/hash-typecheck/src/ops/building.rs
```diff
@@ -530,8 +535,8 @@ impl<'gs> PrimitiveBuilder<'gs> {
         self.create_term(Term::AppTyFn(app_ty_fn))
     }
 
-    /// Release [Self], returning the original [GlobalStorage].
-    pub fn release(self) -> &'gs mut GlobalStorage {
-        self.gs.into_inner()
+    /// Add a [SourceLocation] to a [Term].
+    pub fn add_location_to(&self, subject: TermId, location: SourceLocation) {
```
kontheocharis, line 539: Maybe this should be called `add_location_to_term`?

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -101,11 +110,22 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
+        let location = self.source_location(first_name.span());
+
+        let builder = self.builder();
+        let mut current_term = builder.create_var_term(*first_name.body());
+
+        builder.add_location_to(current_term, location);
+
         for access_name in names {
-            current_term = builder.create_ns_access(current_term, *access_name.body())
+            let name_location = SourceLocation { span: access_name.span(), source_id };
```
kontheocharis, line 125: Can use `self.source_location` here.

feds01: You cannot use `self.source_location` since the builder is being instantiated above and that would mean that it's being borrowed as mutable and the immutable...

PR #325, opened 2022-06-28
TC: Implement declarations and add validation to existing traversals

### LINKED ISSUE #324: TC: Implement traversal of declarations

kontheocharis:
This PR also adds formatting for scopes, which should help in debugging!
Declaration traversal is not yet fully complete, because it does not deal with mutability or visibility (some checks need to be implemented for this). Furthermore, the only patterns supported right now are binding patterns.

Closes #324.

### DIFF
--- a/compiler/hash-typecheck/src/fmt.rs
+++ b/compiler/hash-typecheck/src/fmt.rs
@@ -2,14 +2,29 @@
 //! debug output.
 use crate::storage::{
     primitives::{
-        ArgsId, EnumDef, Level0Term, Level1Term, Level2Term, Level3Term, ModDefId, ModDefOrigin,
-        NominalDef, NominalDefId, ParamsId, StructDef, SubSubject, Term, TermId, TrtDefId,
-        UnresolvedTerm,
+        ArgsId, EnumDef, Level0Term, Level1Term, Level2Term, Level3Term, MemberData, ModDefId,
+        ModDefOrigin, Mutability, NominalDef, NominalDefId, ParamsId, ScopeId, StructDef,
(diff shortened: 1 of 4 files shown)


compiler/hash-typecheck/src/fmt.rs
```diff
@@ -33,6 +48,66 @@ impl<'gs> TcFormatter<'gs> {
...
+        let scope_value_fmt_opts = TcFormatOpts { expand: true, ..TcFormatOpts::default() };
+
+        for member in scope.iter() {
+            let mutability = match member.mutability {
+                Mutability::Mutable => "mut ",
+                Mutability::Immutable => "",
+            };
+            let visibility = match member.visibility {
+                Visibility::Public => "pub ",
+                Visibility::Private => "pri ",
```
feds01, line 64:
```suggestion
                Visibility::Private => "priv ",
```

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -688,14 +702,19 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
             // warn if it isn't?
         }
 
         // Traverse the ending expression, if any, or return void.
         match &node.expr {
-            Some(expr) => self.visit_expression(ctx, expr.ast_ref()),
+            Some(expr) => {
+                let expr_id = self.visit_expression(ctx, expr.ast_ref())?;
+                self.validator().validate_term(expr_id)?;
+                Ok(expr_id)
```
feds01, line 716: should return simplified id

Issue #345, opened 2022-06-30
Bug: Type function types always fail to parse


kontheocharis:
The following snippet:

```rust
x: <T: Type> -> Type
```

fails to parse. This is also the case for any other type function type, as seen in the screenshot below:

![Screenshot 2022-06-30 at 19 10 26](https://user-images.githubusercontent.com/18248100/176748111-ae5a10b2-e99e-42e9-b320-2d195a174cca.png)

PR #347, opened 2022-06-30
TC: Implement traversal of type functions

### LINKED ISSUE #341: TC: Traverse type functions

kontheocharis:
This PR implements AST traversal and typechecking of type functions and type function types. It also fixes the previous implementation of validation/simplification for type functions, to enter parameter scope before validating the return type and value.

Closes #341.


kontheocharis:
Note that type function type traversal cannot currently be tested due to #345.

feds01 (approved):

PR #349, opened 2022-07-01
pipeline: re-organise sources structure

### LINKED ISSUE #306: ErrorReporting: Add panic macros that take Term/Span as argument and print a report

feds01:
This PR attempts to separate `SourceMap` from `Sources` so that it can be referenced independently of `Sources`. This will come in handy when wanting to simply report without giving the whole `SourceMap` to the relevant functions. This also adds missing documentation to relevant modules that store the source map implementation.

closes #306


kontheocharis (approved):

compiler/hash-parser/src/import_resolver.rs
```diff
@@ -3,42 +3,65 @@
...
-pub struct ImportResolver {
+/// The [ImportResolver] contains internal logic for resolving the path
+/// and contents of a module import, in order to prepare it for lexing
+/// and parsing.
+pub struct ImportResolver<'p> {
+    /// The associated [SourceId] with the import resolution.
     source_id: SourceId,
-    root_dir: PathBuf,
+    /// Working directory from where the import path resolution occurs.
+    root_dir: &'p PathBuf,
```
kontheocharis, line 18: Can use Path here

compiler/hash-reporting/src/macros.rs
```diff
@@ -0,0 +1,41 @@
...
+            let mut report = $crate::builder::ReportBuilder::new();
+            report
+                .with_kind($crate::report::ReportKind::Internal)
+                .with_message(message)
+                .add_element($crate::report::ReportElement::CodeBlock($crate::report::ReportCodeBlock::new(
+                    $location,
+                    "here",
+                )));
+
+            println!("{}", $crate::writer::ReportWriter::new(report.build(), $sources));
```
kontheocharis, line 22: Maybe this should go to stderr?

compiler/hash-typecheck/src/diagnostics/macros.rs
```diff
@@ -0,0 +1,137 @@
...
+
+            if let Some(location) = term_location {
+                builder
+                    .add_element(hash_reporting::report::ReportElement::CodeBlock(hash_reporting::report::ReportCodeBlock::new(
+                        location,
+                        "here",
+                    )));
+            }
+
+            println!("{}", hash_reporting::writer::ReportWriter::new(report.build(), sources));
```
kontheocharis, line 40: See stderr comment above

kontheocharis: Also, shouldn't this also print the term itself?

feds01: Should this always happen, or should it be controlled by the formatting arguments?

kontheocharis: I think it should always happen, the user of the macro can always print more terms if they want, but it should print the "subject" term of the panic i think (just like it prints its location).

kontheocharis (approved):

PR #360, opened 2022-07-03
TC: Traverse function calls and constructors

### LINKED ISSUE #328: TC: Traverse function calls and constructors

kontheocharis:
This PR closes #328. It also renames `RuntimeInstantiable` to just `Type` and makes it the default for type parameters. Furthermore, this PR implements `Clone` for all AST nodes, where before this wasn't the case for some reason..

### DIFF
--- a/compiler/hash-typecheck/src/ops/validate.rs
+++ b/compiler/hash-typecheck/src/ops/validate.rs
@@ -323,12 +323,6 @@ impl<'gs, 'ls, 'cd, 's> Validator<'gs, 'ls, 'cd, 's> {
         let reader = self.reader();
         let merge_element_term = reader.get_term(merge_element_term_id);
 
-        println!(
-            "Validating element {} inside merge {}",
-            self.for_fmt(merge_element_term_id),
-            self.for_fmt(merge_term_id)
-        );
-
         // Error helper:
         let invalid_merge_element = || -> TcResult<()> {
             Err(TcError::InvalidMergeElement { term: merge_element_term_id })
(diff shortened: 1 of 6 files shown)


compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -323,6 +323,12 @@ impl<'gs, 'ls, 'cd, 's> Validator<'gs, 'ls, 'cd, 's> {
         let reader = self.reader();
         let merge_element_term = reader.get_term(merge_element_term_id);
 
+        println!(
```
feds01, line 326: Remove print statement

compiler/hash-typecheck/src/storage/core.rs
```diff
@@ -87,7 +87,7 @@ impl CoreDefs {
 
         // Any type
         let any_ty = builder.create_any_ty_term();
-        builder.add_pub_member_to_scope("Type", builder.create_trt_kind_term(), any_ty);
+        builder.add_pub_member_to_scope("AnyType", builder.create_trt_kind_term(), any_ty);
```
feds01, line 90: Maybe just `Any`

feds01 (changes requested):

compiler/hash-typecheck/src/storage/core.rs
kontheocharis: I dont think thats a good idea, since `Trait` for example is not assignable to `AnyType`, and `Any` sounds like it should be anything,

feds01 (approved):

Issue #363, opened 2022-07-08
TC: Internally replace enums with unions of structs.


kontheocharis:

kontheocharis:
- [ ] test

PR #364, opened 2022-07-08
TC: Introduce unions and the "never" type

### LINKED ISSUE #357: TC: Add support for unions and never

kontheocharis:
This is needed to tackle #363 as well as for the "never" type which was missing up until now. In many ways, unions are dual to merges (and to tuples in other contexts). Currently only nominals can be "unioned", to support enum definitions as unions of structs. In the future this might change.

Closes #357.

### DIFF
--- a/compiler/hash-typecheck/Cargo.toml
+++ b/compiler/hash-typecheck/Cargo.toml
@@ -7,11 +7,12 @@ edition = "2021"
 [dependencies]
 slotmap = "1.0"
 dashmap = "5.1"
+itertools = "0.10"
 
 hash-ast = { path = "../hash-ast" }
 hash-source = { path = "../hash-source" }
 hash-utils = { path = "../hash-utils" }
 hash-alloc = { path = "../hash-alloc" }
-hash-reporting = {path = "../hash-reporting" }
-hash-pipeline = {path = "../hash-pipeline" }
-hash-error-codes = {path = "../hash-error-codes" }
+hash-reporting = { path = "../hash-reporting" }
+hash-pipeline = { path = "../hash-pipeline" }
+hash-error-codes = { path = "../hash-error-codes" }
--- a/compiler/hash-typecheck/src/diagnostics/reporting.rs
+++ b/compiler/hash-typecheck/src/diagnostics/reporting.rs
@@ -481,6 +481,25 @@ impl<'gs, 'ls, 'cd, 's> From<TcErrorWithStorage<'gs, 'ls, 'cd, 's>> for Report {
                     // used within this position
                 }
             }
+            TcError::InvalidUnionElement { term } => {
+                builder
+                    .with_error_code(HashErrorCode::InvalidUnionElement)
+                    .with_message("invalid element within a union");
+
+                if let Some(location) = err.location_store().get_location(term) {
+                    builder.add_element(ReportElement::CodeBlock(ReportCodeBlock::new(
+                        location,
+                        format!(
+                            "cannot use the type `{}` within a union",
+                            term.for_formatting(err.global_storage()),
+                        ),
+                    )));
+
+                    // @@Todo(feds01): add more helpful information about why
+                    // this particular type cannot be
+                    // used within this position
+                }
+            }
             TcError::InvalidTypeFunctionParameterType { param_ty } => {
                 builder
                     .with_error_code(HashErrorCode::DisallowedType)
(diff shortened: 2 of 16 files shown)


compiler/hash-typecheck/Cargo.toml
```diff
@@ -7,11 +7,12 @@ edition = "2021"
 [dependencies]
 slotmap = "1.0"
 dashmap = "5.1"
+itertools = "0.10"
```
feds01, line 10: What is this dependency for?

compiler/hash-typecheck/src/diagnostics/reporting.rs
```diff
@@ -481,6 +481,25 @@ impl<'gs, 'ls, 'cd, 's> From<TcErrorWithStorage<'gs, 'ls, 'cd, 's>> for Report {
...
+                if let Some(location) = err.location_store().get_location(term) {
+                    builder.add_element(ReportElement::CodeBlock(ReportCodeBlock::new(
+                        location,
+                        format!(
+                            "cannot use the type `{}` within a union",
+                            term.for_formatting(err.global_storage()),
+                        ),
+                    )));
+
+                    // @@Todo(feds01): add more helpful information about why
```
feds01, line 498: maybe make this an issue?

compiler/hash-typecheck/src/ops/simplify.rs
```diff
@@ -953,6 +974,85 @@ impl<'gs, 'ls, 'cd, 's> Simplifier<'gs, 'ls, 'cd, 's> {
...
+                    if self.unifier().terms_are_equal(first, second) {
+                        simplified_once = true;
+                        merged[*first_idx] = merged[*first_idx].map(|_| first);
+                        merged[*second_idx] = None;
+                    } else {
+                        merged[*first_idx] = merged[*first_idx].map(|_| first);
+                        merged[*second_idx] = merged[*second_idx].map(|_| second);
+                    }
+                }
+                _ => unreachable!(),
```
feds01, line 1043: maybe make this a `tc_panic`?

compiler/hash-typecheck/src/ops/simplify.rs
```diff
@@ -953,6 +974,85 @@ impl<'gs, 'ls, 'cd, 's> Simplifier<'gs, 'ls, 'cd, 's> {
...
+            }
+        }
+        let result: Vec<_> = merged.into_iter().flatten().collect();
+
+        // Only return if it has been simplified
+        if !simplified_once {
+            Ok(None)
+        } else {
+            Ok(Some(result))
+        }
```
feds01, line 1053: does this need to copy locations for anything?

compiler/hash-typecheck/src/ops/unify.rs
```diff
@@ -278,6 +284,44 @@ impl<'gs, 'ls, 'cd, 's> Unifier<'gs, 'ls, 'cd, 's> {
...
+                for inner_target_id in inner_target {
+                    match self.unify_terms(simplified_src_id, inner_target_id) {
+                        Ok(result) => {
+                            return Ok(result);
+                        }
+                        Err(e) if first_error.is_none() => {
+                            first_error = Some(e);
+                            continue;
+                        }
+                        _ => continue,
```
feds01, line 301: maybe use `Option::or` here? It seems to be a nice fit for this, [this](https://doc.rust-lang.org/std/option/enum.Option.html#method.or) is the reference to the API

feds01 (changes requested):

compiler/hash-typecheck/Cargo.toml
kontheocharis: `simplify.rs` line 1030

compiler/hash-typecheck/src/diagnostics/reporting.rs
kontheocharis: Opened #366.

compiler/hash-typecheck/src/ops/simplify.rs
kontheocharis: Well this is not a bug in the typechecker if this happens, it is just an impossible situation because of `.combinations(2)` above. I think using `tc_panic` would be overkill.

compiler/hash-typecheck/src/ops/simplify.rs
kontheocharis: No, this will occur in `simplify_term` below.

feds01 (approved):

PR #365, opened 2022-07-10
tc: Implement nominal definition traversal and validation

feds01:


compiler/hash-ast-passes/src/diagnostics/mod.rs
```diff
@@ -75,6 +76,34 @@ impl Display for BlockOrigin {
     }
 }
 
+/// Denotes where an error occurred in regards to a field within a defined
+/// structural type which contains fields, such as `struct`, `tuple`, or
+/// `function literal`.
```
kontheocharis, line 81: How does a function literal have fields? You mean function parameters?

compiler/hash-typecheck/src/diagnostics/params.rs
```diff
@@ -53,6 +53,29 @@ impl ParamListKind {
...
+        match self {
+            ParamListKind::Params(id) => store.params_store.get(*id).names(),
+            ParamListKind::Args(id) => store.args_store.get(*id).names(),
+        }
+    }
+
+    /// Function used to compute the missing fields from another
+    /// [ParamListKind]. This does not compute a difference as it doesn't
+    /// consider items that are present in the other [ParamListKind] and not
+    /// in the current list as `missing`.
```
kontheocharis, line 68: It does compute a difference, just not a symmetric difference.

compiler/hash-typecheck/src/ops/substitute.rs
```diff
@@ -573,6 +579,27 @@ impl<'gs, 'ls, 'cd, 's> Substituter<'gs, 'ls, 'cd, 's> {
...
+        let mut result = HashSet::new();
+        self.add_free_vars_in_term_to_set(term_id, &mut result);
+
+        result
+            .into_iter()
+            .map(|var| match var {
+                SubSubject::Var(var) => Ok(var),
+                SubSubject::Unresolved(_) => Err(TcError::UnresolvedVariable {
+                    // @@Correctness: it's unclear if we can even get an identifier here? Or if we
+                    // should make an identifier from the provided `ResolutionId`
```
kontheocharis, line 596: You can make a string by `self.for_fmt(var.into())` and then create an identifier from it.

compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -303,10 +303,58 @@ impl<'gs, 'ls, 'cd, 's> Validator<'gs, 'ls, 'cd, 's> {
...
+                Ok(())
+            }
+            NominalDef::Enum(enum_def) => {
+                for (_, variant) in enum_def.variants.clone().iter() {
+                    let variant_fields = self.params_store().get(variant.fields).clone();
+
+                    // Validate the ordering and the number of times parameter field names
+                    // are specified, although the ordering shouldn't matter
+                    //
+                    // @@Unnecessary?
```
kontheocharis, line 338: what do you mean by the ordering shouldn't matter?

compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -303,10 +303,58 @@ impl<'gs, 'ls, 'cd, 's> Validator<'gs, 'ls, 'cd, 's> {
...
+                    validate_param_list_ordering(
+                        &variant_fields,
+                        ParamListKind::Params(variant.fields),
+                    )?;
+
+                    // Validate all fields of an struct def implement `RtInstantiable`
```
kontheocharis, line 344: *enum def variant

compiler/hash-typecheck/src/storage/primitives.rs
```diff
@@ -110,14 +118,19 @@ impl Scope {
...
+        let index = self.members.len() - 1;
+
+        self.member_names.insert(member.name, index);
+        index
     }
 
     /// Get a member by name.
-    pub fn get(&self, member_name: Identifier) -> Option<Member> {
-        Some(self.members[self.member_names.get(&member_name).copied()?])
+    pub fn get(&self, member_name: Identifier) -> Option<(Member, usize)> {
```
kontheocharis, line 130: Can return a scope member here.

compiler/hash-typecheck/src/storage/primitives.rs
```diff
@@ -110,14 +118,19 @@ impl Scope {
 
     /// Add a member to the scope, overwriting any existing member with the same
     /// name.
-    pub fn add(&mut self, member: Member) {
+    pub fn add(&mut self, member: Member) -> usize {
```
kontheocharis, line 121: Can return the scope member here.

compiler/hash-typecheck/src/storage/primitives.rs
```diff
@@ -317,8 +335,11 @@ pub struct EnumVariant {
 /// An enum definition, containing a binding name and a set of variants.
 #[derive(Debug, Clone)]
 pub struct EnumDef {
+    /// The name of the `EnumDef`, useful for error reporting
     pub name: Option<Identifier>,
+    /// Any free variables that occur within the `variants` of the [EnumDef].
```
kontheocharis, line 340: *which are part of higher bounds.

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -1488,44 +1512,157 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
-        _node: hash_ast::ast::AstNodeRef<hash_ast::ast::StructDefEntry>,
+        ctx: &Self::Ctx,
+        node: hash_ast::ast::AstNodeRef<hash_ast::ast::StructDefEntry>,
     ) -> Result<Self::StructDefEntryRet, Self::Error> {
-        todo!()
+        let walk::StructDefEntry { name, ty, default } =
+            walk::walk_struct_def_entry(self, ctx, node)?;
+
+        // Try and figure out a known term...
+        let (ty, default_value) = match (ty, default) {
```
kontheocharis, line 1526: Maybe add a @@Repetition to factor this into a function since it is also used for declarations and tuples.

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -1488,44 +1512,157 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
+            (Some(annotation_ty), Some(default_value)) => {
+                let default_ty = self.typer().ty_of_term(default_value)?;
+
+                // Here, we have to unify both of the provided types...
+                let sub = self.unifier().unify_terms(default_ty, annotation_ty)?;
+
+                // @@Naming
+                let ds_sub = self.substituter().apply_sub_to_term(&sub, default_value);
+                let as_sub = self.substituter().apply_sub_to_term(&sub, annotation_ty);
```
kontheocharis, line 1535: these names could be improved.

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -1488,44 +1512,157 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
-        _ctx: &Self::Ctx,
-        _node: hash_ast::ast::AstNodeRef<hash_ast::ast::EnumDef>,
+        ctx: &Self::Ctx,
+        node: hash_ast::ast::AstNodeRef<hash_ast::ast::EnumDef>,
     ) -> Result<Self::EnumDefRet, Self::Error> {
-        todo!()
+        let walk::EnumDef { entries } = walk::walk_enum_def(self, ctx, node)?;
+
+        // take the declaration hint here...
+        let name = self.state.declaration_name_hint.take();
```
kontheocharis, line 1644: This also needs to be done for type functions

compiler/hash-utils/src/printing.rs
```diff
@@ -3,40 +3,92 @@
...
+}
+
+impl SequenceJoinKind {
+    pub fn as_conjunctive(&self) -> &str {
+        match self {
+            SequenceJoinKind::Either => "or",
+            SequenceJoinKind::All => "and",
+        }
+    }
+}
```
kontheocharis, line 26: Maybe this would be better as `impl Display`?

tests/parser/tests/test_runner.rs
```diff
@@ -15,8 +15,14 @@ use hash_utils_testing_macros::generate_tests;
...
 
+extern crate static_assertions;
+
 /// Whether or not the UI tests should re-generate the output.
-const REGENERATE_OUTPUT: bool = true;
+const REGENERATE_OUTPUT: bool = false;
+
+// Assert that `REGENERATE_OUTPUT` isn't set to `true` in normal running
+// conditions
+static_assertions::const_assert!(!REGENERATE_OUTPUT);
```
kontheocharis, line 25: How is this not hit when one wants to regenerate output?

kontheocharis (changes requested):

tests/parser/tests/test_runner.rs
feds01: This should hit it and alert that the tests are set to `pass` all the time, this essentially prevents it from being set to `true` when it really shouldn't

kontheocharis: But in order to regenerate output one has to comment out this line first?

compiler/hash-typecheck/src/ops/validate.rs
feds01: Since the parameters within the enum cannot be named I believe

kontheocharis: Surely the enum variant parameters have identical semantics to struct parameters?

tests/parser/tests/test_runner.rs
feds01: Ah yes of course, okay Ill put the assertion at the end of the tests

compiler/hash-ast-passes/src/diagnostics/mod.rs
feds01: Yes, ill update the wording

compiler/hash-typecheck/src/ops/substitute.rs
feds01: There is no implementation for `UnresolvedTerm`, should I implement one in this pr?

compiler/hash-typecheck/src/storage/primitives.rs
feds01: There is no access `scope_id` which `ScopeMember` requires

compiler/hash-typecheck/src/storage/primitives.rs
feds01: There is no access scope_id which ScopeMember requires

compiler/hash-typecheck/src/ops/validate.rs
feds01: Ah okay, ill remove the comment

kontheocharis (approved):

Issue #367, opened 2022-07-10
TC: Handle mutually recursive constant definitions, including cyclic imports


kontheocharis:

kontheocharis:
#419  is due to this.

kontheocharis:
Things to be done:
- Handle mutually recursive definitions
- Handle reassignment
- Handle immutable reassignment (i.e. merges)

PR #368, opened 2022-07-10
TC: Implement traversal for imports

### LINKED ISSUE #335: TC: Traverse imports

kontheocharis:
This includes marking sources as already checked. However, this does not handle cyclic imports yet, and currently the module resolution will loop forever. This should be tackled in a separate PR once we figure out how to deal with recursive definitions (see #367). 

Closes #335.

### DIFF
--- a/compiler/hash-typecheck/src/traverse/mod.rs
+++ b/compiler/hash-typecheck/src/traverse/mod.rs
@@ -4,7 +4,10 @@
 use std::collections::HashSet;
 
 use crate::{
-    diagnostics::error::{TcError, TcResult},
+    diagnostics::{
+        error::{TcError, TcResult},
+        macros::tc_panic,
+    },
     ops::{validate::TermValidation, AccessToOpsMut},
     storage::{
         location::LocationTarget,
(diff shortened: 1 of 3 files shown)


compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -685,10 +695,56 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
+                        let module_term = child_visitor.visit_source()?;
+                        Ok(module_term)
+                    }
+                }
+            }
+            None => {
+                // This should never happen because all modules should have been parsed by now.
+                let unresolved_import_term = self.builder().create_unresolved_term();
+                let import_location = self.source_location(node.span());
+                self.builder().add_location_to_target(unresolved_import_term, import_location);
```
feds01, line 740: does this print the source of the import that is requesting the module?

feds01 (approved):

compiler/hash-typecheck/src/traverse/mod.rs
kontheocharis: yes

PR #370, opened 2022-07-11
tc: implement trait definition traversal

feds01:
This PR brings us another step closer to finishing implementing `mod`/`trait def`/`trait impl`s as per #340

### DIFF
--- a/compiler/hash-typecheck/src/ops/building.rs
+++ b/compiler/hash-typecheck/src/ops/building.rs
@@ -432,10 +432,11 @@ impl<'gs> PrimitiveBuilder<'gs> {
         bound_vars: impl IntoIterator<Item = Var>,
     ) -> TrtDefId {
         let name = trait_name.map(|t| t.into());
+        let members = self.create_constant_scope(members);
 
         let trt_def_id = self.gs.borrow_mut().trt_def_store.create(TrtDef {
             name,
-            members: self.create_constant_scope(members),
+            members,
             bound_vars: bound_vars.into_iter().collect(),
         });
         let trt_def_ty = self.create_trt_kind_term();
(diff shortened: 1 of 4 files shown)


compiler/hash-typecheck/src/ops/building.rs
```diff
@@ -424,39 +424,41 @@ impl<'gs> PrimitiveBuilder<'gs> {
...
+        trait_name: Option<impl Into<Identifier>>,
+        members: impl IntoIterator<Item = Member>,
         bound_vars: impl IntoIterator<Item = Var>,
     ) -> TrtDefId {
+        let name = trait_name.map(|t| t.into());
+
         let trt_def_id = self.gs.borrow_mut().trt_def_store.create(TrtDef {
-            name: None,
+            name,
             members: self.create_constant_scope(members),
```
kontheocharis, line 438: Needs to happen on a separate line before the borrow_mut

compiler/hash-typecheck/src/ops/building.rs
```diff
@@ -424,39 +424,41 @@ impl<'gs> PrimitiveBuilder<'gs> {
...
-        trait_name: impl Into<Identifier>,
-        members: impl IntoIterator<Item = Member>,
+        members: impl Iterator<Item = Member>,
         bound_vars: impl IntoIterator<Item = Var>,
     ) -> TrtDefId {
-        let name = trait_name.into();
-        let members = self.create_constant_scope(members);
         let trt_def_id = self.gs.borrow_mut().trt_def_store.create(TrtDef {
-            name: Some(name),
-            members,
```
kontheocharis, line 454: Needs to happen on a separate line before the borrow_mut

Issue #362, opened 2022-07-08
Parser: Add support for unions in types


kontheocharis:
Union syntax (in types) is the same as merging syntax, but using `|` instead of `~`. Union should have higher precedence than merge.

PR #380, opened 2022-07-12
parser+ast: support for `union` type notation

feds01:
This patch adds support within the AST and parser for
union type notation. The patch introduces the notion
of type operators which are equivalent to binary operators
but for types. Currently, `TyOp` has two variants which
are `Union` and `Merge`, where union operator has a
higher precedence then the merge operator.

closes [#362](https://github.com/hash-org/lang/issues/362)

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -397,9 +397,39 @@ pub struct TyFnCall {
 }
 
 /// A merged type meaning that multiple types are considered to be
-/// specified in place of one, e.g. `Conv ~ Eq ~ Print`
+/// specified in place of one, e.g. `Conv ~ Eq`
 #[derive(Debug, PartialEq, Clone)]
-pub struct MergedTy(pub AstNodes<Ty>);
+pub struct MergedTy {
+    pub lhs: AstNode<Ty>,
+    pub rhs: AstNode<Ty>,
+}
+
+/// A merged type meaning that multiple types are considered to be
+/// specified in place of one, e.g. `f64 | i64`
+#[derive(Debug, PartialEq, Clone)]
+pub struct UnionTy {
+    pub lhs: AstNode<Ty>,
+    pub rhs: AstNode<Ty>,
+}
+
+/// Binary type operators enumeration.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum TyOp {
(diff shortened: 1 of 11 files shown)


compiler/hash-ast/src/ast.rs
```diff
@@ -397,9 +397,39 @@ pub struct TyFnCall {
...
+/// specified in place of one, e.g. `f64 | i64`
+#[derive(Debug, PartialEq, Clone)]
+pub struct UnionTy {
+    pub lhs: AstNode<Ty>,
+    pub rhs: AstNode<Ty>,
+}
+
+/// Binary type operators enumeration.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub enum TyOp {
```
kontheocharis, line 417: Maybe a better name is BinaryTyOp?

compiler/hash-ast/src/visitor.rs
```diff
@@ -4372,6 +4416,7 @@ pub mod walk_mut {
             Ty::Named(r) => r,
             Ty::Ref(r) => r,
             Ty::Merged(r) => r,
```
kontheocharis, line 4418: Should this possibly be called Merge rather than Merged?

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -1075,16 +1075,35 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
...
+    type UnionTyRet = TermId;
+
+    fn visit_union_ty(
+        &mut self,
+        ctx: &Self::Ctx,
+        node: hash_ast::ast::AstNodeRef<hash_ast::ast::UnionTy>,
+    ) -> Result<Self::UnionTyRet, Self::Error> {
+        let walk::UnionTy { lhs, rhs } = walk::walk_union_ty(self, ctx, node)?;
+
+        // @@Todo: do we need to potentially flatten the lhs and rhs?
```
kontheocharis, line 1098: This will happen in simplification :)

compiler/hash-ast/src/ast.rs
feds01: I think to follow the same naming conventions as for expressions: `BinTyOp` is appropriate

kontheocharis (approved):

PR #382, opened 2022-07-12
ast+parser: introduce `merge_expr` variant

### LINKED ISSUE #377: Parser: Add support for merge expressions on the expression level
Currently they are only implemented on the type level.

feds01:
This patch adds the `Merge` variant which allows for merge types to be expressed on the `Expression` level.

closes #377


compiler/hash-ast/src/ast.rs
```diff
@@ -791,6 +791,16 @@ pub struct MergeDeclaration {
...
+/// A merge expression.
+#[derive(Debug, PartialEq, Clone)]
+pub struct MergeExpr {
+    /// The left-hand side expression of the merge expr.
+    pub lhs: AstNode<Expression>,
+
+    /// The right-hand side expression of the merge expr.
+    pub rhs: AstNode<Expression>,
+}
+
```
kontheocharis, line 803: Why is this needed if BinaryExpr exists and there is a BinaryOp::Merge?

kontheocharis (approved):

PR #392, opened 2022-07-14
Prelude injection

feds01:
This patch expands the pipeline to automatically inject the `prelude`.  This also implements semantic checks for using the `intrinsic` directive 

Additionally, this fixes several issues within the typechecker.

### DIFF
--- a/compiler/hash-ast-passes/src/analysis/mod.rs
+++ b/compiler/hash-ast-passes/src/analysis/mod.rs
@@ -2,6 +2,7 @@
 //! definition with some shared functions to append diagnostics to the analyser.
 
 use crossbeam_channel::Sender;
+use hash_pipeline::sources::ModuleKind;
 use hash_source::{
     location::{SourceLocation, Span},
     SourceId, SourceMap,
@@ -9,8 +10,9 @@ use hash_source::{
 
 use crate::diagnostics::{
     error::{AnalysisError, AnalysisErrorKind},
+    origins::BlockOrigin,
     warning::{AnalysisWarning, AnalysisWarningKind},
-    BlockOrigin, Diagnostic,
+    Diagnostic,
 };
 
 mod block;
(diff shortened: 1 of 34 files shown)


compiler/hash-ast-passes/src/analysis/mod.rs
```diff
@@ -32,11 +34,17 @@ pub struct SemanticAnalyser<'s> {
...
+    pub(crate) module_kind: Option<ModuleKind>,
 }
 
 impl<'s> SemanticAnalyser<'s> {
     /// Create a new semantic analyser
-    pub fn new(source_map: &'s SourceMap, source_id: SourceId) -> Self {
+    pub fn new(
+        source_map: &'s SourceMap,
+        source_id: SourceId,
+        module_kind: Option<ModuleKind>,
```
kontheocharis, line 46: SourceMap should have a field that maps ModuleId to ModuleKind, instead of having to pass in this parameter. And then this can be looked up based on source_id.

compiler/hash-pipeline/src/lib.rs
```diff
@@ -377,6 +377,28 @@ where
...
+            compiler_state,
+            CompilerJobParams::default(),
+        );
+
+        // @@Hack: this returns an empty error list, but it just signals to the outside
+        // world to abort...
+        match compiler_state.job_has_error {
+            false => Ok(compiler_state),
+            true => Err(vec![]),
+        }
```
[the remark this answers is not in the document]
kontheocharis: This could be a panic and then you don't need the job_has_error flag?

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -492,10 +497,21 @@ impl<'gs, 'ls, 'cd, 'src> visitor::AstVisitor for TcVisitor<'gs, 'ls, 'cd, 'src>
         ctx: &Self::Ctx,
         node: hash_ast::ast::AstNodeRef<hash_ast::ast::DirectiveExpr>,
     ) -> Result<Self::DirectiveExprRet, Self::Error> {
+        let old_intrinsics_state = self.state.within_intrinsics_directive;
+
+        // If the current specified directive is `intrinsics`, then we have to enable
+        // the flag `within_intrinsics_directive` which changes the way that `mod`
+        // blocks are validated and there members appended to the scope
```
kontheocharis, line 504: `changes the parsing of the declarations inside the mod block`

kontheocharis (changes requested):

kontheocharis (approved):

PR #395, opened 2022-07-15
tc: Implement traversal of `binary` and `unary` expressions

### LINKED ISSUE #334: TC: Traverse binary/unary operations

feds01:
This PR closes #334

### DIFF
--- a/compiler/hash-parser/src/parser/expression.rs
+++ b/compiler/hash-parser/src/parser/expression.rs
@@ -388,8 +388,6 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
                     // if the operator is a non-functional, (e.g. as) we need to perform a different
                     // conversion where we transform the AstNode into a
                     // different
-                    // @@Todo: Use `ExpressionKind::TypeExpr` rather than relying on
-                    // `ExpressionKind::CastExpr`...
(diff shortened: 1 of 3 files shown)


compiler/hash-parser/src/parser/expression.rs
```diff
@@ -388,6 +388,8 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
                     // if the operator is a non-functional, (e.g. as) we need to perform a different
                     // conversion where we transform the AstNode into a
                     // different
+                    // @@Todo: Use `ExpressionKind::TypeExpr` rather than relying on
+                    // `ExpressionKind::CastExpr`...
```
kontheocharis, line 392: Maybe this would be a bit annoying to do ?

kontheocharis (approved):

compiler/hash-parser/src/parser/expression.rs
kontheocharis: Yeah, maybe it should just be left like this?

feds01: Yes, ill remove this comment and then merge

kontheocharis (approved):

Issue #397, opened 2022-07-15
parser: re-work the parser to use `AccessExpr` to represent namespacing


feds01:

kontheocharis:
Closed by #404

PR #404, opened 2022-07-16
ast+parser:  Introduce `access` expr

### LINKED ISSUE #402: parser: rename all `expression`s to `expr`
It seems that 'expression' and 'expr' are used interchangeably across the compiler sources. In an effort to standardise naming conventions, `Expr` will become the standard rather than `Expression`.

feds01:
This patch unifies namespace access and property access expressions to be of one expression type. This simplifies the handling logic within the typechecker and will go on to make #403 implementable.

Closes #402 #397


compiler/hash-ast-desugaring/src/desugaring.rs
```diff
@@ -60,10 +59,10 @@ impl<'s> AstDesugaring<'s> {
 
         let (iter_span, pat_span, body_span) = (iterator.span(), pattern.span(), body.span());
 
-        let make_access_name = |label: &str| -> AstNode<AccessName> {
+        let make_access_name = |label: &str| -> AstNode<Namespace> {
             // Create the identifier within the map...
             AstNode::new(
-                AccessName { path: ast_nodes![AstNode::new(label.into(), iter_span)] },
+                Namespace { path: ast_nodes![AstNode::new(label.into(), iter_span)] },
```
kontheocharis, line 65: Maybe AccessPattern instead? Which takes an inner pattern and a name..

kontheocharis: Next pr.

kontheocharis (approved):

PR #411, opened 2022-07-16
TC: Implement pattern matching, as well as pattern and control flow traversal

### LINKED ISSUE #410: TC: Basic pattern matching

kontheocharis:
This PR implements pattern and control flow traversal. There are still a few things to be done:
- Adding constructor pattern bindings to scope: #407 
- Matching or patterns: #409 
- Matching mod patterns: #408 

Closes #410, #331, #329.

### DIFF
--- a/compiler/hash-typecheck/src/diagnostics/error.rs
+++ b/compiler/hash-typecheck/src/diagnostics/error.rs
@@ -16,9 +16,19 @@ pub type TcResult<T> = Result<T, TcError>;
 pub enum TcError {
     /// Cannot unify the two terms.
     CannotUnify { src: TermId, target: TermId },
+    /// Cannot unify the two argument lists. This can occur if the names
+    /// don't match of the arguments or if the number of arguments isn't the
+    /// same.
+    CannotUnifyArgs {
+        src_args_id: ArgsId,
+        target_args_id: ArgsId,
+        src: TermId,
+        target: TermId,
+        reason: ParamUnificationErrorReason,
+    },
     /// Cannot unify the two parameter lists. This can occur if the names
     /// don't match of the parameters or if the number of parameters isn't the
-    /// same.
+    /// same, or the types mismatch.
     CannotUnifyParams {
         src_params_id: ParamsId,
         target_params_id: ParamsId,
--- a/compiler/hash-typecheck/src/diagnostics/error.rs
+++ b/compiler/hash-typecheck/src/diagnostics/error.rs
@@ -1,6 +1,9 @@
 //! Error-related data structures for errors that occur during typechecking.
 
-use crate::storage::primitives::{AccessTerm, ArgsId, ParamsId, PatternId, TermId, TyFnCase};
+use crate::storage::{
+    location::LocationTarget,
+    primitives::{AccessTerm, ArgsId, ParamsId, PatternId, TermId, TyFnCase},
+};
 use hash_source::identifier::Identifier;
 
 use super::{
@@ -32,8 +35,8 @@ pub enum TcError {
     CannotUnifyParams {
         src_params_id: ParamsId,
         target_params_id: ParamsId,
-        src: TermId,
-        target: TermId,
+        src: LocationTarget,
+        target: LocationTarget,
         reason: ParamUnificationErrorReason,
     },
     /// The given term should be a type function but it isn't.
@@ -44,12 +47,17 @@ pub enum TcError {
     MismatchingArgParamLength {
         args_id: ArgsId,
         params_id: ParamsId,
-        params_subject: TermId,
-        args_subject: TermId,
+        params_subject: LocationTarget,
+        args_subject: LocationTarget,
     },
     /// The parameter with the given name is not found in the given parameter
     /// list.
-    ParamNotFound { args_id: ArgsId, params_id: ParamsId, params_subject: TermId, name: Identifier },
+    ParamNotFound {
+        args_id: ArgsId,
+        params_id: ParamsId,
+        params_subject: LocationTarget,
+        name: Identifier,
+    },
     /// There is a argument or parameter (at the index) which is
     /// specified twice in the given argument list.
     ParamGivenTwice { param_kind: ParamListKind, index: usize },
--- a/compiler/hash-typecheck/src/diagnostics/error.rs
+++ b/compiler/hash-typecheck/src/diagnostics/error.rs
@@ -128,4 +128,6 @@ pub enum TcError {
     },
     /// Given match case is never going to match the subject.
     UselessMatchCase { match_case_pattern: PatternId, subject: TermId },
+    /// Cannot use pattern matching in a declaration without an assignment
+    CannotPatternMatchWithoutAssignment { pattern: PatternId },
 }
--- a/compiler/hash-typecheck/src/diagnostics/error.rs
+++ b/compiler/hash-typecheck/src/diagnostics/error.rs
@@ -1,6 +1,6 @@
 //! Error-related data structures for errors that occur during typechecking.
 
-use crate::storage::primitives::{AccessTerm, ArgsId, ParamsId, TermId, TyFnCase};
+use crate::storage::primitives::{AccessTerm, ArgsId, ParamsId, PatternId, TermId, TyFnCase};
 use hash_source::identifier::Identifier;
 
 use super::{
@@ -116,4 +116,6 @@ pub enum TcError {
         // "terms".
         trt_def_missing_member_term_id: TermId,
     },
+    /// Given match case is never going to match the subject.
+    UselessMatchCase { match_case_pattern: PatternId, subject: TermId },
 }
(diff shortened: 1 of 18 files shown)


compiler/hash-typecheck/src/ops/typing.rs
```diff
@@ -320,16 +337,22 @@ impl<'gs, 'ls, 'cd, 's> Typer<'gs, 'ls, 'cd, 's> {
...
                 let builder = self.builder();
-                Ok(builder.create_fn_call_term(constructor_pattern.subject, args_id))
+                Ok(builder.create_tuple_lit_term(params_id))
             }
+            Pattern::Constructor(constructor_pattern) => match constructor_pattern.params {
+                Some(params) => {
+                    // We have params to apply
+                    let args_id = self.args_of_pattern_params(params)?;
+                    let builder = self.builder();
+                    Ok(builder.create_fn_call_term(constructor_pattern.subject, args_id))
```
feds01, line 349: I'm guessing it doesn't actually matter if it is struct/enum? But the naming is quite confusing here

compiler/hash-typecheck/src/ops/typing.rs
```diff
@@ -344,6 +367,52 @@ impl<'gs, 'ls, 'cd, 's> Typer<'gs, 'ls, 'cd, 's> {
...
+    }
+
+    /// Get the parameters of the given tuple term, if possible.
+    ///
+    /// This function returns Some(..) if the term is validated and simplified,
+    /// and is a tuple term (either literal or Rt.). Otherwise it will
+    /// return None.
+    ///
+    /// This function will populate default values if it can (if the tuple is a
+    /// literal).
```
feds01, line 385: work towards #295?

compiler/hash-typecheck/src/ops/unify.rs
```diff
@@ -170,15 +171,62 @@ impl<'gs, 'ls, 'cd, 's> Unifier<'gs, 'ls, 'cd, 's> {
...
+            // Names match
+            if src_param.name != target_param.name {
+                return cannot_unify(ParamUnificationErrorReason::NameMismatch(index));
+            }
+            // Values match
+            let ty_sub = self.unify_terms(src_param.value, target_param.value)?;
+
+            // Add to cumulative substitution
+            cumulative_sub.extend(&ty_sub);
+        }
```
feds01, line 215: there is a function to perform this kind of validation I think?

compiler/hash-typecheck/src/diagnostics/error.rs
```diff
@@ -16,14 +19,24 @@ pub type TcResult<T> = Result<T, TcError>;
...
+    /// same, or the types mismatch.
     CannotUnifyParams {
         src_params_id: ParamsId,
         target_params_id: ParamsId,
-        src: TermId,
-        target: TermId,
+        src: LocationTarget,
+        target: LocationTarget,
         reason: ParamUnificationErrorReason,
     },
```
feds01, line 41: these could probably be one variant

compiler/hash-typecheck/src/diagnostics/error.rs
```diff
@@ -116,4 +134,8 @@ pub enum TcError {
         // "terms".
         trt_def_missing_member_term_id: TermId,
     },
+    /// Given match case is never going to match the subject.
+    UselessMatchCase { match_case_pattern: PatternId, subject: TermId },
```
feds01, line 138:
```suggestion
    UselessMatchCase { pattern: PatternId, subject: TermId },
```

compiler/hash-typecheck/src/ops/patterns.rs
```diff
@@ -0,0 +1,152 @@
...
+                    Err(_) => Ok(None),
+                }
+            }
+            Pattern::Mod(_) => {
+                //  Here we have to basically try to access the given members using ns access...
+                todo!()
+            }
+            Pattern::Constructor(_) => {
+                // Get the term of the constructor and try to unify it with the subject:
+                let tuple_term = self.typer().term_of_pattern(pattern_id)?;
```
feds01, line 131:
```suggestion
                let constructor_term = self.typer().term_of_pattern(pattern_id)?;
```

compiler/hash-typecheck/src/ops/patterns.rs
```diff
@@ -0,0 +1,152 @@
...
+                match self.unifier().unify_terms(tuple_term, simplified_term_id) {
+                    Ok(_) => {
+                        // @@Todo: Get the vars:
+                        Ok(Some(vec![]))
```
feds01, line 135: I can take this one on

compiler/hash-typecheck/src/ops/patterns.rs
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                    Err(_) => Ok(None),
+                }
+            }
+            Pattern::Or(_) => {
+                todo!()
```
feds01, line 141: Basically a union between all of the pattern members themselves?

compiler/hash-typecheck/src/ops/typing.rs
```diff
@@ -253,49 +257,62 @@ impl<'gs, 'ls, 'cd, 's> Typer<'gs, 'ls, 'cd, 's> {
...
     }
 
-    /// From the given [PatternParamsId], infer a [ArgsId] describing the the
+    /// From the given [ArgsId], infer a [ParamsId] describing the the
     /// arguments.
-    pub(crate) fn args_of_pattern_params(
+    ///
+    /// This will populate the default values with the values of the args if
+    /// `populate_defaults` is true.
+    pub(crate) fn params_of_args(
```
feds01, line 265: Maybe a better name which captures that it `infers`?

compiler/hash-typecheck/src/ops/patterns.rs
kontheocharis: Yes but also need to ensure that the bound variables are the same for each branch and of the same type

compiler/hash-typecheck/src/ops/typing.rs
kontheocharis: Yes it doesn't, and I think it's better to tackle removal of enums in favour of structs/unions first before doing the rest of the constructor pattern stuff..

compiler/hash-typecheck/src/ops/typing.rs
kontheocharis: No, unrelated. It just utilises the default value field to save arg values when converting to params. (if they are needed later on)

compiler/hash-typecheck/src/ops/unify.rs
kontheocharis: Which one?

compiler/hash-typecheck/src/diagnostics/error.rs
kontheocharis: You mean src and target location targets?

compiler/hash-typecheck/src/ops/patterns.rs
feds01: makes sense, maybe leave it as a comment for a brave soul :^)

compiler/hash-typecheck/src/ops/patterns.rs
kontheocharis: Sounds good, see my comment above about constructors and enum removal tho.

compiler/hash-typecheck/src/ops/patterns.rs
kontheocharis: Shouldn't be too hard to do this since each branch will return a list of bounded variables.

compiler/hash-typecheck/src/ops/unify.rs
feds01: `pair_args_with_params` in `compiler/hash-typecheck/src/ops/params.rs` is pretty similar, wonder if can unify all of this logic

compiler/hash-typecheck/src/diagnostics/error.rs
feds01: No, I mean both variants `CannotUnifyArgs` and `CannotUnifyParams`, this was the idea behind `ParamListKind` to unify these kinds of errors (more specifically the error reporting logic)

kontheocharis: Ah yeah that would be nice.. I'll see what I can do

compiler/hash-typecheck/src/ops/unify.rs
kontheocharis: It's not really the same since the one you said does not perform any unification, just matching. Where as this one performs no matching, just unification.

compiler/hash-typecheck/src/diagnostics/error.rs
kontheocharis: It's actually not too easy to do this.. I think it is better left to another PR.

kontheocharis: See #415

feds01 (approved):

PR #413, opened 2022-07-16
parser+ast: remove `MethodCall` and finalise work on `Expr::Access`

### LINKED ISSUE #394: parser: expression is not correctly parsed
The following snippet:

```rust
(a.v)::c)<i32>
```

fails to parse and reports the error:

```rs
error: Failed to parse
 --> <interactive>:1:3
1 |    ((a.v)::c)<i32>
  |     ^^^^^ here
  = note: Expected an expression, however received a `:`.

```

### LINKED ISSUE #403: Remove `MethodCall` expr and just use function call

feds01:
This patch reworks how the parser deals with "singular expressions". There is now a loop that essentially tries to continue parsing "chunk" connectors like `::`, `<`, `.`, `[`, `(` and then the relevant inner expression, and then continues to try and parse the next chunk until it runs out of tokens or it didn't match any more applicable connecting tokens.
 
closes #394
closes #403


kontheocharis (approved):
What about access pattern?

feds01:
I think I am going to do this in a separate PR, same with a `type` variant

PR #432, opened 2022-07-28
TC: New Var terms for scopes and bounds

### LINKED ISSUE #393: TC: Cleanup bound variables; Are they even needed?

### LINKED ISSUE #429: TC: Remember scope variables instead of always resolving them

### LINKED ISSUE #430: TC: Improve substitutions to use scopes rather than replacements

kontheocharis:
This PR creates the following new terms:
- `ScopeVar`: A variable that points to some specific member in the stack, which might or might not be ready to be "resolved" into its value (for example if it will be modified later on). This property is called "closed"---if it is closed it means it can be substituted with its value already. This enables us to properly implement out-of-order declaration traversal.
- `BoundVar`: A variable which is bound by some parameter scope, i.e. type function parameter.
- `Var` (already existed): This now serves as an intermediary unsimplified term, that is, `Var` will always simplify to either `BoundVar` or `ScopeVar`. It exists because its simplification essentially implements name resolution.
- `SetBound`: Replaces `AppSub`. Parameter bounds are no longer substitutions, but rather scopes. This makes it easier to perform operations with things that are wrapped in `SetBound` because all that needs to happen is the scope needs to be added to the context and then perform the operation inside it (rather than applying substitutions haphazardly)

This change is a major step forward for tackling #419, #367.

Closes #393, closes #429, closes #430.

## Todo:

- [x] Fix merge conflicts
- [ ] Fix bugs relating to scope resolution: `M := <T> => mod { Z := T }; M<()>` stack overflows, `M<i32>` says that i32 is not valid.


compiler/hash-typecheck/src/ops/scope.rs
```diff
@@ -59,69 +65,139 @@ impl<'gs, 'ls, 'cd, 's> ScopeManager<'gs, 'ls, 'cd, 's> {
...
+            pair_args_with_params(&params, &args, params_id, args_id, params_subject, args_subject)
+                .unwrap_or_else(|err| {
+                    let report: Report = TcErrorWithStorage::new(err, self.storages()).into();
+                    eprintln!("{}", writer::ReportWriter::new(report, self.source_map()));
+                    tc_panic_on_many!(
+                        [params_subject, args_subject],
+                        self,
+                        "Could not pair arguments with parameters"
+                    )
+                });
```
feds01, line 156: An explanation to why this panics would be handy

compiler/hash-typecheck/src/ops/unify.rs
```diff
@@ -468,27 +555,17 @@ impl<'gs, 'ls, 'cd, 's> Unifier<'gs, 'ls, 'cd, 's> {
                 let subject_ty = reader.get_term(subject_ty_id);
                 match subject_ty {
                     Term::TyFnTy(ty_fn_ty) => {
-                        let ty_fn_ty = ty_fn_ty.clone();
+                        let _ = ty_fn_ty.clone();
```
feds01, line 558: does this need to be cloned?

compiler/hash-typecheck/src/storage/location.rs
```diff
@@ -157,6 +213,10 @@ impl LocationStore {
             LocationTarget::Pat(pat) => {
                 self.pat_map.insert(pat, location);
             }
+            _ => {
+                // no-op
+                log::warn!("Tried to set location on a LocationTarget::Location")
```
feds01, line 218: will this not fire on non-debug mode?

compiler/hash-typecheck/src/ops/simplify.rs
```diff
@@ -1399,9 +1460,9 @@ mod test_super {
         (global_storage, local_storage, core_defs, source_map)
     }
 
-    #[test]
-    fn test_simplify() {
-        let (mut global_storage, mut local_storage, core_defs, source_map) = get_storages();
+    // #[test]
+    fn _test_simplify() {
```
feds01, line 1464: broken test?

compiler/hash-typecheck/src/ops/discover.rs
```diff
@@ -0,0 +1,1078 @@
...
+    /// given [HashSet].
+    pub(crate) fn add_free_sub_vars_in_level3_term_to_set(
+        &self,
+        term: &Level3Term,
+        _: &mut HashSet<SubVar>,
+    ) {
+        match term {
+            Level3Term::TrtKind => {}
+        }
+    }
```
feds01, line 144: these are still to-dos?

kontheocharis: no

compiler/hash-typecheck/src/ops/simplify.rs
kontheocharis: I should probably just delete it since it's not even a proper test

compiler/hash-typecheck/src/storage/location.rs
kontheocharis: Yes, well I think we should have logging off for release no? Otherwise what purpose is there to logging?

compiler/hash-typecheck/src/ops/unify.rs
kontheocharis: Mistake, will fix

compiler/hash-typecheck/src/storage/location.rs
feds01: Yeah im just not sure about the actual rules when it is and isn’t turned off

compiler/hash-typecheck/src/ops/simplify.rs
feds01: Sounds good, maybe general tc tests are the way to go

compiler/hash-typecheck/src/ops/discover.rs
feds01: As in, the actual impl for these is as it is?

kontheocharis: Yes, fixed.

feds01 (approved):

PR #436, opened 2022-08-04
TC: Implement exhaustiveness and usefulness checking

### LINKED ISSUE #243: LM: Pattern refutability/usefulness
1. Read the refutability paper http://moscova.inria.fr/~maranget/papers/warn/warn.pdf
2. Check the rust implementation as reference (might be more useful than the paper in terms of actually implementing it)
3. Setup the necessary primitives in the typechecker to be able to check for usefulness (and make it ergonomic).
4. Implement the algorithm
5. Setup the error/warning messages.

feds01:
This PR re-organises the pattern exhaustiveness implementation into `Ops` style operations and now uses `DeconstructedPatId`, and `ConstructorId` to represent ids of constructors and de-constructed patterns within the 
implementation.

This PR also modifies the blanket implementation of the `Oracle` in order to be integrated with exhaustiveness and written 
independently of this PR

closes #243 

Some outstanding things:
- [x] Should we re-work `exhaustiveness::lower::Pat` to be used when printing the possible matches, or should we re-work `primitives::Pat` to circumvent having to lower to the intermediate representation and use the formatting implementation instead...
- [x] Can we have an `Option<PatId>` on `DeconstructedPat` so that we can avoid using the span, and easily have printing methods for `DeconstructedPat`, falling back on the `fmt::Debug` implementation that is available if the `id` is `None`.
- [x] We need to add some kind of validation logic in order to make sure that in some situations that the `rhs` type of the match contains terms that are matchable. For example:
   ```rs
    (a, b) := () => {};  
    ``` 
     Here is should be reported that `() -> ()` cannot be matched on


compiler/hash-typecheck/src/exhaustiveness/construct.rs
```diff
@@ -0,0 +1,324 @@
...
+/// @@Ranges: float ranges
+#[derive(Debug, Clone)]
+pub enum Constructor {
+    /// The constructor for patterns that have a single constructor, like
+    /// tuples, struct patterns and fixed-length arrays.
+    Single,
+    /// Enum variants.
+    Variant(usize),
+    /// Ranges of integer literal values (`2`, `2..=5` or `2..5`).
+    IntRange(IntRange),
```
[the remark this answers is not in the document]
kontheocharis: Would be nice to not use `IntRange` rust type, rather `(u128, u128)` to make this `Copy`. `IntRange` is an iterator which is not really needed in our case?

compiler/hash-typecheck/src/exhaustiveness/construct.rs
```diff
@@ -0,0 +1,324 @@
...
+                    Term::Level1(Level1Term::NominalDef(def)) => {
+                        match reader.get_nominal_def(*def) {
+                            NominalDef::Struct(struct_def) => match struct_def.fields {
+                                StructFields::Explicit(params) => reader.get_params(params).len(),
+                                StructFields::Opaque => 0,
+                            },
+                            // @@Remove: when enums are no longer represented as thi
+                            NominalDef::Enum(_) => unreachable!(),
+                        }
+                    }
```
kontheocharis, line 150: This could possibly be another thing that can benefit from the oracle, maybe `count_params_of_ty` or something.

compiler/hash-typecheck/src/exhaustiveness/construct.rs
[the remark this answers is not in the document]
kontheocharis: I would also possibly rename this variant, but not too bothered.

compiler/hash-typecheck/src/exhaustiveness/deconstruct.rs
```diff
@@ -1,1175 +1,520 @@
...
+        self.span
     }
 
-    /// Whether there are any value constructors for this type that are not
-    /// present in the matrix.
-    fn any_missing(&self, ctx: PatCtx<'tc>) -> bool {
-        self.iter_missing(ctx).next().is_some()
+    /// Function to get the `ty` of the [DeconstructedPat]
+    pub(super) fn ty(&self) -> TermId {
+        self.ty
```
kontheocharis, line 95: Possibly unneeded

compiler/hash-typecheck/src/exhaustiveness/deconstruct.rs
```diff
@@ -1,1175 +1,520 @@
...
             Constructor::Wildcard | Constructor::NonExhaustive => PatKind::Wild,
-            Constructor::Or => panic!("cannot convert an `or` deconstructed pat back into pat"),
+            Constructor::Or => panic!(
+                "cannot convert an `or` deconstructed pat back into pat"
+            ),
             Constructor::Missing => panic!(
-                "trying to convert a `Missing` constructor into a `Pat`; this is probably a bug,
+                "trying to convert a `Missing` constructor into a `Pat`; this is probably a bug,             
                 `Missing` should have been processed in `apply_constructors`"
             ),
```
kontheocharis, line 332: TC panics?

compiler/hash-typecheck/src/exhaustiveness/fields.rs
```diff
@@ -0,0 +1,145 @@
...
+use super::{construct::Constructor, AccessToUsefulnessOps};
+
+/// Representation of the `fields` that are stored by
+/// [DeconstructedPat](super::deconstruct::DeconstructedPat) which are nested
+/// patterns.
+#[derive(Debug, Clone)]
+pub struct Fields {
+    /// Vector of the inner ids stored by the [Fields]
+    pub fields: Vec<DeconstructedPatId>,
+}
```
kontheocharis, line 33: We can potentially inline store this in a store to avoid the overhead of having all these Vecs. Will tackle this later.

compiler/hash-typecheck/src/exhaustiveness/fields.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Create [Fields] from an [Iterator] of [Term]s.
+    pub fn wildcards_from_tys(&self, tys: impl IntoIterator<Item = TermId>) -> Fields {
+        Fields::from_iter(tys.into_iter().map(|ty| {
+            let pat = self.deconstruct_pat_ops().wildcard(ty);
+            self.deconstructed_pat_store().create(pat)
+        }))
+    }
+
+    /// Creates a new list of wildcard fields for a given constructor. The
+    /// result must have a length of `ctor.arity()`.
```
kontheocharis, line 78: `must` -> `will`

kontheocharis: Maybe also add an assert inside the function to ensure this?

compiler/hash-typecheck/src/exhaustiveness/fields.rs
```diff
@@ -0,0 +1,145 @@
...
+            }
+            Constructor::List(_list) => {
+                todo!()
+                // let arity = list.arity();
+
+                // // Use the oracle to get the inner term `T` for the type...
+                // let ty = ...;
+
+                // Fields::wildcards_from_tys((0..arity).map(|_| ty))
+            }
```
kontheocharis, line 134: todo

compiler/hash-typecheck/src/ops/building.rs
```diff
@@ -640,7 +640,7 @@ impl<'gs> PrimitiveBuilder<'gs> {
 
     /// Create an ignore pattern ("_").
     pub fn create_ignore_pat(&self) -> PatId {
-        self.create_pat(Pat::Ignore)
+        self.create_pat(Pat::Wild)
     }
```
kontheocharis, line 644: Call it wild instead of ignore everywhere.

compiler/hash-typecheck/src/ops/exhaustiveness.rs
```diff
@@ -0,0 +1,139 @@
...
+//! 
+//! ts := [(0, 'c'), (1, 'a'), (2, 'b')];
+//!
+//! for (key, value) in ts {
+//!     ...
+//! }
+//! ```
+//!
+//! The above pattern `(key, value)` is irrefutable because it covers
+//! all the possibilities of the type `(u32, char)` which is the type
```
kontheocharis, line 24: i32. We also have to actually make it be i32 by default, currently u64........

compiler/hash-typecheck/src/ops/pats.rs
```diff
@@ -183,8 +183,18 @@ impl<'tc> PatMatcher<'tc> {
                     Err(_) => Ok(None),
                 }
             }
+            Pat::Range { lo, hi, .. } => {
+                // The `lo` and `hi` should always unify with the subject term...
+                self.unifier().unify_terms(lo, simplified_term_id)?;
+                self.unifier().unify_terms(hi, simplified_term_id)?;
+
+                // They should also unify with one another...
+                self.unifier().unifier().unify_terms(hi, lo)?;
```
kontheocharis, line 192: This should happen in pattern validation (a function on validator, or infer ty of pat ?).

Can you also add a @@Fixme at the top of this function to rework level 0 unification (i.e. should never work when Rts are involved), and that this function should unify the *types*, rather than the values, of the pattern and term.

compiler/hash-typecheck/src/ops/reader.rs
```diff
@@ -66,6 +70,16 @@ impl<'gs> PrimitiveReader<'gs> {
         self.gs.pat_args_store.get(id)
     }
 
+    /// Get the associated [DeconstructedPat] from [DeconstructedPatId].
+    pub fn get_deconstructed_pat(&self, id: DeconstructedPatId) -> DeconstructedPat {
+        self.gs.deconstructed_pat_store.get(id)
+    }
+
+    /// Get the associated [Constructor] from [ConstructorId].
+    pub fn get_ctor(&self, id: ConstructorId) -> Constructor {
```
kontheocharis, line 79: Maybe DeconstructedCtor instead of Constructor? Similarly for method name

compiler/hash-typecheck/src/storage/primitives.rs
```diff
@@ -1279,6 +1279,9 @@ pub enum Pat {
     Access(AccessPat),
     /// Resolved binding pattern.
     Const(ConstPat),
+    /// A range pattern `2..5`, the `lo` and `hi` values
+    /// must be `Term::Level0(Level0Term::Lit)`.
+    Range { lo: TermId, hi: TermId, end: RangeEnd },
```
kontheocharis, line 1284: Make this a separate struct please sir.

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -1556,6 +1570,9 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
             self.state.declaration_name_hint = Some(name);
         };
 
+        // @@Todo: is pattern irrefutable?
+        // self.exhaustiveness_checker().is_pat_irrefutable(&[pat], term, None)?;
+
```
kontheocharis, line 1575: remove

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -1577,6 +1594,18 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
...
+        let pat_ty = self.typer().infer_ty_of_pat(pat_id)?;
+        self.unifier().unify_terms(pat_ty, ty)?;
+
+        // Ensure that the given pattern is irrefutable given the type of the term
+        //
+        // @@Investigate: Unclear what to do here if the `rhs` is a type that is
+        // not meant to be checked for refutability, presumably we always declare it
+        // is irrefutable because the `value` and `ty` matched...
+        self.exhaustiveness_checker().is_pat_irrefutable(&[pat_id], ty, None)?;
+
```
kontheocharis, line 1608: Move exhaustiveness check after the `let members = ` line, and then you can remove unification since it happens in `match_pat_with_term`.

compiler/hash-typecheck/src/exhaustiveness/construct.rs
feds01: This would be a very beneficial utility, and possibly things like `as_struct`, `as_enum_variant`...

compiler/hash-ast/src/ast.rs
```diff
@@ -730,11 +875,20 @@ pub enum Pat {
...
     Or(OrPat),
+    /// A pattern that is guarded by an if statement, e.g. `x if x > 5`
     If(IfPat),
-    Ignore(IgnorePat),
+    /// Wildcard pattern, similar to a binding but it is not binded
+    /// to any member.
+    Wild(WildPat),
+    /// Similar to a [Pat::Wild], but does captures a collection of patterns,
+    /// which can be used to ignore a range of elements in a tuple or
+    /// a list.
```
kontheocharis, line 891: Surely this isn't similar to wild?

compiler/hash-ast/src/ast.rs
```diff
@@ -730,11 +875,20 @@ pub enum Pat {
...
     List(ListPat),
+    /// A literal pattern e.g. `c`
     Lit(LitPat),
+    /// An `or` pattern which groups multiple patterns e.g. `a | b | c`
     Or(OrPat),
+    /// A pattern that is guarded by an if statement, e.g. `x if x > 5`
     If(IfPat),
-    Ignore(IgnorePat),
+    /// Wildcard pattern, similar to a binding but it is not binded
+    /// to any member.
```
kontheocharis, line 887: *bound

compiler/hash-ast/src/ast.rs
```diff
@@ -730,11 +875,20 @@ pub enum Pat {
     Module(ModulePat),
     /// A tuple pattern is a collection of patterns, e.g `(1, x, 'c')`
     Tuple(TuplePat),
+    /// A list pattern, which is a collection of patterns e.g `[x, 2, y]`
```
kontheocharis, line 878: Which contains a collection of patterns, including spread, and matches a list

compiler/hash-ast/src/ast.rs
```diff
@@ -730,11 +875,20 @@ pub enum Pat {
...
     List(ListPat),
+    /// A literal pattern e.g. `c`
     Lit(LitPat),
+    /// An `or` pattern which groups multiple patterns e.g. `a | b | c`
```
kontheocharis, line 882: Which matches one of the given patterns

compiler/hash-ast/src/ast.rs
```diff
@@ -642,12 +781,18 @@ pub struct StrLitPat(pub Str);
...
+}
 
 /// A float literal pattern.
-#[derive(Debug, PartialEq, Clone)]
-pub struct FloatLitPat(pub f64);
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub struct FloatLitPat {
+    pub value: f64,
+    pub kind: FloatLitKind,
+}
```
kontheocharis, line 795: Why does this exist?

compiler/hash-ast/src/ast.rs
```diff
@@ -642,12 +781,18 @@ pub struct StrLitPat(pub Str);
 pub struct CharLitPat(pub char);
 
 /// An integer literal pattern.
-#[derive(Debug, PartialEq, Eq, Clone)]
-pub struct IntLitPat(pub u64);
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub struct IntLitPat {
+    pub value: u128,
```
kontheocharis, line 786: Maybe a function to get the "actual" rendered value of the integer? Maybe as a signed bigint or something?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -135,123 +54,319 @@ impl<'tc> LowerPatOps<'tc> {
...
         let reader = self.reader();
+        let pat = reader.get_pat(pat_id);
+
+        let (ctor, fields) = match pat {
+            Pat::Wild
+            | Pat::Mod(_)
+            | Pat::Binding(_)
+            | Pat::Spread(_)
+            | Pat::Access(_)
+            | Pat::Const(_) => (DeconstructedCtor::Wildcard, vec![]),
```
kontheocharis, line 68: How is const a wildcard?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -135,123 +54,319 @@ impl<'tc> LowerPatOps<'tc> {
...
-    /// the pattern.
-    pub fn lower_pat(&self, id: PatId) -> Pat {
+    /// Convert a [Pat] into a [DeconstructedPat].
+    pub(crate) fn deconstruct_pat(&self, ty: TermId, pat_id: PatId) -> DeconstructedPat {
         let reader = self.reader();
+        let pat = reader.get_pat(pat_id);
+
+        let (ctor, fields) = match pat {
+            Pat::Wild
+            | Pat::Mod(_)
```
kontheocharis, line 64: How is mod a wildcard? Is it always exhaustive?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -135,123 +54,319 @@ impl<'tc> LowerPatOps<'tc> {
...
+            | Pat::Binding(_)
+            | Pat::Spread(_)
+            | Pat::Access(_)
```
kontheocharis, line 67: How is access a wildcard?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -135,123 +54,319 @@ impl<'tc> LowerPatOps<'tc> {
...
+                    Term::Level1(Level1Term::NominalDef(_)) => {
+                        let _children = pat.fields.iter_patterns().map(|p| self.construct_pat(*p));
 
-                match nominal_def {
-                    NominalDef::Struct(_) => PatKind::Leaf { pats },
-                    // @@EnumToUnion: get the variant index here
-                    NominalDef::Enum(_) => unreachable!(),
+                        // let args = self.builder().create_pat_args(children);
+                        // Pat::Constructor(ConstructorPat { subject: pat.ty, args })
+                        todo!()
```
kontheocharis, line 266: todo left

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -135,123 +54,319 @@ impl<'tc> LowerPatOps<'tc> {
...
+                        let _children = pat
+                            .fields
+                            .iter_patterns()
+                            .map(|p| self.construct_pat(*p))
+                            .collect_vec();
+
+                        // let args = self.builder().create_pat_args(children);
+                        // Pat::Tuple(args)
+
+                        todo!()
```
kontheocharis, line 259: todo left

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -135,123 +54,319 @@ impl<'tc> LowerPatOps<'tc> {
...
+                        let spread = self.pat_store().create(dummy);
+
+                        // Now create an inner collection of patterns with the inserted
+                        // spread pattern
+                        let _inner = prefix.into_iter().chain(once(spread)).chain(suffix);
+                        let _term = self.oracle().term_as_list(pat.ty).unwrap();
+
+                        // let elements = self.builder().create_pat_args(inner);
+                        // Pat::List(ListPat { term, inner: elements })
+                        todo!()
```
kontheocharis, line 308: todo left

kontheocharis (changes requested):

compiler/hash-ast/src/ast.rs
feds01: It's similar in the sense that it matches all elements and doesn't bind any variable, but the only difference is that it captures multiple elements or patterns.

compiler/hash-ast/src/ast.rs
feds01: It exists so that it can give a concrete error to why this occurred, and it might be used in ranges as the `lo` and `hi`

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: It has the same properties as a wild pattern in terms of exhaustiveness

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: Yes it is always exhaustive, you can't create a non-exhaustive `mod` pattern

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: It gets simplified to a wildcard anyway, there are no situations where it can be a non-equivalent of a binding

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: The todo is dependant on the `builder` having immutable access

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: The todo is dependant on the `builder` having immutable access

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: The todo is dependant on the `builder` having immutable access

compiler/hash-ast/src/ast.rs
feds01: how do you mean?

compiler/hash-ast/src/ast.rs
kontheocharis: yeah, and it can *only* be used in list context. I think that last one is quite important

compiler/hash-ast/src/ast.rs
kontheocharis: So do we support float pattern ranges? Does the literal `1.0` always match `1.0..2.0`?

compiler/hash-ast/src/ast.rs
kontheocharis: Like, if the integer is negative in the Hash code for example, the `value` field will appear like garbage if you try to print it. We should have a function to get a representable integer which is true to what it originally was in the source. That would make it appear correctly in `fmt`.

compiler/hash-typecheck/src/exhaustiveness/lower.rs
kontheocharis: Const is just a constant member of an enum; `Color := enum(Red, Green, Blue); match c { Red => ... }` -> `Red` is `Pat::Const`.

compiler/hash-typecheck/src/exhaustiveness/lower.rs
kontheocharis: What about `{ value as Some(value) } := import("my_mod");`

compiler/hash-typecheck/src/exhaustiveness/lower.rs
kontheocharis: what about `Option::Some` ? I guess it would get simplified to a `Pat::Const`?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
kontheocharis: Maybe write a `@@Todo: immutable builder`?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: Yes it does get simplified to a `Pat::Const`

compiler/hash-ast/src/ast.rs
feds01: Ah yes, it does need to be transformed properly

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: Yeah, so that's one of the cases that should be handled, so presumably,  for `Access`  and `Const`, it should try and get the type of the actual pattern that's passed in and create a Destructor based on that...

Does this mean that the only valid type to correspond to an access pattern would be an enum variant?

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: Yes, I mean it could basically be treated as a struct-like pattern, where these fields are represented as these patterns, and the rest are treated as wildcards.

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: But also this would be an invalid match... since it could never match the type of `Option::Some` with the subject

kontheocharis: I meant Option::None

compiler/hash-typecheck/src/exhaustiveness/lower.rs
kontheocharis: Not necessarily, `my_mod::a::b::MyStruct(a)` is a valid pattern and is first and foremost `Pat::Access`

compiler/hash-ast/src/ast.rs
feds01: Hmm, you are right, it is probably impossible to get a sound implementation, but I don't see much point in getting rid of the node just for consistency's sake

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: yeah then it is a `Constructor` pat which is handled differently

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: Yes, then the comment about access/const pats applies

compiler/hash-ast/src/ast.rs
kontheocharis: I guess it might be useful for errors..

compiler/hash-typecheck/src/exhaustiveness/lower.rs
kontheocharis: Yeah. Also to note that `Const` is logically equivalent to `Constructor` with zero arguments.

compiler/hash-ast/src/ast.rs
feds01: Yes, I think the error reporting could benefit from this, I don't think it adds bloat, but we could make a `Pat::Lit` in the AST which just has an inner `Literal` rather than having copies for each variant for both literals and pattern literals

compiler/hash-typecheck/src/exhaustiveness/lower.rs
feds01: Yes, it is, so I think that the only valid case for the type would be an enum, like say `None`?

kontheocharis: Yes. However, with unions, this essentially translates to unit structs; like `None := unit(); Some := <T> => struct(value: T); Option := <T> => type Some<T> | None`. There isn't syntax for this, but internally this will need to happen (i.e. nominal defs are either structs or units)

kontheocharis: The only difference between `unit()` and `struct()` is that you don't need `()` to instantiate a unit.

feds01: I will split the two branches, but I think it will be clearer to implement this when unions replace enums internally, so I'll just leave a `todo!()`

compiler/hash-ast/src/ast.rs
feds01: Maybe this is the time to move this into its own crate like `hash_abi`?

kontheocharis (approved):

kontheocharis (approved):

PR #447, opened 2022-08-05
TC: Implement generic stores with interior mutability

kontheocharis:
These are meant to replace all the currently hardcoded implementations of stores, without interior mutability. Hopefully with this change, three things can happen:
1. The typechecker ops structs can take `&self` rather than `&mut self`, which should save a lot of unnecessary clones. The unavoidable clones (if needed) are done in the new store implementations.
2. Term lists such as `Merge` and `Union` do not need to hold `Vec`s anymore, and instead can point to `SequenceStore`s. `SequenceStore` elements are stored inline, which should be a performance boost.
3. The plan is to use `SequenceStore` as a base to implement `ParamList` and `Scope` in a more performant way and with interior mutability.

Works towards #445.

### DIFF
--- a/compiler/hash-typecheck/src/storage/store.rs
+++ b/compiler/hash-typecheck/src/storage/store.rs
@@ -1,0 +1,399 @@
+use std::{cell::RefCell, hash::Hash, marker::PhantomData, ops::Range};
+
+/// Represents a key that can be used to index a [`Store`].
+pub trait StoreKey: Copy + Eq + Hash {
+    /// Turn the key into an index.
+    fn to_index(self) -> usize;
+    /// Create a key from an index.
+    ///
+    /// This should generally not be used in client code.
+    fn from_index_unchecked(index: usize) -> Self;
+}
+
+/// Create a new [`StoreKey`] with the given name.
+#[macro_export]
+macro_rules! new_store_key {
+    ($visibility:vis $name:ident) => {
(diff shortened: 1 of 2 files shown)


compiler/hash-typecheck/src/storage/store.rs
```diff
@@ -0,0 +1,401 @@
+//! Provides generic data structures to store values by generated keys in an
+//! efficient way, with interior mutability.
+use std::{cell::RefCell, hash::Hash, marker::PhantomData, ops::Range};
+
+/// Represents a key that can be used to index a [`Store`].
+pub trait StoreKey: Copy + Eq + Hash {
+    /// Turn the key into an index.
+    fn to_index(self) -> usize;
```
feds01, line 8: change to `u32`

compiler/hash-typecheck/src/storage/store.rs
```diff
@@ -0,0 +1,401 @@
...
+        let ret = f(&mut value);
+        self.set(key, value);
+        ret
+    }
+}
+
+/// Represents a key that can be used to index a [`SequenceStore`].
+pub trait SequenceStoreKey: Copy + Eq + Hash {
+    /// Turn the key into an index and a length.
+    fn to_index_and_len(self) -> (usize, usize);
```
feds01, line 129: change to `u32`

feds01 (changes requested):

feds01 (approved):

PR #448, opened 2022-08-05
TC/Utils: Move store to `hash_utils` and refine implementation

kontheocharis:
The stores are now traits, so that they can be built upon (any struct that can provide an appropriate `RefCell<Vec<_>>` can easily inherit a store implementation). This is needed for things like param stores etc.

PR already to keep the change sets small and manageable.


feds01 (approved):

feds01:
alloc label?

PR #452, opened 2022-08-07
lang: add support for range pats

### LINKED ISSUE #438: Add integer range patterns
In an effort to consolidate the typechecking `primitives::Pat` and `exhaustiveness::lower::Pat`  data structures, integer ranges need to be added to the language:

Questions to answer:

- What is the syntax? My proposition would be:
-  - `a..b` for open range, as in ` a <= x < b`
-  - `a..=b` for open range, as in ` a <= x <= b`
-  - `a..` for unbounded range, as in ` a <= x <= type_of(a)::MAX` with the assumption that `a` is of an integral type
-  - should we have some kind of stepping variant?

Things that need to be done:
- AST:
-  - New `pat` variant
- Support in parsing
- Semantic Analysis (ensure range makes sense)
-  - lower bound is smaller or equal to the upper bound 
-  Support in tc traversal
- Support in exhaustiveness checking (needs to be hooked up with existing logic)

@kontheocharis thoughts?

feds01:
This PR adds support for range patterns on the AST, Parser, Semantics, and typechecking levels.

closes #438


compiler/hash-typecheck/src/ops/validate.rs
```diff
@@ -1192,4 +1194,56 @@ impl<'tc> Validator<'tc> {
...
+
+    /// Validate that a [Pat::Range] has matching `lo`, and `hi` types, and that
+    /// the values of ranges are semantically correct
+    pub(crate) fn validate_range_pat(&mut self, pat: &RangePat) -> TcResult<()> {
+        let RangePat { lo, hi, end } = *pat;
+
+        // Expect to be literal terms, which should definitely unify
+        let hi_ty = self.typer().infer_ty_of_term(hi)?;
+        let lo_ty = self.typer().infer_ty_of_term(lo)?;
+        let _ = self.unifier().unify_terms(hi_ty, lo_ty)?;
```
kontheocharis, line 1206: This should be a bi-directional unification

kontheocharis (changes requested):

kontheocharis:
Nice one!

compiler/hash-typecheck/src/ops/validate.rs
feds01: Should it use `terms_are_equal()`

kontheocharis: Yeah, maube create `unify_terms_bi` which is the same as terms_are_equal but retains the TcResult.

feds01: Should it discard the substitutions or apply them on the terms?

kontheocharis: the unify_terms_bi should merge the two subs and return them. the validator should ensure empty.

feds01:
@kontheocharis re-review?

kontheocharis (approved):

PR #459, opened 2022-08-08
reporting: Introduce `Diagnostics` trait and intergrate the `Lexer` as a proof of concept

### LINKED ISSUE #437: Add diagnostics trait that unifies compiler error/warning handling

### LINKED ISSUE #443: Migrate semantic analysis to use `Diagnostics` trait

feds01:
This patch adds the `Diagnostics` trait into reporting and integrates it as a proof of concept with the Lexer. This also means that the lexer is now capable of error recovery and multiple error reporting!

Additionally, this simplifies the interaction between error types within the Parser and the Lexer. Things like the `LexerError` don't need to be converted into `ParseError` and the reported, this now happens directly. Similarly, this also removes the intermediate `AstGenError` since this has now been de-coupled from `ImportError` and `LexerError`, so it is now just `ParseError`, and `ParseResult` 

closes #437
closes #443

### DIFF
--- a/compiler/hash-lexer/src/error.rs
+++ b/compiler/hash-lexer/src/error.rs
@@ -1,5 +1,7 @@
 //! Hash Compiler lexer error data types.
 
+use std::convert::Infallible;
+
 use derive_more::Constructor;
 use hash_reporting::{
     builder::ReportBuilder,
@@ -81,7 +83,7 @@ pub struct LexerDiagnostics {
     errors: Vec<LexerError>,
 }
 
-impl Diagnostics<LexerError, LexerError> for Lexer<'_> {
+impl Diagnostics<LexerError, Infallible> for Lexer<'_> {
     type DiagnosticsStore = LexerDiagnostics;
 
     fn store(&self) -> &Self::DiagnosticsStore {
@@ -95,8 +97,8 @@ impl Diagnostics<LexerError, LexerError> for Lexer<'_> {
 
     /// The lexer does not currently emit any warnings and so if this
     /// is called, it should panic.
-    fn add_warning(&mut self, _: LexerError) {
-        unimplemented!()
+    fn add_warning(&mut self, warning: Infallible) {
+        match warning {}
     }
 
     fn has_errors(&self) -> bool {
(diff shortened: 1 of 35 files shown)


compiler/hash-lexer/src/error.rs
```diff
@@ -58,38 +61,63 @@ pub enum LexerErrorKind {
...
+
+    /// Add an error into the store
+    fn add_error(&mut self, error: LexerError) {
+        self.diagnostics.errors.push(error);
+    }
+
+    /// The lexer does not currently emit any warnings and so if this
+    /// is called, it should panic.
+    fn add_warning(&mut self, _: LexerError) {
+        unimplemented!()
```
kontheocharis, line 99: if you use Infallible for the warning then here you can put an empty match {} and it will typecheck

compiler/hash-lexer/src/error.rs
```diff
@@ -58,38 +61,63 @@ pub enum LexerErrorKind {
...
+}
+
+/// Lexer error store, the lexer stores an internal store so that
+/// it can implement [Diagnostics::DiagnosticsStore]
+#[derive(Default)]
+pub struct LexerDiagnostics {
+    errors: Vec<LexerError>,
+}
+
+impl Diagnostics<LexerError, LexerError> for Lexer<'_> {
```
kontheocharis, line 84: Maybe use Infallible for the warning ?

kontheocharis (changes requested):

compiler/hash-lexer/src/error.rs
feds01: But the `Diagnostics` trait specifies that `W` needs to implement `Into<Report` which `Infallible` does not

kontheocharis: Okay then maybe enum LexerWarning {}

feds01: Ive left it as Infallible for now, im sure `LexerWarning` will be added in the future, but let’s keep it like this for now?

compiler/hash-reporting/src/report.rs
```diff
@@ -152,3 +155,20 @@ impl Report {
...
+
+        // @@ErrorReporting: we might want to show a bit more info here.
+        report.with_kind(ReportKind::Error).with_message(err.to_string());
+        report.build()
+    }
+}
+
+impl From<Infallible> for Report {
+    fn from(_: Infallible) -> Self {
+        unreachable!()
```
kontheocharis, line 172: empty match will suffice

kontheocharis (changes requested):

kontheocharis (approved):

PR #467, opened 2022-08-11
Interned literals and numeric type suffixes

### LINKED ISSUE #465: Add integer/float type ascriptions
The parser and lexer should support annotating integer and float literals with type ascriptions:

```
6i32
3.2f32
18usize
```

feds01:
This patch does the following
- make `integer` and `float` literals become interned. 
- support numeric literal suffixes that describe which kind of numeric literal it is, e.g. `3usize` `3.2f32`, etc.
- fixed unhandled case in lexer where no digits would follow a prefix
- ensure float literals can't have a base prefix that's non-decimal.
- simplify the way the lexer and parser deal with negated numerics.

closes #465


compiler/hash-source/src/constant.rs
```diff
@@ -0,0 +1,291 @@
...
+
+use crate::identifier::{Identifier, CORE_IDENTIFIERS};
+
+/// Interned float constant which stores the value of the float, and
+/// an optional `type ascription` which is a suffix on the literal
+/// describing which float kind it is, either being `f32` or `f64`.
+#[derive(Debug, Clone, Copy)]
+pub struct FloatConstant {
+    /// Raw value of the float
+    pub value: f64,
```
kontheocharis, line 23: Would it be valid to have NaN or Infinity here ?

compiler/hash-source/src/constant.rs
```diff
@@ -0,0 +1,291 @@
...
+
+/// Value of the [IntConstant], this kind be either the `inlined`
+/// variant where we just fallback to using `u128` for small sized
+/// integer constants, and then in the unlikely scenario of needing
+/// more than a [u128] to represent the constant, we will then
+/// fallback to [BigInt].
+#[derive(Debug, Eq, PartialEq, Hash)]
+pub enum IntConstantValue {
+    /// For small values, we store inline
+    Small(u128),
```
kontheocharis, line 62: Maybe u64? 99.99% of the time no one will write a literal > 64 bits

compiler/hash-source/src/constant.rs
```diff
@@ -0,0 +1,291 @@
...
+/// Interned literal constant which stores the raw value of the
+/// integer and an optional `type ascription` which binds the
+/// defined literal value to some ascribed type.
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct IntConstant {
+    /// Raw value of the integer. This is stored as either a `u128` which can be
+    /// directly stored as the value which happens in `99%` of cases, if the
+    /// constant is not big enough to store this integer, then we resort to
+    /// using [IntConstant
+    pub value: IntConstantValue,
```
kontheocharis, line 91: This probably shoudn't be pub, someone might use the small variant and treat it as a number rather than bytes..

compiler/hash-source/src/constant.rs
kontheocharis: Also, maybe use an inline byte array instead  `[u8; 8]` (or `[u8; 16]`)? better communicates the fact that this data is "serialised" in the case of negative integers eg.

compiler/hash-token/src/lib.rs
```diff
@@ -272,6 +280,46 @@ impl std::fmt::Display for TokenKind {
...
+    Binary,
+    /// Octal base, denoted in literals as `0o26317261`
+    Octal,
+    /// Decimal base, written as `102391`
+    Decimal,
+    /// Hexadecimal base, written as `0xdeadbeef`
+    Hex,
+    /// Unsupported base, the language doesn't support the
+    /// provided radix as a base.
+    Unsupported,
```
kontheocharis, line 296: Why is this needed?

kontheocharis (changes requested):

compiler/hash-source/src/constant.rs
feds01: I believe that `NaN` and `inf` can be represented within `f64`, so yes?

compiler/hash-token/src/lib.rs
feds01: It's not really needed, it's just that I thought that maybe instead of using an `unwrap()` when converting the `u32` into `Base`, it wouldn't panic and just fails gracefully. By the term "gracefully", I mean that it would say "unsupported base", and then someone can submit a bug report if that ever does occur which it shouldn't.

kontheocharis: Maybe this should live inside lexer then

kontheocharis (approved):

PR #476, opened 2022-08-18
Tc: error reporting and bug fixes

### LINKED ISSUE #456: Make variables either constant or variable depending on their type

feds01:
This patch brings several fixes to reported locations in the reported terms.

- fix overwriting location when visiting variables, this should only occur for literal values.
- improve printing of sequences by introducing a `limit` term
- fix `term_of_pat` overwriting locations when it shouldn't
- fix validation of non-existent fields in constructor pattern validation

closes #456

### DIFF
--- a/compiler/hash-typecheck/src/traverse/mod.rs
+++ b/compiler/hash-typecheck/src/traverse/mod.rs
@@ -119,7 +119,7 @@ impl<'tc> TcVisitor<'tc> {
 
     /// Create a [SourceLocation] from a [Span].
     pub(crate) fn source_location(&self, span: Span) -> SourceLocation {
-        SourceLocation { span, source_id: self.local_storage().current_source() }
+        SourceLocation { span, id: self.local_storage().current_source() }
     }
 
     /// Create a [SourceLocation] at the given [hash_ast::ast::AstNode].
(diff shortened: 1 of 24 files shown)


compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -2138,27 +2148,40 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
...
+        if let Ok(scope_member) = self.scope_manager().resolve_name_in_scopes(name, var_term) {
+            let term = scope_member.member.value_or_ty();
+            let ty = scope_member.member.ty();
+
+            // So this should only become a constant if we are a `enum` or `struct`
+            // definition. Otherwise, we shadow the variable
+            if self.oracle().term_is_struct_def(term) || self.oracle().term_is_enum_def(term) {
+                // @@Hack: we technically want to use the `term` from the scope here, but
+                // because the locations are messed up, instead we just copy the
+                // location to the created term
```
kontheocharis, line 2162: What do you mean by "the locations are messed up"?

kontheocharis (approved):

compiler/hash-typecheck/src/traverse/mod.rs
feds01: I think the comment is now inaccurate, since the term location should originate from the type that's of the variable. However, it doesn't matter since it will be simplified immediately in `visit_constructor_term` which then takes the struct/enum def as the term and thus the same location. So actually, this location copy could probably be removed

feds01: actually this is not a correct statement, the copy is necessary but the comment is off so I am going to remove it.

kontheocharis (approved):

PR #478, opened 2022-08-19
tc: refactor `Term::{Merge|Union}` to use `TermListId`

### LINKED ISSUE #387: TC: Factor `Vec<Term>` into `TermListId` in `Merge` and `Union` terms
Benefits:
- Smaller size of Term
- Term can implement Copy which prevents tons of cloning
- Brings Merge and Union in line with Params and Args.

feds01:
closes #387

### DIFF
--- a/compiler/hash-typecheck/src/fmt.rs
+++ b/compiler/hash-typecheck/src/fmt.rs
@@ -3,7 +3,7 @@
 use core::fmt;
 use std::{cell::Cell, fmt::Display, rc::Rc};
 
-use hash_utils::store::Store;
+use hash_utils::store::{SequenceStore, SequenceStoreKey, Store};
 
 use crate::storage::{
     arguments::ArgsId,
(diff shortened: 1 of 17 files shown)


compiler/hash-typecheck/src/fmt.rs
```diff
@@ -390,30 +392,13 @@ impl<'gs> TcFormatter<'gs> {
...
-                    if i != terms.len() - 1 {
-                        write!(f, " ~ ")?;
-                    }
-                }
-                Ok(())
-            }
+            Term::Merge(terms) => self.fmt_term_list(f, terms, "~", opts),
             Term::Union(terms) => {
-                if terms.is_empty() {
+                if terms.len() == 0 {
```
kontheocharis, line 397: Why? is_empty exists too

kontheocharis (approved):

PR #488, opened 2022-08-22
Lang: support numeric field access operations.

### LINKED ISSUE #473: Support numeric field access on tuples and un-named structs/enums

### LINKED ISSUE #474: Support named fields in enums

### LINKED ISSUE #475: Support un-named struct fields

feds01:
closes #473
closes #474 
closes #475


compiler/hash-ast/src/ast.rs
```diff
@@ -1634,13 +1634,24 @@ impl Display for AccessKind {
     }
 }
 
+/// The kind of property that's being accessed, either being
+/// named or numeric, e.g `foo.x`, `foo.1`, etc.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum PropertyKind {
+    /// A named field like
+    Field(Identifier),
```
kontheocharis, line 1642: Maybe NamedField?

compiler/hash-parser/src/parser/expr.rs
```diff
@@ -778,10 +778,41 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
...
+                let interned_lit = CONSTANT_MAP.lookup_int_constant(value);
+
+                if let Some(suffix )= interned_lit.suffix {
+                    return self.err_with_location(ParseErrorKind::DisallowedSuffix(suffix), None, None, token.span)?;
+                }
+
+                self.skip_token();
+                let value = usize::try_from(interned_lit.to_big_int()).map_err(|_| {
+                    // @@Todo: deal with invalid numeric field access
+                    todo!()
```
kontheocharis, line 796: Todo

compiler/hash-typecheck/src/ops/simplify.rs
```diff
@@ -371,18 +407,21 @@ impl<'tc> Simplifier<'tc> {
...
                     }
                     NominalDef::Enum(enum_def) => {
                         // Enum type access results in the runtime value of the variant
                         // (namespace operation).
                         does_not_support_prop_access(access_term)?;
-                        match enum_def.variants.get(&access_term.name) {
+
+                        let Field::Named(name) = access_term.name else {
+                            unreachable!()
+                        };
```
kontheocharis, line 422: Shouldn't this be a user error?

compiler/hash-typecheck/src/ops/simplify.rs
```diff
@@ -414,11 +453,15 @@ impl<'tc> Simplifier<'tc> {
             Level2Term::Trt(trt_def_id) => {
                 does_not_support_prop_access(access_term)?;
 
+                let Field::Named(name) = access_term.name else {
+                    unreachable!()
+                };
```
kontheocharis, line 458: Same as above

compiler/hash-typecheck/src/ops/simplify.rs
feds01: It's handled in `does_not_support_prop_access` in the call above.

compiler/hash-typecheck/src/ops/simplify.rs
feds01: It's handled in `does_not_support_prop_access` in the call above.

compiler/hash-parser/src/parser/expr.rs
feds01: Not sure what the behaviour should be here...

kontheocharis: Surely it will never occur? What are the cases where it would

compiler/hash-typecheck/src/ops/simplify.rs
kontheocharis: Maybe that function could return the `name` to avoid having to write unreachable.

compiler/hash-parser/src/parser/expr.rs
feds01: Only when the specified `name` index is bigger than a `usize`

kontheocharis (approved):

Issue #389, opened 2022-07-14
TC: Properly ensure impls implement their traits, and infer original types while at it


kontheocharis:
1. Ensure all members in the impl appear in the trait, and have correct types (extension: if types are incomplete, infer from trait. Maybe this should be done in the future). (Unification needs to happen with src=impl and target=trait)
3. Ensure all members in the trait appear in the impl.

feds01:
Is `1` not already implemented here:
https://github.com/hash-org/lang/blob/b8b30751a648016b10e71bda13163be4ec099a9a/compiler/hash-typecheck/src/ops/validate.rs#L235-L241

?

kontheocharis:
Yes, I was just writing everything for completeness.

feds01:
So everything is complete here? Ill be putting a PR that polishes error reporting a bit and then this will close it?

kontheocharis:
> So everything is complete here? Ill be putting a PR that polishes error reporting a bit and then this will close it?

Actually yes, it seems so.

PR #495, opened 2022-08-23
reporting: minimalise report code blocks

feds01:
This patch improves the printing of large code blocks by collapsing intermediate lines with a `...`. This was initially entailed in #290 and is now implemented with this PR.

In order to test the implementation, the trait implementation validation logic was partially implemented, specifically the detection of methods present in the trait implementation that are not defined in the trait definition. partially completes #389 

This also fixes a bug with `visit_constant_scope` trying to take the declaration hint after it visits children which meant that the parent declaration hint was overwritten by the children before it could be used by the parent.

### DIFF
--- a/compiler/hash-reporting/src/render.rs
+++ b/compiler/hash-reporting/src/render.rs
@@ -279,11 +279,12 @@ impl ReportCodeBlock {
 
         let ReportCodeBlockInfo { start_row, end_row, start_col, end_col, .. } = self.info(modules);
 
-        // If the difference between the rows is longer than 5 lines, then we
-        // essentially begin to collapse the view by using `...` as the filler
-        // for those lines...
+        // If the difference between the rows is longer than `LINE_SKIP_THRESHOLD`
+        // lines, then we essentially begin to collapse the view by using `...`
+        // as the filler for those lines...
         let skip_lines_range = if end_row - start_row > LINE_SKIP_THRESHOLD {
-            Some((start_row + 3)..=(end_row - 3))
+            let mid = LINE_SKIP_THRESHOLD / 2;
+            Some((start_row + mid)..=(end_row - mid))
         } else {
             None
         };
--- a/compiler/hash-reporting/src/render.rs
+++ b/compiler/hash-reporting/src/render.rs
@@ -30,6 +30,10 @@ const BLOCK_DIAGNOSTIC_MARKER: char = '-';
 /// Character used to connect block views
 const DIAGNOSTIC_CONNECTING_CHAR: &str = "|";
 
+/// The maximum number of lines a block display can use before the lines in the
+/// center of the block are skipped.
+const LINE_SKIP_THRESHOLD: usize = 6;
+
 /// Struct to represent the column and row offset produced from converting a
 /// [hash_source::location::Span].
 pub(crate) struct ColRowOffset {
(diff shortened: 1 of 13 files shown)


compiler/hash-typecheck/src/diagnostics/error.rs
```diff
@@ -1225,6 +1228,30 @@ impl<'tc> From<TcErrorWithStorage<'tc>> for Report {
                     )));
                 }
             }
+            TcError::MethodNotAMemberOfTrait { trt_def_term_id, member, name } => {
+                builder.with_error_code(HashErrorCode::MethodNotAMemberOfTrait).with_message(
+                    format!(
+                        "method `{name}` is not a member of trait `{}`",
+                        trt_def_term_id.for_formatting_with_opts(
```
kontheocharis, line 1235: Should say `function`, unless it should also check presence of `self`.

compiler/hash-source/src/location.rs
```diff
@@ -94,3 +94,14 @@ pub struct SourceLocation {
     /// The id of the source that the span is referencing.
     pub id: SourceId,
 }
+
+impl SourceLocation {
+    /// Join the span of a [SourceLocation] with another [SourceLocation].
+    ///
+    /// *Note*: the `id` of both [SourceLocation]s must be the same.
+    pub fn join(self, other: Self) -> Self {
+        assert!(self.id == other.id);
```
kontheocharis, line 103: Why ?

compiler/hash-reporting/src/render.rs
```diff
@@ -275,14 +279,23 @@ impl ReportCodeBlock {
 
         let ReportCodeBlockInfo { start_row, end_row, start_col, end_col, .. } = self.info(modules);
 
+        // If the difference between the rows is longer than 5 lines, then we
+        // essentially begin to collapse the view by using `...` as the filler
+        // for those lines...
+        let skip_lines_range = if end_row - start_row > LINE_SKIP_THRESHOLD {
+            Some((start_row + 3)..=(end_row - 3))
```
kontheocharis, line 286: Magic constants? Are you meant to use `LINE_SKIP_THRESHOLD/2`?

compiler/hash-reporting/src/render.rs
```diff
@@ -300,6 +313,31 @@ impl ReportCodeBlock {
                 " "
             };
 
+            // So if we're at the start of the 'skip' range, use '...' instead
+            if let Some(range) = skip_lines_range.clone() {
+                if *range.start() == index {
+                    let range_line_number = format!(
+                        "{:<pad_width$}",
+                        ".".repeat(3),
+                        pad_width = longest_indent_width + 2
```
kontheocharis, line 322: Why +2?

compiler/hash-source/src/location.rs
feds01: How else can you join the spans if they have different source-ids?

compiler/hash-reporting/src/render.rs
feds01: For the `|` and the space between the line number and the separator.

compiler/hash-typecheck/src/diagnostics/error.rs
feds01: But it is a method associated with the trait?

kontheocharis: Well, even static functions are called methods ? If so then okay.

compiler/hash-source/src/location.rs
kontheocharis: Oh right, i was just unclear on what the ID was for. So it is SourceId ? Maybe that field should be renamed..

compiler/hash-reporting/src/render.rs
kontheocharis: Also maybe update comment above to use it instead of `5`

compiler/hash-typecheck/src/diagnostics/error.rs
feds01: Yes sir

compiler/hash-source/src/location.rs
feds01: It was previously named `source_id` but I thought that was too verbose, I guess if it does create confusion then maybe we can revert

kontheocharis (approved):

PR #496, opened 2022-08-23
TC: Attach type information to nodes

### LINKED ISSUE #240: LM: AST traversal to typecheck and add types to nodes

kontheocharis:
This PR adds a store `NodeInfoStore` which maps `AstNodeId`s to `TermId`s and `PatId`s. It is a necessary step for implementing mono/lowering to IR, as well as any post-TC semantic stages.

Closes #240.

### DIFF
--- a/compiler/hash-typecheck/src/traverse/mod.rs
+++ b/compiler/hash-typecheck/src/traverse/mod.rs
@@ -16,7 +16,7 @@ use hash_source::{
     location::{SourceLocation, Span},
     ModuleKind, SourceId,
 };
-use hash_utils::store::Store;
+use hash_utils::store::{PartialStore, Store};
 use itertools::Itertools;
 
 use self::scopes::VisitConstantScope;
@@ -29,6 +29,7 @@ use crate::{
     ops::{scope::ScopeManager, AccessToOps},
     storage::{
         location::{IndexedLocationTarget, LocationTarget},
+        nodes::NodeInfoTarget,
         pats::PatId,
         primitives::{
             AccessOp, Arg, BindingPat, ConstPat, EnumVariant, Field, Member, ModDefOrigin,
@@ -133,7 +134,7 @@ impl<'tc> TcVisitor<'tc> {
     /// Copy the [SourceLocation] of the given [hash_ast::ast::AstNode] to the
     /// given [LocationTarget].
     pub(crate) fn copy_location_from_node_to_target<N>(
-        &mut self,
+        &self,
         node: AstNodeRef<N>,
         target: impl Into<LocationTarget>,
     ) {
@@ -145,7 +146,7 @@ impl<'tc> TcVisitor<'tc> {
     /// the given [LocationTarget] list represented by a type `Target` where
     /// `(Target, usize)` implements [Into<LocationTarget>].
     pub(crate) fn copy_location_from_nodes_to_targets<'n, N: 'n>(
-        &mut self,
+        &self,
         nodes: impl IntoIterator<Item = AstNodeRef<'n, N>>,
         targets: impl Into<IndexedLocationTarget> + Clone,
     ) {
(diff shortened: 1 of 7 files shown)


tests/cases/typecheck/property_access/numeric_access_on_struct.stderr
```diff
@@ -1,6 +1,6 @@
 error[0015]: the field `2` is not present within `Frobulate(data = 23_i32, other = 'c')`
- --> $DIR/numeric_access_on_struct.hash:8:6
+ --> $DIR/numeric_access_on_struct.hash:8:1
 7 |   
 8 |   t := Frobulate(data = 23);
-  |        ^^^^^^^^^^^^^^^^^^^^ `Frobulate(data = 23_i32, other = 'c')` does not contain the property `2`
```
feds01, line 5: Is this not a regression?

tests/cases/typecheck/structs/missing-fields.stderr
```diff
@@ -1,18 +1,12 @@
 error[0017]: struct literal is missing the fields `age`, and `width`
-  --> $DIR/missing-fields.hash:14:5
-13 |   
-14 |       Dog(name, height) := viktor;
```
feds01, line 4: Regression?

feds01 (commented):
Why have a bunch of the test cases changed?

feds01:
Also, the entries that map `AstNodeId` need to be added when traversing.

And, the tc state needs to be somehow externally accessible

tests/cases/typecheck/property_access/numeric_access_on_struct.stderr
kontheocharis: It possibly makes more sense since those errors are due to variable access.. I can change it back to the way it was too.

kontheocharis:
> And, the tc state needs to be somehow externally accessible, which means all the tc data structures need to be public or in their own crate?

@feds01 Yeah, I guess that is the next step once these entries will actually be used. However, I think that whichever crate needs them should probably just depend on TC. So they could just stay where they are in global storage for now..

kontheocharis:
Yeah lowering should definitely depend on TC, because that is the utility it uses to actually know what to lower it to... I guess we can split TC impl from TC primitives into different crates but that might not bring so many advantages.

compiler/hash-typecheck/src/traverse/mod.rs
```diff
@@ -391,11 +400,9 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
...
             CORE_IDENTIFIERS.r#false
         });
-
-        // add the location of the term to the location storage
-        self.copy_location_from_node_to_target(node, term);
-
-        Ok(self.validator().validate_term(term)?.simplified_term_id)
+        let term = self.validator().validate_term(term)?.simplified_term_id;
+        self.register_node_info(node, term);
+        Ok(term)
```
feds01, line 405: validation occurs before location is set on term... although that might not be relevant here

tests/cases/typecheck/spread-pats/tuple-spread-pats.stderr
```diff
@@ -1,6 +1,12 @@
...
 28 |   test_all_spread := (f: (name: str, bax: char)) -> (name: str, bax: char) => {
+29 |       (...t) := f;
+   |       ^^^^^^ the named tuple type `{value (name: str, bax: char)}` is being coerced into an un-named tuple
+30 |       t
+
+  --> $DIR/tuple-spread-pats.hash:29:5
+28 |   test_all_spread := (f: (name: str, bax: char)) -> (name: str, bax: char) => {
 29 |       (...t) := f;
    |       ^^^^^^ the coercion occurs here
 30 |       t
```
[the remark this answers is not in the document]
kontheocharis: Hmm.. You are right

compiler/hash-typecheck/src/traverse/mod.rs
kontheocharis: Good catch!

feds01 (approved):

PR #502, opened 2022-08-25
TC: Add unit nominal primitive

### LINKED ISSUE #501: TC: Add unit nominal type
This is a prerequisite for #363.

kontheocharis:
This PR adds `NominalDef::Unit`. This is meant to be to constant enum variants what structs are to enum variants with data. Hopefully that makes sense. It is equivalent in spirit to a struct with an empty field list, or a simply a unit struct in rust `struct Foo;`.

In subsequent commits, this will be used to replace enums; enum variants will become either structs or units, depending on whether they have data members or not.

Closes #501.

### DIFF
--- a/compiler/hash-typecheck/src/exhaustiveness/lower.rs
+++ b/compiler/hash-typecheck/src/exhaustiveness/lower.rs
@@ -181,41 +181,38 @@ impl<'tc> LowerPatOps<'tc> {
                 match reader.get_term(ty) {
                     Term::Level1(Level1Term::NominalDef(nominal_def)) => {
                         // @@Todo: deal with variants
-                        let (ctor, members) = match reader.get_nominal_def(nominal_def) {
+                        match reader.get_nominal_def(nominal_def) {
                             NominalDef::Struct(struct_def) => match struct_def.fields {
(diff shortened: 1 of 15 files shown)


compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -180,36 +180,43 @@ impl<'tc> LowerPatOps<'tc> {
...
                             },
+                            NominalDef::Unit(_) => (DeconstructedCtor::Single, None),
                             // @@EnumToUnion: when enums aren't a thing, do this with a union
                             NominalDef::Enum(_) => unreachable!(),
                         };
 
-                        // Lower the fields by resolving what positions the actual fields are
-                        // with the reference of the constructor's type...
-                        let fields = self.pat_lowerer().deconstruct_pat_fields(args, members);
+                        if let Some(members) = members {
```
feds01, line 198: It would be nice if this was short-circuiting...
```suggestion
                        let members = Some(members) = members else {
                             return (ctor, vec![]);
                        }
```

feds01 (approved):

PR #505, opened 2022-08-26
TC: Polish trait validation checking and error reporting

### LINKED ISSUE #389: TC: Properly ensure impls implement their traits, and infer original types while at it

feds01:
closes #389

### DIFF
--- a/compiler/hash-typecheck/src/diagnostics/error.rs
+++ b/compiler/hash-typecheck/src/diagnostics/error.rs
@@ -1208,7 +1208,7 @@ impl<'tc> From<TcErrorWithStorage<'tc>> for Report {
 
                 builder.with_error_code(HashErrorCode::TraitImplMissingMember).with_message(
                     format!(
-                        "trait `{}` is missing the trait item{} {missing_members}",
+                        "trait `{}` is missing the member{} {missing_members}",
                         trt_def_term_id.for_formatting(ctx.global_storage()),
                         pluralise!(missing_trt_members.len())
                     ),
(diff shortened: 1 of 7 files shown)


compiler/hash-typecheck/src/diagnostics/error.rs
```diff
@@ -1189,46 +1189,50 @@ impl<'tc> From<TcErrorWithStorage<'tc>> for Report {
...
+                // Create a sequence display for displaying member names
+                let missing_members = SequenceDisplay::new(
+                    &missing,
+                    SequenceDisplayOptions::with_limit(SequenceJoinMode::All, 6),
+                );
+
                 builder.with_error_code(HashErrorCode::TraitImplMissingMember).with_message(
                     format!(
-                        "trait `{}` is missing the member `{}`",
+                        "trait `{}` is missing the trait item{} {missing_members}",
```
kontheocharis, line 1211: missing space? Also maybe "member" instead of "trait item"?

kontheocharis (approved):

compiler/hash-typecheck/src/diagnostics/error.rs
feds01: No missing space, `pluralise!` basically converts the number into a suffix for a given word, which means that `1` yields nothing and anything greater yields `s`.

feds01: the two phrases are synonymous so I don't mind which one we use

kontheocharis: Probably member is best for the sake of consistency

kontheocharis (approved):

PR #506, opened 2022-08-28
TC: Implement enums as unions

kontheocharis:
This PR implements the basics of enums-as-unions. To do this, a couple of modifications were made:
1. Accessing on unions now looks for the nominal in the union matching the accessed name. So `Foo := Red | Green | Blue; Foo::Red` works out of the box.
2. For now, instantiating unit structs requires `()` after it, so `true(), None()` etc. This is hopefully temporary and is due to the difficulty in differentiating between the type of the unit and the (singular) value.

Another thing that needs to be done is type enum access as the enum type rather than the variant type, under certain conditions (i.e. when the variable is mutable). This will be done in a future PR.


feds01 (commented):
Using `()` to instantiate unit structs is a deal breaker in my opinion

kontheocharis:
> Using `()` to instantiate unit structs is a deal breaker in my opinion

Yeah no I agree, the question is how to get around this tho..

feds01 (dismissed):
This looks good to me, however, I still don't understand the problem of needing to instantiate unit structs with `()`. Could you give some more detail as to why this is currently occurring?

kontheocharis:
Well, since unit structs are also types,
```
X := Colour::Red
```
is ambiguous since it either means the type Red or the unit value Red. Sort of like how there is the void type () and the empty tuple value ().

kontheocharis:
This problem can be fixed by making the simplifier aware of context, so that `type Colour::Red` means the type, and `Colour::Red` means the unit value

kontheocharis:
Fixed this problem by hooking into `visit_expr` and coercing to value. This pattern can be expanded to more things in the future!

feds01 (approved):
Looks good, we could also look at simplifying the term structure with this change in place

PR #507, opened 2022-08-31
TC+IR: initial sketch of IR Builder and move type storage into `hash_types`

feds01:
- Simplify `visitor` pattern for ast using macros to derive mutable and immutable variants.
- add basic constructs to IR and introduce the `hash_lower` crate
- split typechecking storage into the `hash_types` crate


.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
 [build]
 rustdoc = "rustdoc"
 rustdocflags = ["--show-type-layout", "-Z", "unstable-options"]
+
+[unstable]
+sparse-registry = true
```
kontheocharis, line 6: what's this

compiler/hash-ir/src/ir.rs
```diff
@@ -1,59 +1,71 @@
...
+    /// `i32` type, 32bit signed integer
+    I32,
+    /// `i64` type, 64bit signed integer
+    I64,
+    /// `f32` type, 32bit float
+    F32,
+    /// `f64` type, 64bit float
+    F64,
+    /// `char` type, 32bit characters
+    Char,
```
kontheocharis, line 40: Maybe this can just be a u32 like we said

compiler/hash-ir/src/ir.rs
```diff
@@ -1,59 +1,71 @@
...
+    /// A `void` type
+    Void,
+    /// Represents any collection of types in a specific order.
+    Structural(&'i [Ty<'i>]),
+    /// Essentially an enum representation
```
kontheocharis, line 45: Maybe a better comment? Do values of this have a discriminant

compiler/hash-ir/src/ir.rs
```diff
@@ -1,59 +1,71 @@
...
+    Union(&'i [Ty<'i>]),
+    /// Pointer that points to some type
```
kontheocharis, line 47: Type for a pointer to some value ?

kontheocharis: Also any reason to not use the Ref terminology here?

compiler/hash-ir/src/ir.rs
```diff
@@ -1,59 +1,71 @@
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Const {
+    /// Character constant
+    Char(char),
+    /// Integer constant that is defined within the program source.
+    Int(InternedInt),
+    /// Float constant that is defined within the program source.
+    Float(InternedFloat),
+
+    /// String has to become a struct that has a pointer
```
kontheocharis, line 62: Maybe this is a misleading comment? These are static strings no?

compiler/hash-ir/src/ir.rs
```diff
@@ -117,77 +129,183 @@ pub enum Mutability {
...
-    /// An expression which represents a re-assignment to a pattern with a right
-    /// hand-side expression and a binary operator that combines assignment
-    /// and an operator, e.g. `x += 2`
-    AssignOp(&'i Pat<'i>, BinOp, &'i Expr<'i>),
+    Assign(Local, &'i Statement<'i>),
+
     /// An expression which is taking the address of another expression with an
     /// mutability modifier e.g. `&mut x`.
-    AddrOf(Mutability, &'i Expr<'i>),
+    Ref(Mutability, &'i Statement<'i>, AddressMode),
```
kontheocharis, line 187: What are the valid statements that can be used here?

compiler/hash-ir/src/ir.rs
```diff
@@ -117,77 +129,183 @@ pub enum Mutability {
...
+
+    /// Allocate some value on the the heap using reference
+    /// counting.
+    Alloc(Local),
```
kontheocharis, line 191: What about non-rc? Also why does this take a local?

compiler/hash-ir/src/ir.rs
```diff
@@ -117,77 +129,183 @@ pub enum Mutability {
...
+pub enum TerminatorKind {
+    /// A simple go to block directive.
+    Goto(BasicBlock),
+
+    /// Return from the parent function
+    Return,
+
+    /// Perform a function call
+    Call {
+        // op: todo!(),
```
kontheocharis, line 215: What is op?

compiler/hash-ir/src/ir.rs
```diff
@@ -117,77 +129,183 @@ pub enum Mutability {
...
+        /// Where to return after completing the call
+        target: Option<BasicBlock>,
+    },
+
+    /// Denotes that this terminator should never be reached, doing so will
+    /// break IR control flow invariants.
+    Unreachable,
+
+    /// Essentially a `jump if <0> to <1> else go to <2>`
+    Switch(Local, Vec<(Const, BasicBlock)>, BasicBlock),
```
kontheocharis, line 227: Could this be smallvec? Or maybe &'i slice?

compiler/hash-lower/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use hash_ir::ir::Body;
+use hash_pipeline::traits::Lowering;
+use hash_source::{
+    location::{SourceLocation, Span},
+    SourceId,
+};
+
+/// The [IrLowerer] is used as a bootstrapping mechanism to kick off the
+/// [Builder] working on function that it discovers as the the lower traverses
```
kontheocharis, line 19: functions

compiler/hash-typecheck/src/exhaustiveness/mod.rs
```diff
@@ -132,3 +135,32 @@ pub(crate) trait AccessToUsefulnessOps: AccessToStorage {
 }
 
 impl<T: AccessToStorage> AccessToUsefulnessOps for T {}
+
+/// Wraps a type `T` in a structure that contains information to be able to
+/// format `T` using [TcFormatter].
+///
+/// This can wrap any type, but only types that have corresponding `fmt_*`
+/// methods in [TcFormatter] are useful with it.
+pub struct PatForFormatting<'tc, T> {
```
kontheocharis, line 144: Why is this needed?

compiler/hash-typecheck/src/ops/reader.rs
```diff
@@ -1,105 +1,102 @@
...
+        exhaustiveness::{DeconstructedCtorId, DeconstructedPatId},
+        StorageRef,
     },
 };
 
 /// Helper to read various primitive constructions (from
 /// [crate::storage::primitives]).
-pub struct PrimitiveReader<'gs> {
-    gs: &'gs GlobalStorage,
+pub struct PrimitiveReader<'tc> {
```
kontheocharis, line 27: This can implement AccessToStorage and then you can use the shorthand methods in AccessToOps!

kontheocharis (changes requested):

compiler/hash-typecheck/src/ops/reader.rs
feds01: Yeah, before it was just relying on a reference to `gs`, but now due to this refactor, we can now convert it

compiler/hash-typecheck/src/exhaustiveness/mod.rs
feds01: Since `exhaustiveness_store` and `global_storage` are now split, but the formatting for types doesn't interact with `exhaustiveness_storage`

compiler/hash-ir/src/ir.rs
feds01: Could be indeed

compiler/hash-ir/src/ir.rs
feds01: This will essentially be the operand of the call, so the function type.

.cargo/config.toml
feds01: just switches the cargo indexing to use `http` rather than their default git one, it makes ci builds and package fetching a lot faster.

feds01: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#sparse-registry

compiler/hash-ir/src/ir.rs
feds01: what else would it take?

compiler/hash-ir/src/ir.rs
feds01: Yeah, well all of the strings here will be static, everything else is a pointer

compiler/hash-ir/src/ir.rs
feds01: Probably `Local` and `Const`... 

Additionally, we could eliminate the `Index` StatementKind since it can be represented as some kind ascription on a local

.cargo/config.toml
kontheocharis: Are we using any of these?

compiler/hash-ir/src/ir.rs
kontheocharis: Maybe there should be a note for that. Also what do you mean by ascription?

compiler/hash-ir/src/ir.rs
kontheocharis: What are the semantics of this? It does memcpy of the local into the location? Would it be possible to have any statement in there (prevent unnecessary copy to local first)?

.cargo/config.toml
feds01: well no it's just how packages are fetched

compiler/hash-ir/src/ir.rs
kontheocharis: You mean the pointer to the function location?

compiler/hash-ir/src/ir.rs
feds01: not *ascription*, more like a transformation, similarly to how we should represent property access

compiler/hash-typecheck/src/exhaustiveness/mod.rs
kontheocharis: Why is exhaustiveness_store separate from global_storage?

compiler/hash-ir/src/ir.rs
feds01: what do you mean by the discriminant?

.cargo/config.toml
kontheocharis: Alright then

compiler/hash-ir/src/ir.rs
kontheocharis: For the union, like union values. My question is basically, what is the value of a Union ?

feds01: the value of a union is an initial discriminant byte or bytes to denote which variant it is, proceeded by the value stored in the variant.

compiler/hash-typecheck/src/exhaustiveness/mod.rs
feds01: I didn't think it made sense putting `DeconstructedCtor` and `DeconstructedPat` into `hash_types` since these types are implementation detail for the typechecker exhaustiveness checker.

compiler/hash-ir/src/ir.rs
feds01: No, just the type that will be called, it doesn't actually matter where the function is "currently" in the IR, that is to be resolved later...

kontheocharis (approved):

PR #512, opened 2022-09-15
IR: add `Place`s `RValue`s and other needed constructs

feds01:


compiler/hash-ir/src/ir.rs
```diff
@@ -155,33 +160,77 @@ pub enum AddressMode {
...
+    pub projections: Vec<PlaceProjection>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum AggregateKind {
+    Tuple,
+    Array(TermId),
+    Enum(NominalDefId, usize),
+    Struct(NominalDefId),
+}
```
kontheocharis, line 193: So type information is available in the IR?

compiler/hash-ir/src/ir.rs
```diff
@@ -155,33 +160,77 @@ pub enum AddressMode {
...
+    /// copy/move?
+    Use(Place),
+
+    /// Compiler intrinsic operation, this will be computed in place and
+    /// replaced by a constant.
+    ///
+    /// @@Future: maybe in the future this should be replaced by a compile-time
+    /// API variant which will just run some kind of operation and return the
+    /// constant.
+    ConstOp(ConstOp, &'ir RValue<'ir>),
```
kontheocharis, line 210: How does sizeof work? How does it know the type of the value

compiler/hash-ir/src/ir.rs
feds01: Yeah it will need to be available in order to query information about fields of types, layout, etc.

compiler/hash-ir/src/ir.rs
feds01: There is an associated term with the value, so it can always compute the size of the term, that is if the term is "sized", otherwise it should panic since we shouldn't be getting un-sized type arguments here

kontheocharis: But even still, given any RValue, how can one procedurally figure out its type? (And thus its size)

feds01: It should probably just take a `TermId`

kontheocharis: Hmm, is it a good idea to let TC concepts leak into here? Isn't there a way of having a more generalised way to represent type layouts (separate crate) and use that in here rather than relying on all the TC machinery?

feds01: Well, it would need to convert to some other representation of types before it reaches this stage. In my opinion, this is quite the task since it would mean that we would have to represent the whole type system in a different representation than the current one. 

I think this work could be done in the future (where this system is as simple as possible, and erases most un-needed things, possibly even smaller in terms of memory footprint), but I don't think this is the current priority.

kontheocharis: The only problem is that having TC types in the IR means that the IR needs to be able to handle all of the representable types (or alternatively assert that only a subset of the types exist at this stage). Surely not the entire type system, but rather a small part of it, needs to be representable in the IR.

kontheocharis (approved):

PR #514, opened 2022-09-15
parser: reduce size of `ParseError` and properly detect unclosed delimiters

feds01:


kontheocharis (approved):
nice one sir

feds01:
> nice one sir

It's quite the reduction: from `136bytes` to `72bytes`! and could be even more in the future

kontheocharis (approved):

PR #515, opened 2022-09-15
semantics+tc: verify no `self` in free standing functions

feds01:


compiler/hash-ast-passes/src/visitor.rs
```diff
@@ -600,6 +597,32 @@ impl AstVisitor for SemanticAnalyser<'_> {
...
         let _ = walk::walk_param(self, ctx, node);
 
+        if matches!(node.origin, ParamOrigin::Fn) {
+            match self.current_block {
+                // Check that `self` cannot be within a free standing functions
+                BlockOrigin::Root | BlockOrigin::Mod => {
+                    if let Some(name) = node.name.as_ref() && name.is(CORE_IDENTIFIERS.self_i) {
+                        self.append_error(AnalysisErrorKind::SelfInFreeStandingFn, node);
+                    }
+                }
```
kontheocharis, line 607: Mod is valid having self as it is exactly the same as `impl`. Unless we want to change that..

kontheocharis:
What do you think about renaming `hash-ast-passes`?

feds01:
What should the name be?

compiler/hash-ast-passes/src/visitor.rs
feds01: What does `self` reference in `mod`?

kontheocharis: Well, `Foo := struct() ~ mod {..}` is completely the same as `Foo := struct() ~ impl {..}` with the current spec. The only difference is that members of `mod` are private by default, and members of `impl` are public by default. But in both cases writing `self` inside would refer to the struct.

kontheocharis:
Maybe `hash-untyped-semantics`? And then we can also have `hash-typed-semantics`?

feds01:
> Maybe `hash-untyped-semantics`? And then we can also have `hash-typed-semantics`?

So is `untyped-semantics` this crate, and then `typed-semantics` is the typechecker

compiler/hash-ast-passes/src/visitor.rs
feds01: From what we have discussed, is this subject to change in the future?

kontheocharis: No

kontheocharis:
Well no, I meant like after typechecking, maybe extra checks can happen inside another crate instead of polluting the typechecker.

feds01:
> Well no, I meant that after typechecking, maybe extra checks can happen inside another crate instead of polluting the typechecker.

I think this could be a pretty good separation, and we could possibly move some of the current typechecking stuff out of the typechecker and move it into this crate. However, before we do that, we would need to re-tool the pipeline and avoid each stage having a specific name, owned data, etc.

kontheocharis (approved):

PR #516, opened 2022-09-15
Parser: handle trailing semi colons gracefully

feds01:


compiler/hash-ast/src/ast.rs
```diff
@@ -1826,6 +1830,8 @@ pub enum ExprKind {
     BinaryExpr(BinaryExpr),
     /// Unary Expression composed of a unary operator and an expression
     UnaryExpr(UnaryExpr),
+    /// Empty expression, just an expression that is a trailing semi-colon
+    Empty(EmptyExpr),
```
kontheocharis, line 1834: Do we really need a new AST variant for this? Can't it be removed during parsing?

feds01: Yeah I thought this might be better if it was just removed. The function `parse_top_level_expr()` could be changed tk just to skip it and then try parse a expression.

feds01: Actually, there is a bit of a problem with this approach. The problem is that if there are trailing semi-colons and then it hits an `EOF` within the current token tree, it cannot parse a declaration or an expression and thus fails. I don't see any way (at least a non-elegant way) to solve this problem without having a filler `Expr::Empty` variant.

kontheocharis: what if the parse function returns an optional?

kontheocharis (approved):

PR #520, opened 2022-09-21
AST: Add scaffolding and definitions for the AST definition macro

kontheocharis:
The macro is named `define_tree!` and it essentially takes all the items of the AST definition and (will eventually) produce visitor and walker definitions for them. This is an intermediate PR to keep the code size small; the parsing has been implemented but the emitting has not.

Working towards #519.


compiler/hash-tree-def/src/parsing.rs
```diff
@@ -0,0 +1,274 @@
...
+/// Helper type so that we can implement [`TryFrom<&Item>`] for optional
+/// [`TreeDefOpts`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct MaybeTreeDefOpts(Option<TreeDefOpts>);
+
+impl TryFrom<&Item> for MaybeTreeDefOpts {
+    type Error = syn::Error;
+    fn try_from(value: &Item) -> Result<Self, Self::Error> {
+        // Options are given by a macro define_tree_opts! { ... }
+        println!("{:#?}", value);
```
feds01, line 224: remove?

compiler/hash-tree-def/src/parsing.rs
```diff
@@ -0,0 +1,274 @@
...
+                        let node_name = ensure_single_type_argument(segment)?;
+                        Ok(StructNodeFieldData::ChildList { node_name })
+                    } else {
+                        Ok(StructNodeFieldData::Other { ty: value.clone() })
+                    }
+                }
+                _ => Ok(StructNodeFieldData::Other { ty: value.clone() }),
+            },
+            _ => Ok(StructNodeFieldData::Other { ty: value.clone() }),
+        }
```
feds01, line 129: would be nice to use `if_chain!` here but not necessarily required.

feds01 (changes requested):
mui bien sie

feds01 (approved):

compiler/hash-tree-def/src/parsing.rs
kontheocharis: Will tackle in final PR

PR #521, opened 2022-09-22
tc: Implement traversal for assigning with operands

### LINKED ISSUE #498: TC: Implement assign op expr in traversal

feds01:
- split `SkopeKind::Constant` into each specific variant of a constant scope kind. This is needed for checking the initialisation rules of declarations within constant scopes.
- verify that items cannot be declared as `mut` within constant blocks.
- verify that variables must be declared as `mut` for assign operations, and assigning operations with operators.
- closes #498

### DIFF
--- a/compiler/hash-typecheck/src/traverse/visitor.rs
+++ b/compiler/hash-typecheck/src/traverse/visitor.rs
@@ -1679,7 +1679,7 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
 
         // We want to create the type that represents this operation, and then
         // ensure that it typechecks...
-        let ty = self.create_operator_fn(member.ty(), rhs, node.operator.ast_ref(), true);
+        let ty = self.create_operator_fn(var_term, rhs, node.operator.ast_ref(), true);
         let _ = self.validate_and_register_simplified_term(node, ty)?;
 
         // Now check that the declared local item is declared as mutable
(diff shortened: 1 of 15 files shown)


kontheocharis (approved):

compiler/hash-typecheck/src/traverse/visitor.rs
```diff
@@ -1640,80 +1654,65 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
...
-        };
+        let ScopeMember { member, scope_id, index } =
+            self.scope_manager().resolve_name_in_scopes(name, var_term)?;
 
-        let lazy_operator_fn = |visitor: &mut Self, trait_name: &str| -> TcResult<TermId> {
-            let lhs_ty = visitor.typer().infer_ty_of_term(lhs)?;
-            let rhs_ty = visitor.typer().infer_ty_of_term(rhs)?;
+        // We want to create the type that represents this operation, and then
+        // ensure that it typechecks...
+        let ty = self.create_operator_fn(member.ty(), rhs, node.operator.ast_ref(), true);
```
kontheocharis, line 1682: Why are you getting its type? Aren't you meant to get the member's value?

kontheocharis (changes requested):

compiler/hash-typecheck/src/traverse/visitor.rs
feds01: yes this is wrong, can we assume that `.value()` will always yield a value?

kontheocharis: No, but you can just pass in the member using `Var` term.

kontheocharis (approved):

PR #526, opened 2022-09-24
Implement a macro to generate an AST (or any other tree definition)

### LINKED ISSUE #519: Make an AST macro that generates visitor/walker/defs

kontheocharis:
This generates a visitor (mut and non-mut), and a walker with types (mut and non-mut). In the future it can be extended to add more visitors/folders/etc with much ease.

Closes #519.

Next PR will apply this to the AST!


compiler/hash-tree-def/src/lib.rs
```diff
@@ -3,20 +3,33 @@
...
+        match $x {
+            Ok(value) => value,
+            Err(error) => return error.to_compile_error().into(),
+        }
+    };
+}
+
 ///! Define a tree structure.
 ///
 /// @@Todo: docs
```
feds01, line 28: todo here?

feds01 (dismissed):

feds01 (approved):

PR #531, opened 2022-09-26
lexer: un-intern numeric literals

feds01:
This patch reverses the behaviour of the lexer from interning `-` with numeric literals. This change is reverted due to the overeagerness of this approach when it comes to binary expressions. It is reasonably trivial to create a case where the interning approach does not correctly intern at the right time which causes valid syntax to cause a syntactic error.


compiler/hash-source/src/constant.rs
```diff
@@ -20,12 +20,22 @@ use crate::identifier::{Identifier, CORE_IDENTIFIERS};
 #[derive(Debug, Clone, Copy)]
 pub struct FloatConstant {
     /// Raw value of the float
-    pub value: f64,
+    pub value: [u8; 8],
```
kontheocharis, line 23: Why not store as f64?

feds01: When the value is `f32`, it would inappropriate to store it as an `f64`. Both variants should be stored as a byte scalar, which is then appropriately used as required. This is particularly important for operations that occur on the constants, later on in the compilation stage.

kontheocharis: Why not `enum FloatConstant { F32(f32), F64(f64) }` ?

feds01: I think this is a better approach for now, however it would be nice to later unify all these stores to just use a scalar. Especially in the future when `constructors` are also stored as constants.

kontheocharis: How would constructors be stored as constants? I agree about the unification but not needed yet.

feds01: Constructors should be stored as a byte scalar which has an attached layout of how it is represented.

kontheocharis: But that is only relevant at IR? This is parsing related no?

compiler/hash-lexer/src/lib.rs
```diff
@@ -537,28 +530,25 @@ impl<'a> Lexer<'a> {
...
-                            let (value, sign) =
-                                if negated { (-parsed, Sign::Minus) } else { (parsed, Sign::None) };
-
                             // Create interned float constant
-                            let float_const = CONSTANT_MAP.create_float_constant(value, suffix);
-                            TokenKind::FloatLit(sign, float_const)
+                            let float_const = if let Some(suffix_ident) = suffix && suffix_ident == CORE_IDENTIFIERS.f32 {
+                                CONSTANT_MAP.create_f32_float_constant(parsed as f32, suffix)
+                            } else {
+                                CONSTANT_MAP.create_f64_float_constant(parsed, suffix)
```
kontheocharis, line 537: Don't you need to check that the suffix is either empty or f64?

kontheocharis (approved):

compiler/hash-lexer/src/lib.rs
feds01: invalid suffixes are later caught by the parser, and if there is no suffix then it is automatically assumed as `f64`.

PR #533, opened 2022-09-26
parser: remove `;` requirement after expressions

feds01:
This patch removes the requirement for all expressions to have a `;` semi-colon present to signify the end of the expression. However, this does not mean that the parser does not support the semi-colon anymore. 

The semi-colon is still present and handled by the parser in order to terminate expressions. It is unclear if it should be removed entirely later, but it will be supported until there is some way to disambiguate unary operators and binary expressions.


kontheocharis:
Maybe remove semis from tests ?

kontheocharis (approved):

PR #539, opened 2022-09-29
pipeline: use general `CompilerStage` for representing compiler stages

feds01:
This patch moves away from the `Compiler` using traits that implement abstract `Stage` interfaces and just to a single `CompilerStage`. This significantly simplifies the pipeline code and makes it more standard per stage.

### DIFF
--- a/compiler/hash-pipeline/src/lib.rs
+++ b/compiler/hash-pipeline/src/lib.rs
@@ -55,13 +55,11 @@ pub struct Compiler<'pool, P, D, S, C, L, V> {
 /// instance. Each stage of the compiler contains a `State` type parameter which
 /// the compiler stores so that incremental executions of the compiler are
 /// possible.
-pub struct CompilerState<'c, C: Tc<'c>> {
+pub struct CompilerState {
     /// The collected workspace sources for the current job.
     pub workspace: Workspace,
     /// Any diagnostics that were collected from any stage
     pub diagnostics: Vec<Report>,
-    /// The typechecker state.
-    pub tc_state: C::State,
 }
 
 impl<'c, 'pool, P, D, S, C, L, V> Compiler<'pool, P, D, S, C, L, V>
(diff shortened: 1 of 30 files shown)


compiler/hash-pipeline/src/lib.rs
```diff
@@ -15,127 +13,61 @@ pub mod traits;
...
+    stages: Vec<Box<dyn CompilerStage<'pool>>>,
+
     /// Various settings for the compiler.
     pub settings: CompilerSettings,
     /// The pipeline shared thread pool.
     pool: &'pool rayon::ThreadPool,
 
     /// A record of all of the stage metrics
-    metrics: HashMap<CompilerMode, Duration>,
+    metrics: HashMap<CompilerStageKind, Duration>,
```
kontheocharis, line 39: What if two different stages have the same kind?

compiler/hash-pipeline/src/sources.rs
```diff
@@ -247,6 +286,11 @@ impl Workspace {
         &self.source_map
     }
 
+    /// Get a mutable reference to [SourceMap]
+    pub fn source_map_mut(&mut self) -> &mut SourceMap {
```
kontheocharis, line 290: Are all these getters necessary?

tests/cases/semantics/directives/unknown_directive.hash
```diff
@@ -0,0 +1,17 @@
+// run=pass, stage=semantic
+
+Bar := #erase trait<T> {
+    into_bar: (T) -> Self;
```
kontheocharis, line 4: Are semis still necessary here?

kontheocharis (approved):

compiler/hash-pipeline/src/lib.rs
feds01: Multiple stages are allowed to have the same `CompilerStageKind`, and for the metrics, it will essentially be summed.

kontheocharis (approved):

PR #540, opened 2022-09-29
ast-expand: `#dump_ast` directive now allows user to dump the AST for a particular expression

### LINKED ISSUE #541: Add a directive to "dump" generated AST for expressions

feds01:
This patch adds a new stage to the compiler called `hash-ast-expand` which will perform any expansions that need to happen for macros (which will be added in the future) or directives which are essentially a superset of macros, but allow the user to perform more powerful metaprogramming operations.

Addditionally, this patch refactors how line/column numbers are calculated, and where the logic is housed for this:

> Previously, when the compiler needed to convert `SourceLocation`s
and `Span`s into their respective row and columns values for line
number information, it was being done as a second thought in the
reporting sub-system.

> Now that there has been the need to access this logic from other crates,
it has been decided that this logic should live in `hash-source` and
be accessible by everyone who requires converting locations into
row and column values. This means that:

> - computing `row` and `column` values logic now lives in `hash-source`.
> - `RowColSpan` is a structure that represents a `Span` but with row
and column values.
> - `RowCol` is a single unit data structure to group a `row` and
`column` value.

closes #541

### DIFF
--- a/compiler/hash-pipeline/src/sources.rs
+++ b/compiler/hash-pipeline/src/sources.rs
@@ -78,6 +78,11 @@ impl Module {
         &self.path
     }
 
+    /// Get the canonicalised `path` from the [Module].
+    pub fn canonicalised_path(&self) -> String {
+        adjust_canonicalisation(self.path())
+    }
+
     /// Set the `node` for given [Module]
     pub fn set_node(&mut self, node: ast::AstNode<ast::Module>) {
         self.node = Some(node);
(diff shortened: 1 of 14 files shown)


compiler/hash-pipeline/src/sources.rs
```diff
@@ -215,6 +220,12 @@ pub struct Workspace {
     /// Stores all of the generated AST for modules and nodes
     pub node_map: NodeMap,
 
+    /// Sources that have passed from the `expansion` stage of the compiler.
+    /// @@Todo: Use bit-flags to represent which module has been
```
kontheocharis, line 224: is this Todo necessary

compiler/hash-source/src/location.rs
```diff
@@ -105,3 +108,117 @@ impl SourceLocation {
...
+pub fn compute_row_col_from_offset(offset: usize, source: &str, non_inclusive: bool) -> RowCol {
+    let source_lines = source.split('\n');
+
+    let mut bytes_skipped = 0;
+    let mut total_lines: usize = 0;
+    let mut last_line_len = 0;
+
+    let mut line_index = None;
+    for (line_idx, line) in source_lines.enumerate() {
+        // One byte for the newline
```
kontheocharis, line 180: what about windows ?

kontheocharis (dismissed):

kontheocharis (approved):

compiler/hash-source/src/location.rs
feds01: I think the best approach in the future is to `normalise` sources when they are first to read in so that we don't have to worry about them at all after reading them in.

PR #543, opened 2022-10-02
pipeline: re-work stage and stage data dependencies

feds01:
It is clear that each stage cannot depend on some global "compiler state" since this would create crate dependency problems, and in general would make crates un-clean.

The new approach allows for a `CompilerStage` to specify a trait that provides an interface for querying information from the `CompilerSession` which will be able to retrieve all of the information the stage needs. This way, all stages can remain "opaque" to `CompilerSession` and vice-versa.

This meant that a new crate called "hash-session" needed to be created. This crate contains the global compiler session and all of the implementations for the stage-specified context interfaces.

Additionally:
- moved `NodeMap` out of `hash_pipeline` and into `hash_ast`


compiler/hash-ast-desugaring/src/lib.rs
```diff
@@ -93,14 +102,11 @@ impl<'pool> CompilerStage<'pool> for AstDesugarer {
...
-        entry_point: SourceId,
-        workspace: &mut Workspace,
-        settings: &hash_pipeline::settings::CompilerSettings,
-    ) {
+    fn cleanup(&self, entry_point: SourceId, ctx: &mut Ctx) {
+        let settings = ctx.settings();
+
         if settings.stage > CompilerStageKind::Parse && settings.dump_ast {
-            workspace.print_sources(entry_point);
+            ctx.workspace_mut().print_sources(entry_point);
```
kontheocharis, line 109: this doesn't need to be mut?

compiler/hash-lower/src/lib.rs
```diff
@@ -26,19 +28,27 @@ use self::builder::Builder;
...
-    fn run_stage(
-        &mut self,
-        entry_point: SourceId,
-        workspace: &mut hash_pipeline::sources::Workspace,
-        pool: &'pool rayon::ThreadPool,
-    ) -> CompilerResult<()> {
-        // We need to iterate all of the modules and essentially perform
-        // a discovery process for what needs to be lowered...
+    fn run_stage(&mut self, entry_point: SourceId, ctx: &mut Ctx) -> CompilerResult<()> {
+        let _node_map = ctx.node_map();
```
kontheocharis, line 51: unneeded

compiler/hash-lower/src/lib.rs
```diff
@@ -26,19 +28,27 @@ use self::builder::Builder;
...
     }
 
-    fn run_stage(
-        &mut self,
-        entry_point: SourceId,
-        workspace: &mut hash_pipeline::sources::Workspace,
-        pool: &'pool rayon::ThreadPool,
-    ) -> CompilerResult<()> {
-        // We need to iterate all of the modules and essentially perform
-        // a discovery process for what needs to be lowered...
```
kontheocharis, line 41: why was this comment removed

compiler/hash-parser/src/lib.rs
```diff
@@ -179,38 +182,42 @@ impl<'pool> Parser {
...
-    ) {
-        // Any other stage than `semantic_pass` is valid when `--dump-ast` is specified.
+    /// Any other stage than `semantic_pass` is valid when `--dump-ast` is
+    /// specified.
+    fn cleanup(&self, entry_point: SourceId, ctx: &mut Ctx) {
+        let settings = ctx.settings();
+
         if settings.stage < CompilerStageKind::SemanticPass && settings.dump_ast {
-            workspace.print_sources(entry_point);
+            ctx.workspace_mut().print_sources(entry_point);
```
kontheocharis, line 220: unneeded mut

compiler/hash-pipeline/src/interface.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// might be needed to perform some additional operations when the stage
+    /// has completed.
+    ///
+    /// An example of this use case might be for the parser. The parser stage
+    /// checks whether the `--dump-ast` flag has been set within the
+    /// compiler settings, if the flag is specified, then the compiler must
+    /// emit the parsed AST. The parser does this by checking this condition
+    /// and then invoking a function to emit all of the ASTs for each module
+    /// within the workspace.
+    fn cleanup(&self, _entry_point: SourceId, _stage_data: &mut StageCtx) {}
```
kontheocharis, line 36: maybe it should take mut self?

compiler/hash-pipeline/src/lib.rs
```diff
@@ -108,14 +82,14 @@ impl<'pool> Compiler<'pool> {
     fn run_stage(
         &mut self,
         entry_point: SourceId,
-        workspace: &mut Workspace,
+        workspace: &mut W,
```
kontheocharis, line 85: rename variable to ctx

compiler/hash-pipeline/src/lib.rs
```diff
@@ -6,78 +6,52 @@
...
-use traits::CompilerStage;
 
 pub type CompilerResult<T> = Result<T, Vec<Report>>;
 
 /// The Hash Compiler interface. This interface allows a caller to create a
 /// [Compiler] with the specified components. This allows external tinkerers
 /// to add their own implementations of each compiler stage with relative ease
 /// instead of having to scratch their heads.
-pub struct Compiler<'pool> {
+pub struct Compiler<W: CompilerInterface> {
```
kontheocharis, line 29: rename generic parameter to I

compiler/hash-session/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        &self.workspace.source_map
+    }
+}
+
+impl ParserCtx for CompilerSession {
+    fn data(&mut self) -> (&mut Workspace, &rayon::ThreadPool) {
+        (&mut self.workspace, &self.pool)
+    }
+}
+
```
kontheocharis, line 115: Maybe a function to make the compiler stages here, and import that from main/tests? Also, it might be good to typedef `Vec<Box<dyn CompilerStage<..>>>` to `CompilerStages`?

kontheocharis (changes requested):

kontheocharis (approved):

PR #544, opened 2022-10-26
IR: initial work to hook up ir generation stage with the rest of the compiler

feds01:
This patch is a progress update on the IR generation process of the compiler; adding the initial steps to lower the source into Hash IR.

- a visitor pattern for the IR to "discover" items that need to be lowered
- IR builder work, sketch out the initial process of lowering functions
- use `bitflags!` within the pipeline to represent the "status" of a module, whether it has been typechecked, de-sugared, semantically checked, etc.
- ast: rename `Expr::LitExpr` to `Expr::Lit`

The next steps are to expose scopes from the typechecker, which will be done after `hash-types` crate is re-organised (split `lib.rs`) to hold relevant data structures to each file that is already within the crate.


compiler/hash-ast/src/origin.rs
```diff
@@ -0,0 +1,48 @@
...
+    Body,
+    /// A dummy [BlockOrigin] kind, only used for error
+    /// reporting and grouping `constant` blocks together.
+    Const,
+}
+
+impl BlockOrigin {
+    /// Convert the [BlockOrigin] into a string which can be used for displaying
+    /// within error messages.
+    fn to_str(self) -> &'static str {
```
kontheocharis, line 27: Why not directly impl Display?

compiler/hash-ast/src/origin.rs
```diff
@@ -0,0 +1,48 @@
...
+            BlockOrigin::Const => "constant",
+        }
+    }
+}
+
+impl Default for BlockOrigin {
+    fn default() -> Self {
+        BlockOrigin::Root
+    }
+}
```
kontheocharis, line 42: Is it useful to have a default impl for this?

compiler/hash-pipeline/src/workspace.rs
```diff
@@ -18,6 +19,121 @@ use hash_ast::{
 use hash_source::{InteractiveId, ModuleId, ModuleKind, SourceId, SourceMap};
 use hash_utils::tree_writing::TreeWriter;
 
+bitflags! {
```
kontheocharis, line 22: What if we want more than 32 flags?

compiler/hash-pipeline/src/workspace.rs
```diff
@@ -18,6 +19,121 @@ use hash_ast::{
...
+        Self(HashMap::new())
+    }
+
+    /// Add a new source to the [SourceStageInfo] with the given [SourceId].
+    pub fn add(&mut self, source_id: SourceId, stage: SourceStageInfo) {
+        self.0.insert(source_id, stage);
+    }
+
+    /// Update the [SourceStageInfo] for a particular module.
+    pub fn update(&mut self, source: SourceId, info: fn(SourceStageInfo) -> SourceStageInfo) {
```
kontheocharis, line 114: Would be nice to take a closure rather than fn pointer

compiler/hash-pipeline/src/workspace.rs
```diff
@@ -18,6 +19,121 @@ use hash_ast::{
 use hash_source::{InteractiveId, ModuleId, ModuleKind, SourceId, SourceMap};
 use hash_utils::tree_writing::TreeWriter;
 
+bitflags! {
+    /// Defines the flags that can be used to control the compiler pipeline.
+    pub struct SourceStageInfo: u32 {
+        /// Indicates the source has not undergone any stages after parsing.
+        const NONE = 0;
```
kontheocharis, line 26: `bitflags` docs says to avoid defining zero flags

kontheocharis: https://docs.rs/bitflags/latest/bitflags/#zero-flags

kontheocharis:
LGTM after comments

compiler/hash-ast/src/origin.rs
feds01: I think this can be removed

compiler/hash-pipeline/src/workspace.rs
feds01: Yeah, I will remove it and add documentation with this to say that `is_empty()` means that the module has at-least been parsed.

compiler/hash-ast/src/origin.rs
feds01: This method is being used in other contexts that don't demand `fmt::Display` to be implemented.

feds01:
> LGTM after comments. I mostly skipped over the IR details, I guess that is still WIP anyway

Yes, work will begin on the IR as soon as `hash-types` is re-structured, and scopes become visible on AST nodes.

compiler/hash-pipeline/src/workspace.rs
feds01: So the type would be `impl FnOnce(...) -> ...` ?

kontheocharis: yeah

kontheocharis (approved):

PR #547, opened 2022-10-26
`hash-types`: re-organise `lib.rs` into more appropriate files

feds01:
This patch brings the following changes:
- Split up `lib.rs` defined data structures e.g. `Term`, and `Scope` into files `term.rs` and `scope.rs` respectively. 
- Re-organise `fmt.rs` implementation for printing types into more appropriate files e.g.
the implementation, for printing for scopes, was moved from `fmt.rs` into `scope.rs`, whilst `fmt.rs` retained core type printing implementation details. 
- The `is_atomic` flag is removed from the type writer implementation and printing parentheses logic is inferred from the local context (but this **needs testing**).
- Fix the long-standing bug with interactive mode duplicating errors in each interactive session.


kontheocharis (approved):
Very nice!

Issue #545, opened 2022-10-26
AST Visitor: Add macros to default impl void methods


kontheocharis:
Something like

```
ast_default_impl_void!(A, B, C)
ast_default_impl_void_hiding!(A, B, C)
```

PR #549, opened 2022-10-26
Add ability to automatically generate default visitor methods that just "walk" the AST and return void

kontheocharis:
This adds the macros:
```
ast_visitor*_default_impl!(Expr, Name);
```

Providing the arguments `Expr`, `Main`, the macro will generate default implementations for those AST nodes. In subsequent PRs this can be used to replace all the manually written methods with this macro.

It would also be nice to have a way to "hide" AST nodes, but macros make that a little bit more complicated. It is left as a @@Todo.


compiler/hash-tree-def/src/emit.rs
```diff
@@ -760,19 +757,113 @@ fn emit_walker(
...
+    // let all_nodes = tree_def
+    //     .nodes
+    //     .keys()
+    //     .map(|node_name| {
+    //         let node_name = format_ident!("{}",
+    // node_name.to_string().to_case(Case::Pascal));         quote! {
+    // #node_name, }     })
+    //     .collect::<TokenStream>();
+
+    // For each node name given, emit a default impl by calling the appropriate `@@`
```
feds01, line 824: Is there a better name than `@@`, this seems a little too cryptic.

compiler/hash-tree-def/src/emit.rs
```diff
@@ -760,19 +757,113 @@ fn emit_walker(
...
-        emit_visitor(tree_def, true, true),
-        emit_visitor(tree_def, false, true),
-        emit_visitor(tree_def, false, false),
-        emit_walker(tree_def, true, true)?,
-        emit_walker(tree_def, false, true)?,
-        emit_walker(tree_def, false, false)?,
-    ]
-    .into_iter()
-    .collect())
+    let mutability_combinations = [(true, true), (false, true), (false, false)];
```
feds01, line 850: interesting that there is no case for `(true, false)`

feds01:
Very nice, should shave off another couple of 100loc within the visitor impls!

compiler/hash-tree-def/src/emit.rs
kontheocharis: It wasn't used by anything, I guess I can add it but `(true, false)` is almost the same as `(true, true)` because most of the time when mutating ast nodes you have to mutate self as well in some way.

feds01: yeah, no changes are required, just thought it was interesting that it wasn't a use case.

feds01 (approved):

kontheocharis:
Closes #545.

PR #559, opened 2022-11-01
Add the ability to assign `ScopeId`s to AST nodes

### LINKED ISSUE #558: TC: Add scope IDs to AST nodes

kontheocharis:
This PR adds the ability for AST nodes to have attached scopes, in the same way that they currently can have attached `TermId`s or `PatId`s. This is because scope information is required during code generation. The way to access `ScopeId`s is through the `NodeInfoStore` struct on `GlobalStorage`, by providing an `AstNodeId` as an input and retrieving a scope using the `scope_id` method on the resultant `NodeInfoTarget`.

The following nodes have scopes attached to them:
- `MatchCase`
- `BodyBlock`
- `TraitDef`
- `TraitImpl`
- `ModBlock`
- `ImplBlock`
- `Modules`

Therefore, it is only valid to call `scope_id` on `NodeInfoTarget`s accessed through an `AstNodeId` corresponding to one of the above AST nodes. (Otherwise there will be a panic.)

Closes #558

### DIFF
--- a/compiler/hash-typecheck/src/traverse/scopes.rs
+++ b/compiler/hash-typecheck/src/traverse/scopes.rs
@@ -20,9 +20,10 @@ impl<'tc> TcVisitor<'tc> {
     /// The `scope_to_use` parameter is optional and is used to override the
     /// scope that is used to append all the members into. If not given, a new
     /// constant scope is created and used.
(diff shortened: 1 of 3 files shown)


feds01 (approved):
LGTM

PR #561, opened 2022-11-01
Add patterns and literals to the new term structure.

### LINKED ISSUE #560: TC: Add new definitions for patterns

kontheocharis:
This PR completes the `pats` module of the new term structure. The pattern structure is simplified compared to the old structure. The notable differences are:

- The `Mod` pattern kind is not used anymore. Instead, module destructuring statements at constant scope level will be resolved to actual members during an AST traversal pass, so they do not need to be present in the term structure. This also removes the need to have `visibility` on `BindingPat`.
- The `Spread` pattern kind is not used anymore. Instead, list, tuple and constructor patterns have an optional `spread` field which records an index position and an optional name. It denotes that a spread pattern is present at the given position, with the given name. This is possible due to the invariant that spread patterns can only appear at most once in a sequence of patterns. (It is possible that in the future this restriction can be relaxed for tuples and constructors but for now it is kept to make things simple.)
- The `Lit` pattern does not store any `TermId` anymore, but rather stores the exact literal being matched. To complement this, a new module called `lits` is added, which contains pattern and term literal types. For now these use the definitions from `hash_ast` because they would just be duplicated otherwise.
- The `Range` pattern again uses literals for its interval endpoints.

Closes #560.

### DIFF
--- a/compiler/hash-types/src/new/control.rs
+++ b/compiler/hash-types/src/new/control.rs
@@ -55,14 +55,28 @@ pub enum LoopControlTerm {
 }
 
 /// A conditional pattern, containing a pattern and an condition.
+///
+/// This is `A if c` in its most general form, where `A` is a pattern and `c` is
+/// a boolean-valued term.
+///
+/// This pattern matches iff `A` matches and `c` is met.
 #[derive(Clone, Debug, Copy)]
 pub struct IfPat {
+    /// The subject pattern `A`.
     pub pat: PatId,
+    /// The condition `c`.
     pub condition: TermId,
 }
 
-/// Represents a list of alternative patterns (A | B | C).
+/// A list of alternative patterns.
+///
+/// This is `A | B | C | ...` in its most general form, where `A`, `B`, `C`,
+/// ..., are patterns.
+///
+/// This pattern matches iff at least one of the patterns in the sequence
+/// matches.
 #[derive(Copy, Clone, Debug)]
 pub struct OrPat {
+    /// The sequence of alternative patterns.
     pub alternatives: PatListId,
 }
--- a/compiler/hash-types/src/new/control.rs
+++ b/compiler/hash-types/src/new/control.rs
@@ -1,6 +1,6 @@
 //! Definitions related to control flow.
 
-use super::pats::PatListId;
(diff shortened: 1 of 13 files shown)


compiler/hash-types/src/new/control.rs
```diff
@@ -53,3 +53,30 @@ pub enum LoopControlTerm {
...
+/// This is `A if c` in its most general form, where `A` is a pattern and `c` is
+/// a boolean-valued term.
+///
+/// This pattern matches iff `A` matches and `c` is met.
+#[derive(Clone, Debug, Copy)]
+pub struct IfPat {
+    /// The subject pattern `A`.
+    pub pat: PatId,
+    /// The condition `c`.
+    pub condition: TermId,
```
feds01, line 68: If the condition is always a boolean type, is there any need to store the term here?

compiler/hash-types/src/new/pats.rs
```diff
@@ -2,12 +2,73 @@
...
+pub struct RangePat {
+    /// The beginning of the range.
+    pub start: LitPat,
+    /// The end of the range.
+    pub end: LitPat,
+    /// If the range includes the `end` or not.
+    ///
+    /// If `true`, this is the pattern `start..<end`, otherwise it is
+    /// `start..end`.
+    pub end_exclusive: bool,
```
feds01, line 52: I don't think `RangeEnd` should be replaced here with a `bool`. In this instance, I think verbosity is preferred because use cases of this fact become more self-documenting, especially in the exhaustiveness checking sources.

feds01 (changes requested):

compiler/hash-types/src/new/control.rs
kontheocharis: The condition is of a boolean type, but it can still be any arbitrary term that is typed as a boolean

compiler/hash-types/src/new/pats.rs
kontheocharis: That's fine, I don't think it is really that much of a difference. However I would want RangeEnd to not live in AST, but rather in source, same as you have done in your PR for int/float/etc literals.

PR #562, opened 2022-11-01
IR: introduce `IrTy` to represent types at the IR level

feds01:
This patch introduces `IrTy` which is an enum that represents all of the types that can occur at the IR level. This enum is a simplified version of the `Term` structure that is used in the typechecking implementation.  This is introduced in order to avoid the complexities of dealing with the entire term structure during IR generation since only a part of the type term structure is present at the time of generation.

### DIFF
--- a/compiler/hash-ir/src/ty.rs
+++ b/compiler/hash-ir/src/ty.rs
@@ -1,0 +1,249 @@
+//! Simplified Hash type hierarchy. The [IrTy] is used a simplified
+//! variant of the available Hash [Term]s that are defined in the
+//! `hash-types` crate. This data structure is used to remove unnecessary
+//! complexity from types that are required for IR generation and
+//! analysis.
+
+use std::fmt;
+
+use hash_source::{
+    constant::{FloatTy, SIntTy, UIntTy},
+    identifier::Identifier,
+};
+use hash_utils::{
+    new_sequence_store_key, new_store_key,
+    store::{CloneStore, DefaultSequenceStore, DefaultStore, SequenceStore, Store},
+};
+use index_vec::IndexVec;
+
+use crate::write::{ForFormatting, WriteTyIr};
+
+/// Mutability of a particular variable, reference, etc.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Mutability {
+    /// Mutable variable, reference, etc.
+    Mutable,
+    /// Immutable variable, reference, etc.
+    Immutable,
+}
+
+impl Mutability {
+    /// Returns true if the mutability is mutable.
+    pub fn is_mutable(&self) -> bool {
+        matches!(self, Mutability::Mutable)
+    }
+
+    /// Returns true if the mutability is immutable.
+    pub fn is_immutable(&self) -> bool {
+        matches!(self, Mutability::Immutable)
+    }
+
+    /// Get [Mutability] as a printable name
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Mutability::Mutable => "mut",
+            Mutability::Immutable => "",
+        }
+    }
+}
+
+impl fmt::Display for Mutability {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.as_str())
+    }
+}
+
+/// Simplified type structure used by the IR and other stages to reason about
+/// Hash programs once types have been erased and simplified.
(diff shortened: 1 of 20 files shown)


compiler/hash-ir/src/ty.rs
```diff
@@ -0,0 +1,202 @@
...
+    Bool,
+
+    /// Character type.
+    Char,
+
+    /// The never type
+    Never,
+
+    /// A reference type, referring to a another type, e.g. `&T`
+    Ref(IrTyId),
```
kontheocharis, line 48: What about different reference variants?

compiler/hash-ir/src/ty.rs
```diff
@@ -0,0 +1,202 @@
...
+    Int(SIntTy),
+
+    /// A primitive unsigned integer type, e.g. `u32`.
+    UInt(UIntTy),
+
+    /// A primitive floating point type, e.g. `f32`.
+    Float(FloatTy),
+
+    /// String type.
+    Str,
```
kontheocharis, line 36: When will strings be converted to a lower representation?

compiler/hash-ir/src/ty.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// Defines the flags that denote which kind of ADT is being represented with
+    /// [AdtData].
+    ///
+    /// @@Todo: do we need add specific flags for `data` variant??
+    pub struct AdtFlags: u32 {
+        /// The ADT is a enum.
+        const ENUM = 0b00000001;
+
+        /// The ADT is a struct.
+        const STRUCT = 0b00000010;
```
kontheocharis, line 99: These are mutually exclusive, does it really make sense to use bitflags for it?

compiler/hash-ir/src/ty.rs
```diff
@@ -0,0 +1,202 @@
...
+///     - add layout randomisation configuration
+///     - add `C` layout configuration
+#[derive(Clone)]
+pub struct AdtRepresentation {}
+
+#[derive(Clone)]
+pub struct AdtVariant {
+    /// The name of the variant, if this is a struct variant, this inherits
+    /// the name of the struct.
+    pub name: Identifier,
```
kontheocharis, line 118: Why can variant names not be index/numbers? Same for field names

compiler/hash-lower/src/lib.rs
```diff
@@ -67,37 +68,36 @@ impl<Ctx: IrLoweringCtx> CompilerStage<Ctx> for AstLowerer {
...
+        let bodies_to_dump = lowered_bodies
+            .iter()
+            .enumerate()
+            .filter(|(index, body)| body.needs_dumping())
+            .map(|(index, _)| index)
+            .collect::<Vec<_>>();
+
+        for (index, body_index) in bodies_to_dump.into_iter().enumerate() {
+            if index > 0 {
+                println!();
```
kontheocharis, line 94: ?

compiler/hash-ir/src/ty.rs
kontheocharis: Also this information is not retained by the typechecker at all anyway, only perhaps for error message purposes, so it shouldn't be relevant at this stage---both structs and enums desugar to `datatype` definitions.

compiler/hash-ir/src/ty.rs
feds01: The other variants are `RawRef` and ...?

compiler/hash-ir/src/ty.rs
feds01: Probably at a later stage when things get emitted into actual machine code.

compiler/hash-lower/src/lib.rs
feds01: This is to print a newline between all of the dumped bodies that it has to dump.

compiler/hash-ir/src/ty.rs
feds01: This is mainly for debugging purposes, when the IR is printed it will use the variant name

compiler/hash-ir/src/ty.rs
kontheocharis: Mutable ones, local refs vs RC refs

feds01: For local ones, do you mean ones that point to the stack? If yes, is there really a need to differentiate between `Ref` and `StackRef`?

kontheocharis (approved):

Issue #563, opened 2022-11-02
TC: Add context definitions


kontheocharis:

kontheocharis:
Closed by #565

PR #565, opened 2022-11-02
Add definitions for context in new term structure

### LINKED ISSUE #564: TC: Add coercion definitions

kontheocharis:
This PR adds definitions for the current "typechecking" context for the new term structure. This essentially consists of a stack of scopes, each of which is a stack of local bindings. Each binding might refer to a different thing: a mod member, a trait member, a stack member, or a bound varable.

It also adds a coercion term for explicit type coercion, as well as self reference IDs to various definitions like arguments and parameters.

Some more things have been simplified in the term structure, removing as many unneeded "unsimplified" variants as possible.

Closes #564, #563


compiler/hash-types/src/new/access.rs
```diff
@@ -1,17 +1,30 @@
...
+    /// Accessing a constructor field, like `f := Foo(name=3); f.name`.
+    CtorField,
+    /// Accessing a tuple field, like `f := (2, 3); f.0`.
+    TupleField,
+    /// Accessing a module member, like `X := mod { y := 3 }; X.y`.
+    ModMember,
+    /// Accessing a trait member, like `T := trait { y := 3; z := self.y }`
+    TrtMember,
+    /// Accessing a datatype constructor, like `Colour := enum(Red, Green,
+    /// Blue); Colour.Red`
```
feds01, line 17: The access syntax is `Colour::Red`

kontheocharis: What do you think about keeping `.` for everything? looks cleaner imo

feds01: Maybe change this at a later stage when the new type-checking implementation is used, right now it's hard to disambiguate between a namespace access and a property access, we've tried it before and it wasn't so easy to make it work with the current implementation.

kontheocharis: Yeah right now it's not possible. But yeah it should be with the new TC because subjects can always be disambiguated fully (no merges). So for now we will keep `.` and `::`

feds01 (approved):

PR #568, opened 2022-11-03
lower: implement logic for converting `Term`s into `IrTy`s for the IR builder

feds01:


compiler/hash-lower/src/build/ty.rs
```diff
@@ -0,0 +1,212 @@
...
+                let params = tcx
+                    .params_store
+                    .get_owned_param_list(members)
+                    .into_positional()
+                    .into_iter()
+                    .map(|param| convert_term_into_ir_ty(param.ty, tcx, ir_ctx));
+
+                let params_id = ir_ctx.ty_list_store().create_from_iter(params);
+                IrTy::Tuple(params_id)
+            }
```
kontheocharis, line 124: Maybe directly replace with adt here?

kontheocharis (approved):

compiler/hash-lower/src/build/ty.rs
feds01: Yes, ill add it as another variant of the ADT

kontheocharis (approved):

PR #585, opened 2022-11-04
IR: lower primitive literal constants and declarations

feds01:
This patch implements IR lowering for constant literals such as integers, floats, etc. Additionally, this patch also implements the handling of declarations that appear within body blocks.

### DIFF
--- a/compiler/hash-lower/src/build/expr.rs
+++ b/compiler/hash-lower/src/build/expr.rs
@@ -1,5 +1,6 @@
 use hash_ast::ast::{AstNodeRef, BlockExpr, Expr, UnsafeExpr};
 use hash_ir::ir::{BasicBlock, Place, RValue};
+use hash_reporting::macros::panic_on_span;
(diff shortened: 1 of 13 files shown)


compiler/hash-lower/src/build/expr.rs
```diff
@@ -86,14 +93,39 @@ impl<'tcx> Builder<'tcx> {
...
+    /// body, and thus we should not lower it.
     ///
+    /// @@Todo: deal with non-trivial dead-ends, i.e. compound patterns that
+    /// have dead ends...
     pub(crate) fn handle_expr_declaration(
         &mut self,
-        place: Place,
         block: BasicBlock,
-        body: AstNodeRef<'tcx, Expr>,
+        declaration: AstNodeRef<'tcx, Expr>,
```
kontheocharis, line 106: Why not directly take an `Expr::Declaration`?

kontheocharis (approved):

Issue #574, opened 2022-11-04
lower unary expressions


feds01:
Currently blocked by the typechecking implementation not properly typing unary expressions.

PR #590, opened 2022-11-06
typechecker: introduce workaround for currently broken trait resolution

feds01:
Currently, the typechecker cannot properly check the types of binary and unary expressions because the trait resolution system is not working properly and that the `prelude` does not contain trait definitions for these types. Once the new typechecking system is introduced by @kontheocharis, this issue will be fixed and this hack will be removed.

Relevant issue: https://github.com/hash-org/lang/issues/419

This patch is needed to implement: #574, #582

### DIFF
--- a/compiler/hash-typecheck/src/ops/oracle.rs
+++ b/compiler/hash-typecheck/src/ops/oracle.rs
@@ -2,7 +2,7 @@
 //! constructs.
 use hash_ast::ast::ParamOrigin;
 use hash_source::{
-    constant::{IntTy, SIntTy, UIntTy},
+    constant::{FloatTy, IntTy, SIntTy, UIntTy},
     identifier::Identifier,
 };
 use hash_types::{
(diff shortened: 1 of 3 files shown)


compiler/hash-typecheck/src/traverse/visitor.rs
```diff
@@ -1662,6 +1662,35 @@ impl<'tc> visitor::AstVisitor for TcVisitor<'tc> {
...
+        // @@Hack: currently, trait resolution logic is broken, so we introduce
+        // a workaround for primitives to support all operators so that we can
+        // continue onto lowering...
+        //
+        // @@Note: this does not handle floats.
+        if self.oracle().term_is_primitive(lhs) && self.oracle().term_is_primitive(rhs) {
+            let lhs_term = if self.oracle().term_is_literal(lhs) {
+                self.typer().infer_ty_of_term(lhs)?
+            } else {
+                lhs
```
kontheocharis, line 1674: you can't just return `lhs` otherwise, because in one branch you're returning a type and in another you're returning a value

compiler/hash-typecheck/src/ops/oracle.rs
```diff
@@ -56,6 +56,20 @@ impl<'tc> Oracle<'tc> {
         }
     }
 
+    /// Check if the [Term] is a primitive type.
+    pub fn term_is_primitive(&self, term: TermId) -> bool {
```
kontheocharis, line 60: You're mixing literal values and primitive types here

kontheocharis (changes requested):

compiler/hash-typecheck/src/ops/oracle.rs
feds01: Yes, this is to find out if it is of a **primitive** type, ideally, this would handle values as well, and just try to get the type of it.

compiler/hash-typecheck/src/traverse/visitor.rs
feds01: Yes, I see. I guess it should check if it is a `value` overall and then try and get the type of it. Is this safe to do in all cases where the term is `Level0`?

kontheocharis (approved):

PR #591, opened 2022-11-10
TC: Add a scope discovery AST pass to find modules/traits (Partial)

kontheocharis:
This PR adds the first AST pass (partially), which discovers top level definitions such as modules and traits. This is partially completed, as this stage should also discover local scopes, as well as impl blocks. A lot of documentation is not present yet (see `@@Docs` comments), and will be progressively added as more of the traversal is implemented. This is a checkpoint PR to keep the change sizes somewhat small.

A `USE_NEW_TC` environment flag is checked, and the typechecker implementation is decided based on its value. This is an ugly hack that is needed while both typecheckers are in use, and will be eventually removed.


kontheocharis:
Closing this in favour of `ast-passes-3` branch.

PR #592, opened 2022-11-10
pipeline: introduce `Target` to denote what compilation target the compiler is compiling to

feds01:
This patch adds a new crate `hash-target` that contains all data structures that are needed for denoting which compilation target the compiler is emitting code for. There are two main data structures introduced:

- `Target` specifies the `point_width` in bytes and the name of the target (if applicable). It is likely that many more configurations/settings will be added to this struct.
- `TargetInfo` which specifies the `target` to which we are compiling, and `host` which is the host machine of the compiler represented as a `Target`. The `host` variant will be needed later for compile-time execution/VM operations.

Additionally, this patch integrates `pointer_width` based on the compilation target when computing the size, minimum and maximum values of `isize` and `usize` in `hash-source` and other places.

### DIFF
--- a/compiler/hash-source/src/constant.rs
+++ b/compiler/hash-source/src/constant.rs
@@ -113,14 +113,13 @@ pub enum UIntTy {
 impl UIntTy {
     /// Get the size of [IntTy] in bytes. Returns [None] for
     /// [UIntTy::UBig] variants
-    pub const fn size(&self) -> Option<u64> {
+    pub const fn size(&self, ptr_width: usize) -> Option<u64> {
         match self {
             UIntTy::U8 => Some(1),
             UIntTy::U16 => Some(2),
             UIntTy::U32 => Some(4),
             UIntTy::U64 => Some(8),
-            // @@Todo: actually get the target pointer size, don't default to 64bit pointers.
-            UIntTy::USize => Some(8),
+            UIntTy::USize => Some(ptr_width as u64),
             UIntTy::U128 => Some(16),
             UIntTy::UBig => None,
         }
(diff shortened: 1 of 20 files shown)


compiler/hash-target/src/lib.rs
```diff
@@ -0,0 +1,64 @@
+//! Definitions to describe the target of Hash compilation.
+
+/// The target that the compiler should compile for.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Target {
+    /// The size of the pointer for the target in bytes.
+    pub pointer_width: usize,
```
kontheocharis, line 7: invariant that <= 8

compiler/hash-source/src/constant.rs
```diff
@@ -198,29 +199,36 @@ impl SIntTy {
...
+    pub fn min(&self, ptr_width: usize) -> Option<BigInt> {
         match self {
             SIntTy::I8 => Some(BigInt::from(i8::MIN)),
             SIntTy::I16 => Some(BigInt::from(i16::MIN)),
             SIntTy::I32 => Some(BigInt::from(i32::MIN)),
             SIntTy::I64 => Some(BigInt::from(i64::MIN)),
             SIntTy::I128 => Some(BigInt::from(i128::MIN)),
-            SIntTy::ISize => Some(BigInt::from(isize::MIN)),
+            SIntTy::ISize => {
+                let min = (i64::MAX) << ((ptr_width * 8) - 1);
```
kontheocharis, line 229: some quick tests would be nice

compiler/hash-target/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+impl Target {
+    /// Create a new target from the given name and pointer width.
+    pub fn new(name: String, pointer_width: usize) -> Self {
+        Self { name, pointer_width }
+    }
+
+    /// Create a new target from the given name and pointer width.
+    pub fn from_string(name: String) -> Option<Self> {
+        match name.as_str() {
+            "x86_64" => Some(Self::new(name, 8)),
```
kontheocharis, line 22: Maybe these should be an enum?

compiler/hash-target/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+            "aarch64" => Some(Self::new(name, 8)),
+            "arm" => Some(Self::new(name, 4)),
+            _ => None,
+        }
+    }
+}
+
+impl Default for Target {
+    fn default() -> Self {
+        Self { pointer_width: 8, name: "x86_64".into() }
```
kontheocharis, line 33: This should either not be here or deducted from host

compiler/hash-typecheck/src/exhaustiveness/constant.rs
```diff
@@ -29,8 +29,8 @@ impl Constant {
     /// Function to convert a [BigInt] into a [Constant]. The only
     /// constraint is that it can fit into a [u128], otherwise the
     /// function will currently panic.
-    pub fn from_int(int: BigInt, kind: IntTy, ty: TermId) -> Self {
-        let size = kind.size().unwrap_or_else(|| int.bits());
+    pub fn from_int(int: BigInt, kind: IntTy, ty: TermId, ptr_width: usize) -> Self {
+        let size = kind.size(ptr_width).unwrap_or_else(|| int.bits());
```
kontheocharis, line 33: int.bytes()?

feds01: Although, I think it is easier to multiply the returned integer size by 8 to convert it into a bit count instead of using `bigint` to get the number of bytes since it seems they don't have an API for providing the size of the number in bytes.

kontheocharis (approved):

PR #611, opened 2022-12-08
IR: lower `match` expressions with full range of patterns

### LINKED ISSUE #577: lower match cases

feds01:
This PR implements the process of lowering match cases. Notably, the algorithm is quite complicated due to it avoiding some performance and space traps. There is a notion of a `Candidate` that represents a `match` arm which contains information about the arm, whether it has sub-candidates (which represent `or`-pattern variants, compound patterns), `BasicBlock` points of where a program should jump to if the pattern is matched, whether it should jump in the `otherwise` case, etc. The main meat of the algorithm is:

- that when `Candidate`s are all generated, a candidate is selected, and a test for that candidate is generated.  A `Test` is a representation of what kind of test needs to be performed in order to check whether a value matches the specified pattern.
- Next, all of the un-selected candidates are sorted by the most likely branches that would appear in the `otherwise` condition. For example, when a test is performed to check if `x == 3`, we want branches that want values that are `x != 3` to succeed in this branch. This is because we want to reduce the number of blocks we generate for the various value checks (to keep the amount of code generated as linear as possible relative to the number of patterns specified). If the test fails, it is likely that the matching arm can make progress to fully matching (as we don't want to re-check values more than once).
- Once all of the candidates have been processed using this algorithm, all of their blocks are sewn up between one another in order to preserve the jumping order between one branch and the previous.
- Next, we have to deal with `if`-guards on patterns that will only run if a particular pattern is successfully matched, and it has to deal with the situation where the `if`-guard does not match, and it resumes jumping to the next successor `Candidate`.
- When this is all complete, all of the `match` arm bodies are lowered and again joined up with their corresponding candidates (which there may be multiple because of `or`-patterns).
- All arm bodies are terminated with the successor block of the `match` expression.

IR changes:
- `Const::Byte` is now explicitly `Const::Bool` and is only really used as an auxiliary constant kind for pretty printing the IR.
- Introduce the right value `RValue::Len` variant that is used to compute the length of a slice or list directly.
- Introduce the `PlaceProjection::ConstantIndex` projection in order to express indexing projections on constant slices, that appear within match cases, and patterns in general.
- Introduce the `PlaceProjection::Subslice` projection that allows for capturing a subslice of a given slice with specified offsets.
- Fix some minor issues with pretty-printing IR.
- Add some functionality for evaluating if two constants (specifically integral constants) have the same value.

Additionally:
- tests: fix tests not running on windows due to line endings.
- core: when `optimisation_level` is specified as `release`, the lower should not generate checked operations.
- tc: set the `TermId` on `BlockExpr` from traversing the inner block.

closes #577


kontheocharis (approved):
You left a file.png

kontheocharis (approved):

kontheocharis (approved):

kontheocharis (approved):

PR #614, opened 2022-12-09
Reporting: Better fluent API for creating reports

kontheocharis:
This PR replaces `ReportBuilder` in favour of `Reporter`. `Reporter` allows the creation of multiple reports, and provides methods to create errors, warnings, etc in an easy way. Also, the `Report` struct now implements `Default` and has more convenience methods to add notes, spans, etc.


compiler/hash-reporting/src/report.rs
```diff
@@ -149,16 +150,95 @@ impl Report {
...
+        )))
+    }
+
+    /// Add a code block at the given location to the [Report].
+    pub fn add_span(&mut self, location: SourceLocation) -> &mut Self {
+        self.add_element(ReportElement::CodeBlock(ReportCodeBlock::new(location, "")))
+    }
+
+    /// Add a named code block at the given location to the [Report].
+    pub fn add_named_span(
```
feds01, line 205: Probably better to call it `labelled_span` rather than `named_span`

compiler/hash-reporting/src/report.rs
```diff
@@ -149,16 +150,95 @@ impl Report {
     pub fn is_warning(&self) -> bool {
         self.kind == ReportKind::Warning
     }
+
+    /// Add a general message to the [Report].
+    pub fn message(&mut self, message: impl ToString) -> &mut Self {
+        self.message = message.to_string();
+        self
+    }
```
feds01, line 158: maybe name this `title()` since it is exclusive to all other messages that a report can contain

feds01 (changes requested):
Really nice, finally refactoring the old API away!

feds01 (approved):

PR #616, opened 2022-12-15
tc+exhaustiveness: add partial support for exhaustiveness checking for `enum`s

feds01:
This patch implements partial support for `enum`s within exhaustiveness checking. In particular, the lowering process of `enum` variant patterns has been implemented to convert them into `DeconstructedCtor::Variant`, and then let the core of the algorithm check exhaustiveness. In order to make this change work, there was an adjustment from converting `EnumDef`s into `Unit` definitions, and rather into the old `Enum` format since most of the code was already purposed for this representation. When the new type-checking is stable and implemented, this will need to be migrated (albeit it should be easier to migrate to the new representation since the initial work has been done).

Additionally:
- this fixes a bug with bootstrapping process of the compiler on Windows where the `prelude.hash` module path could not be resolved due to differing path separators.
- some style fixes in the old type checking visitor.

### DIFF
--- a/compiler/hash-typecheck/src/exhaustiveness/fields.rs
+++ b/compiler/hash-typecheck/src/exhaustiveness/fields.rs
@@ -92,9 +92,7 @@ impl<'tc> FieldOps<'tc> {
         let ctor = reader.get_deconstructed_ctor(ctor);
 
         match ctor {
-            DeconstructedCtor::Single | DeconstructedCtor::Variant(_) => {
-                let reader = self.reader();
-
+            ctor @ (DeconstructedCtor::Single | DeconstructedCtor::Variant(_)) => {
                 match reader.get_term(ctx.ty) {
                     Term::Level1(Level1Term::Tuple(TupleTy { members })) => {
                         let members = reader.get_params_owned(members);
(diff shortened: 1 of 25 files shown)


compiler/hash-typecheck/src/exhaustiveness/fields.rs
```diff
@@ -103,31 +101,41 @@ impl<'tc> FieldOps<'tc> {
...
-                                    self.for_fmt(ctx.ty),
-                                ),
-                            },
-                            NominalDef::Unit(_) => {
-                                Fields::empty()
-                            },
-                            // @@EnumToUnion: when enums aren't represented as this anymore
-                            NominalDef::Enum(_) => unreachable!(),
-                        }
+                        // get the varaint index from the deconstructed ctor
```
kontheocharis, line 104: typo

compiler/hash-typecheck/src/exhaustiveness/lower.rs
```diff
@@ -177,52 +182,63 @@ impl<'tc> LowerPatOps<'tc> {
...
+                                            panic!("got unexpected opaque struct-def here")
+                                        }
+                                    },
+                                    NominalDef::Unit(_) => (DeconstructedCtor::Single, vec![]),
+                                    NominalDef::Enum(_) => {
+                                        println!("{:?}", self.for_fmt(subject));
+                                        // We need to deduce the variant index from the type of the
+                                        // pattern?
+                                        todo!()
+                                    }
```
kontheocharis, line 229: Add a @@Todo?

kontheocharis (changes requested):

kontheocharis (approved):

PR #623, opened 2022-12-17
lower: support lowering constant blocks

### LINKED ISSUE #620: lower constant items

feds01:
This patch adds support to the lowerer for constant blocks and items. This patch also allows for constant items (i.e. items that are referenced from scopes above the function) to be referenced within lowered function bodies. These items are now represented as `ConstKind::Unevaluated`, and they are resolved at a later stage during constant propagation and graph simplification which will occur before transitioning to lowering into a specific backend.

closes #620


compiler/hash-ir/src/ir.rs
```diff
@@ -111,6 +111,35 @@ pub fn compare_constant_values(left: Const, right: Const) -> Option<Ordering> {
     }
 }
 
+/// A constant value that is used within the IR. A [ConstantValue] is either
+/// an actual [Const] value, or an un-evaluated reference to a constant
+/// expression that comes outside of a particular function body. These
+/// "unevaluated" values will be removed during IR simplification stages since
+/// all of the items are inlined.
```
kontheocharis, line 118: What is an example of an unevaluated constant?

feds01: For example, in this code snippet:
```rust
l := 1;
k := 2 + l;

#dump_ir main := (y: bool, x: i32) => {
    if x == 2 || x == 3 {
        k
    } else {
        2
    }
}
```

within the constant body `k`, there is a reference to `l` which at the time of lowering is considered as un-evaluated since it will later be resolved, and likely inlined. Similarly,  in `main`, there is a reference to `k` which at the time is also un-evaluated.

After some point of the lowering process, it will be illegal for there to be any **un-evaluated** constants.

kontheocharis (approved):

PR #625, opened 2022-12-19
Traverse functions and blocks in scope discovery

### LINKED ISSUE #601: Implement scope discovery for functions

kontheocharis:
This PR adds code to traverse function definitions during scope discovery and add them to the storage. At this stage, all types and terms are left as holes, and will be resolved further down the line when all the declarations have been discovered.

Closes #601

### DIFF
--- a/compiler/hash-typecheck/src/new/passes/scope_discovery.rs
+++ b/compiler/hash-typecheck/src/new/passes/scope_discovery.rs
@@ -6,7 +6,10 @@
 /// - Function definitions in modules
 /// - Data definitions in modules, including struct, enum
 /// - Stack scopes, and their member names and mutabilities
-use std::cell::Cell;
+use std::{
+    cell::Cell,
+    iter::{empty, once},
+};
 
 use derive_more::From;
 use hash_ast::{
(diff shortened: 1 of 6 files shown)


compiler/hash-typecheck/src/new/passes/scope_discovery.rs
```diff
@@ -565,4 +619,79 @@ impl<'tc> ast::AstVisitor for ScopeDiscoveryPass<'tc> {
...
+            name: fn_def_name,
+            body: FnBody::Defined(self.new_term_hole()),
+            ty: FnTy {
+                implicit: false,
+                is_unsafe: false,
+                params: self.create_hole_params(&node.params),
+                pure: false,
+                return_type: self.new_ty_hole(),
+            },
+        });
```
feds01, line 639: When does this deal closures?

kontheocharis: It already does, though what is still to be done is to have info on fndefs about what variables they capture

kontheocharis: This will happen in the symbol resolution stage

feds01: But how can you determine if a function definition is a function or a closure?

PR #628, opened 2022-12-19
Add unsupported error message for merge declarations

### LINKED ISSUE #627: Do not support merge declarations

kontheocharis:
Since the new typechecker does not currently support traits, it doesn't make sense for it to support merge declarations, so this PR adds an error message when a merge declaration is encountered.

#626 should be merged first.
Closes #627


feds01:
rebase?

feds01 (approved):

PR #635, opened 2022-12-20
Implement entering scopes during AST traversal in symbol resolution

### LINKED ISSUE #630: Implement context entry during symbol resolution

### LINKED ISSUE #634: Resolve symbols to their bindings and store them in the term store

kontheocharis:
This is the first in a line of PRs that will implement symbol resolution. So far, the symbol resolver walks all the scopes and adds them to the `Context` appropriately depending on their type. The next step is to visit `Variable`s and resolve them from the current context to their correct `Binding`s, and store those in the term store as `ResolvedVar`s.

Closes #630 .


compiler/hash-typecheck/src/new/diagnostics/error.rs
```diff
@@ -6,7 +6,7 @@ use hash_reporting::{
 use hash_source::location::SourceLocation;
 use hash_types::new::terms::TermId;
 
-use crate::new::{environment::tc_env::WithTcEnv, ops::common::CommonOps};
+use crate::new::{environment::tc_env::WithTcEnv, ops::common_ops::CommonOps};
```
feds01, line 9: overly verbose naming in my opinion `ops::<...>_ops`

kontheocharis: Hmm, fine, i will change it. I initially did this cause it was annoying to have files called `tuple.rs`

feds01 (approved):

PR #639, opened 2022-12-20
Fix: Perform scope discovery on match cases too

### LINKED ISSUE #637: Add bindings from matches to the context
Currently, only declaration patterns are handled in scope discovery. This should be extended to matches.

kontheocharis:
The scope discovery and symbol resolution stages did not handle match cases before. This PR implements this. Match cases are similar to declarations in that bindings inside the pattern are added to the child scope. Consequently, match cases create their own scopes.

Closes #637.


compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -48,6 +51,66 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+            }
+            ast::Pat::Spread(spread_pat) => {
+                if let Some(name) = spread_pat.name.as_ref() {
+                    if let Some(member_id) =
+                        self.ast_info().stack_members().get_data_by_node(name.ast_ref().id())
+                    {
+                        f(member_id);
+                    }
+                }
+            }
```
feds01, line 101: I suggest we wait until I put up my changes for the spread pattern stuff before merging this

kontheocharis: I don't think that will change anything?

feds01: Yes it will, spread patterns don't exist as an AST node anymore, they are present on tuples, lists and constructors only.

kontheocharis: Sure, i see. However, this will have to be changed in `add_stack_members_in_pat_to_buf` as well, so it might be easier to just merge this first and then modify once your PR is up? I don't mind either way but otherwise i'd have to branch off this PR for further work

feds01: I've already made changes in `add_stack_members_in_pat_buf` in my PR to deal with this case. It shouldn't be too long until this is in

kontheocharis: Okay, as long as you're about to put it up then that's fine

feds01:
rebase?

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -48,6 +51,75 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+    fn for_each_stack_member_of_pat(
+        &self,
+        node: ast::AstNodeRef<ast::Pat>,
+        f: impl Fn(StackMemberId) + Copy,
+    ) {
+        let for_spread_pat = |spread: &ast::AstNode<ast::SpreadPat>| {
+            if let Some(name) = &spread.name {
+                if let Some(member_id) =
+                    self.ast_info().stack_members().get_data_by_node(name.ast_ref().id())
+                {
```
feds01, line 70: could use a `let_chain` here:

```suggestion
            if let Some(name) = &spread.name && let Some(member_id) =
                    self.ast_info().stack_members().get_data_by_node(name.ast_ref().id())
                {
```

feds01 (approved):
LGTM, I don't mind if the suggestion is implemented or not.

PR #640, opened 2022-12-20
ast: remove `Pat::Spread`

feds01:
This patch removes the `Pat::Spread` variant from the pattern AST. This simplifies the AST since the patterns can only occur in lists, tuples, and constructors. This means that the special handling for these patterns (at least at the AST level) is contained to when parsing a tuple, constructor or list pattern.

With this change, the lowering process now supports the full range of patterns for declarations!

Ideally, this representation is carried onwards to the typechecking stage and further on in later work. There was an attempt to propagate this change into the **old** typechecking, however, there were some problematic cases encountered during what the lowering expected (since it was written as treating a spread pattern as just another pattern). There are ways to fix this problem, but will require some more thought.


compiler/hash-ast/src/origin.rs
```diff
@@ -44,3 +44,44 @@ impl Display for BlockOrigin {
...
+/// (hold multiple children patterns).
+#[derive(Clone, Copy, Debug)]
+pub enum PatOrigin {
+    /// The parent is a tuple, i.e `(x, 1, 2)`
+    Tuple,
+    /// The parent is a pseudo-pattern originating as a named field of some
+    /// other parent pattern. Named fields could occur within tuples or
+    /// constructor patterns, i.e `Bar(max = 123)`, where the `max` would be
+    /// the named field.
+    NamedField,
```
kontheocharis, line 60: What if the parent is a positional field? How is this different from `Tuple`/`Constructor`?

compiler/hash-ast/src/tree.rs
```diff
@@ -995,7 +1000,13 @@ impl AstVisitor for AstTreeGenerator {
         &self,
         node: ast::AstNodeRef<ast::TuplePat>,
     ) -> Result<Self::TuplePatRet, Self::Error> {
-        let walk::TuplePat { fields } = walk::walk_tuple_pat(self, node)?;
+        let walk::TuplePat { mut fields, spread } = walk::walk_tuple_pat(self, node)?;
+
+        // @@Note: this does not respect the order of the fields in the pattern
```
kontheocharis, line 1005: What does this mean?

compiler/hash-parser/src/parser/block.rs
```diff
@@ -120,13 +120,14 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
         let subject = self.parse_expr_with_precedence(0)?;
 
         let mut cases = AstNodes::empty();
-
         let mut gen = self.parse_delim_tree(Delimiter::Brace, None)?;
 
+        // @@Future: re-work syntax to avoid always specifying a semi-colon!!
```
kontheocharis, line 125: Does this result in a change in syntax?

compiler/hash-typecheck/src/new/passes/scope_discovery.rs
```diff
@@ -443,20 +459,32 @@ impl<'tc> ScopeDiscoveryPass<'tc> {
...
-            ast::Pat::List(list_pat) => {
-                for pat in list_pat.fields.ast_ref_iter() {
+            ast::Pat::List(ast::ListPat { fields, spread }) => {
+                for pat in fields.ast_ref_iter() {
                     self.add_stack_members_in_pat_to_buf(pat, buf);
                 }
+
+                if let Some(spread) = spread {
+                    register_spread_pat(spread, buf);
+                }
```
kontheocharis, line 487: Ideally the register_spread_pat call should happen at the iteration corresponding to the position of the spread in the pattern list.

tests/cases/semantics/pats/multiple_spread.stderr
```diff
@@ -2,12 +2,12 @@ error: spread patterns `...` can only be used once in a list pattern
  --> $DIR/multiple_spread.hash:4:15
 3 |   main := () => {
 4 |       [...x, t, ...] := a
-  |                 ^^^ here
+  |                 ^^^
```
kontheocharis, line 5: Maybe add a note?

compiler/hash-ast/src/tree.rs
feds01: it appends the item to the end of the fields, not necessarily where the spread pattern is actually specified.

compiler/hash-parser/src/parser/block.rs
feds01: Yes, it would, since it is expected that a `match` case is terminated with a `;`, but it is not necessarily needed.

tests/cases/semantics/pats/multiple_spread.stderr
feds01: what would the note specify?

compiler/hash-ast/src/origin.rs
kontheocharis: Could you add this as a comment?

tests/cases/semantics/pats/multiple_spread.stderr
kontheocharis: `cannot specify another spread pattern here` ?

compiler/hash-ast/src/tree.rs
kontheocharis: I see, though wouldn't it be better if it did?

kontheocharis: Something like `fields.insert(node.spread.unwrap().index, TreeNode::branch("spread", vec![spread]))`

kontheocharis (approved):

PR #641, opened 2022-12-22
IR: implement a basic visitor for mutable context, and a modifying visitor

### LINKED ISSUE #632: implement an IR Visitor for both mutable and immutable contexts
This visitor should be very similar to the AST visitor, but it should support potential modifications along the visiting process.

### LINKED ISSUE #633: implement IR optimisation that eliminate useless declarations

feds01:
This patch adds new IR visitors that are able to traverse the generated IR with the following parameters:
- a visitor with a mutable context, but immutable read the IR
- a visitor with an immutable context, but mutably walk the IR

The current visitor implementation is written by hand (and it is quite ugly, especially the `modify`s), and I solemnly promise to create a macro for the visitor and walking code for the following attributes:
- a visitor with a mutable context, immutably walk the IR
- a visitor with an immutable context, immutably walk the IR
- a visitor with a mutable context, mutably walk the IR
- a visitor with an immutable context, mutably walk the IR

Furthermore, it would be nice to find a way to avoid constantly `modify`ing the IR (since it is a `clone()`) but attempts to use `fast_modify()` have led to runtime panics with borrowing errors. 

As a proof of concept for the visitors, the `eliminate_dead_locals` pass has been implemented. This pass removes all dead `Local`s that are never referenced within an `RValue`. The pass does multiple sweeps until it reaches a fixed point, where all `Local`s are alive.

fixes #632 
fixes #633

### DIFF
--- a/compiler/hash-ir/src/ir.rs
+++ b/compiler/hash-ir/src/ir.rs
@@ -10,7 +10,7 @@ use hash_source::{
     location::{SourceLocation, Span},
     SourceId,
 };
-use hash_types::{scope::ScopeId, terms::TermId};
+use hash_types::scope::ScopeId;
 use hash_utils::{
     new_sequence_store_key, new_store_key,
     store::{DefaultSequenceStore, DefaultStore, SequenceStore, Store},
@@ -24,6 +24,8 @@ use crate::{
     IrStorage,
 };
 
+/// A specified constant value within the Hash IR. These values and their
+/// shape is always known at compile-time.
 #[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
 pub enum Const {
     /// Nothing, it has zero size, and is associated with a particular type.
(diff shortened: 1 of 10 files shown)


kontheocharis (approved):
I'm guessing parts of this are still incomplete given the commented out code?

feds01:
It's just somewhat unclear if it will be needed or not, since most things don't need to "return" something when visiting the IR.  
Currently, I only see two use cases:

1) A visitor is "looking" for some occurrence of something, and when they match a predicate, then it is recorded in the state of the visitor.

2) A visitor traverses the IR with the intent to modify some part of the IR, which may or may not have a mutable state. This visitor is painful because it essentially clones the whole `RValue` tree because of the multiple references problem. 

After some thought, it could be possible to "change" the approach of the case (2) to "flag" an `RValue` which will require some kind of modification, and then run the modification on the flagged `RValue`s. This way, it should mean only one `RValue` is modified at once. However, integrating this into a visitor pattern seems to me that it is too specific for a general visitor pattern involving ids. What do you think about this?

PR #644, opened 2022-12-23
ir+lowering improvements

feds01:
### Changes to IR
This patch makes the `RValue` tree simpler by introducing the notion of an `Operand`, which is essentially the leaf of an `RValue` tree. Introducing this item will do the following:
- reduce the number of `RValue`s created overall.
- simplify and improve the performance of traversals that may modify the IR as they go.

### Additionally:

- This patch also greatly improves the performance of the `simplify_graph` pass making it run in about `<1ms`, where as previously it was taking `~550ms` for a particular test case. This optimisation was achieved by keeping an in-place up-to-date record of graph block predecessors when eliminating `goto` chains.

- fix bug where `reached_terminator` wasn't being reset in `Builder` causing all future blocks to not be lowered.

### Some future work and questions:
- Is the whole `Operand::Const(ConstKind::Value(Const::<...>)))` naming sensible or should we rethink this?
- ~~I still think it might be worth giving a try to see if the `RValue` store can be completely eliminated so that we 
  just operate on pointers rather than being behind a store (this should be easier to do now that the tree is relatively flat).~~ (done)
- can we go back to the original `SwitchTargets` representation of `table` which groups value and target pairs, unlike it is now?
- Are there any other dead easy optimisations we should spend time implementing for the IR before moving onto codegen?
- Should we move optimisation passes on IR into its own crate?

### DIFF
--- a/compiler/hash-utils/src/store.rs
+++ b/compiler/hash-utils/src/store.rs
@@ -775,7 +775,7 @@ impl<K: Copy + Eq + Hash, V: Clone> PartialStore<K, V> for DefaultPartialStore<K
 /// [`RefCell<Vec<_>>`], so that there isn't additional overhead when
 /// reading/writing elements; you can take references out of it. The trade-off
 /// is that the value needs to implement [`Copy`].
-pub trait CellStore<Key: StoreKey, Value: Copy> {
+pub trait AppendOnlyStore<Key: StoreKey, Value: Copy> {
     /// Get a reference to the internal data of the store.
     ///
     /// This should only be used to implement new store methods, not to access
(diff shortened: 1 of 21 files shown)


compiler/hash-utils/src/store.rs
```diff
@@ -765,30 +780,59 @@ pub trait CellStore<Key: StoreKey, Value: Copy> {
...
         &data[key.to_index()]
     }
+
+    // fn set(&self, key: Key, value: Value) {
+    //     let data = self.internal_data();
+    //     data[key.to_index()] = value;
+    // }
+}
+
+/// A default implementation of [`PartialStore`].
```
kontheocharis, line 814: Wrong comment

compiler/hash-utils/src/store.rs
```diff
@@ -765,30 +780,59 @@ pub trait CellStore<Key: StoreKey, Value: Copy> {
...
+    fn get(&self, key: Key) -> &Value {
         let data = self.internal_data();
         &data[key.to_index()]
     }
+
+    // fn set(&self, key: Key, value: Value) {
+    //     let data = self.internal_data();
+    //     data[key.to_index()] = value;
+    // }
+}
```
kontheocharis, line 812: Delete?

compiler/hash-utils/src/store.rs
```diff
@@ -765,30 +780,59 @@ pub trait CellStore<Key: StoreKey, Value: Copy> {
     ///
     /// This should only be used to implement new store methods, not to access
     /// the store.
-    fn internal_data(&self) -> &AppendOnlyVec<Cell<Value>>;
+    fn internal_data(&self) -> &AppendOnlyVec<Value>;
```
kontheocharis, line 783: Maybe now the store should be called `AppendOnlyStore`?

kontheocharis (changes requested):

kontheocharis (approved):

Issue #658, opened 2022-12-27
Fix issue with definition parameters not working properly


kontheocharis:
Currently, writing:
```
i32 := struct();

Bar := mod {
  Bor := enum <T> (Bing, Bang(i32))
};

foo := () -> Bar::Bor => {
  x := Bar::Bor<()>::Bang(i32());
};
```

produces

```
error[0035]: mismatch in parameter groups: expected 1 groups but got 0

error[0035]: mismatch in parameter groups: expected 0 groups but got 1
```

when it should typecheck instead.

PR #663, opened 2022-12-27
Add a symbol resolution pass to the typechecker

### LINKED ISSUE #634: Resolve symbols to their bindings and store them in the term store

### LINKED ISSUE #664: Add better error messages for symbol not found "Cannot find X in Y"

kontheocharis:
This PR adds the symbol resolution pass to the typechecker, which resolves all variables, accesses, and constructors to their  correct types from the definitions found in the scope discovery stage. Here, scoping errors are caught and reported.

The pass first traverses each type and expression (and soon pattern), and tries to parse it as an access path (with optional arguments). If this succeeds, its type/term is resolved relative to the environment and recorded on the AST node. The next stages will perform further type inference and checking based on the results of this stage.

Closes #634, closes #664.

This PR does not tackle the following items, which will instead be part of subsequent PRs:

- Handle pattern accesses and constructors (#656)
- Add scoping tests to ensure the correctness of the implementation (#657)
- Fix some issues regarding definition arguments (#658)
- Improve the formatting of some typechecking errors (#659)


compiler/hash-typecheck/src/new/passes/state.rs
```diff
@@ -0,0 +1,102 @@
+//! Helper structs for keeping track of state during AST traversal.
```
feds01, line 1: You could move this into `utils`, I am sure there are other places that benefit this.

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -45,13 +102,112 @@ impl<'tc> AstPass for SymbolResolutionPass<'tc> {
...
+        f: impl FnOnce() -> T,
+    ) -> T {
+        self.context_ops().enter_scope(kind, || {
+            self.bindings_by_name.enter(
+                |b| {
+                    let current_level = self.context().get_current_scope_level();
+
+                    // Populate the map with all the bindings in the current
+                    // scope. Any duplicate names will be shadowed by the last entry.
+                    // @@Semantics: Should we report an error if there are duplicate names?
```
feds01, line 169: This should be allowed to be done

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
-/// @@Temporary: for now this visitor just walks the AST and enters scopes. The
-/// next step is to resolve symbols in these scopes!.
+/// An argument group in the AST.
+///
+/// This is either a group of explicit `(a, b, c)` arguments, or a group of
+/// implicit `<a, b, c>` arguments. The former corresponds to the
+/// [`ast::ConstructorCallArg`], while the latter corresponds to the
+/// [`ast::TyArg`].
+#[derive(Copy, Clone, Debug)]
+#[allow(dead_code)]
```
feds01, line 286: Leave a `@@Temporary` so this isn't forgotten

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+enum AstArgGroup<'a> {
+    /// A group of explicit `(a, b, c)` arguments.
+    ExplicitArgs(&'a ast::AstNodes<ast::ConstructorCallArg>),
+    /// A group of implicit `<a, b, c>` arguments.
+    ImplicitArgs(&'a ast::AstNodes<ast::TyArg>),
+}
```
feds01, line 292: I don't quite understand the naming conventions, why is `(a, b, c)` "explicit", but `<a, b, c>` is "implicit"?

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+/// A path component is either a name, or a name followed by a list of
+/// argument groups, each of which is a [`AstArgGroup`].
+///
+/// For example: `Foo::Bar::Baz<T, U>(a, b, c)::Bin(3)`.
+#[derive(Clone, Debug)]
+struct AstPathComponent<'a> {
+    pub(self) name: Identifier,
+    pub(self) name_span: Span,
+    pub(self) args: Vec<AstArgGroup<'a>>,
+    pub(self) original_node_id: ast::AstNodeId,
```
feds01, line 316: Probably just `node_id`

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+type AstPath<'a> = Vec<AstPathComponent<'a>>;
+
+impl AstPathComponent<'_> {
+    /// Get the span of this path component.
+    pub(self) fn span(&self) -> Span {
+        self.name_span.join(
+            self.args
+                .iter()
+                .fold(self.name_span, |acc, arg| arg.span().map(|s| acc.join(s)).unwrap_or(acc)),
+        )
```
feds01, line 331: For args, it's good enough to probably skip until the last and then `join`

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+        let mut errors: Vec<TcError> = vec![];
+        for (param_group_index, arg_group_index) in
+            def_params.to_index_range().zip(created_def_args.to_index_range())
+        {
+            let def_param_group =
+                self.stores().def_params().get_element((def_params, param_group_index));
+            let def_arg_group =
+                self.stores().def_args().get_element((created_def_args, arg_group_index));
+
+            if def_param_group.params.len() != def_arg_group.args.len() {
```
feds01, line 488: Add a note here about default params, as it looks like it's not dealing with it.

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+                // Collect errors and only report at the end.
+                errors.push(TcError::WrongArgLength {
```
feds01, line 490: Why not push to error to store as you go?

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+    /// parameters, checking that the lengths match in the process.
+    ///
+    /// If successful, returns the [`DefArgsId`] that represents the arguments.
+    ///
+    /// Otherwise, returns an error.
+    fn apply_ast_args_to_def_params(
+        &self,
+        def_params: DefParamsId,
+        args: &[AstArgGroup],
+    ) -> TcResult<DefArgsId> {
```
feds01, line 465: In the future, it could be interesting to experiment with using an error id rather than returning the error, this error points to the actual error and can be read instead of propagating it up the call stack. This has the advantage of making the return size only 8 or 12 bytes instead of however big `TcError` is. This change wouldn't be overly difficult to do, and if in the unlikely scenario that the caller wishes to read the error, they still can.

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
```diff
@@ -120,8 +276,644 @@ impl<'tc> SymbolResolutionPass<'tc> {
...
+                        }
+                    }
+                }
+            }
+            BindingKind::Ctor(_, ctor_def_id) => {
+                // A constructor cannot be namespaced further, so it becomes
+                // a term.
+                if in_expr == InExpr::Ty {
+                    // @@Improvement: support shorthand refinement like `None` as `Option<T> of
+                    // None`.
```
feds01, line 573: I think even rust hacks this by just doing `use Option::*` in the prelude

feds01 (changes requested):

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
kontheocharis: As in `<a, b, c>` is optional so you don't have to necessarily provide it and it will be implicitly inferred if you don't

feds01: But default arguments would produce a similar effect when you don't specify them... they would also be "implied"

kontheocharis: Yeah, but that is only for specific arguments in the explicit args, where as a `<..>` group can be fully inferred

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
kontheocharis: Because it should short-circuit if the method has any errors, which would require an additional flag anyway

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
kontheocharis: I agree with this, maybe something to look into later on.

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
kontheocharis: Yeah, this is just to note that this should be supported.

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
feds01: Would it not be better to call it something like `ValueArgs` and `TyArgs`?

kontheocharis: No, because that's not what they are; types and values are valid in both arguments.

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
feds01: It might be worth introducing a similar language feature like `use` which can bring members of data types into scope, but of course not now

compiler/hash-typecheck/src/new/passes/symbol_resolution.rs
feds01: But `ast::ConstructorCallArg` must have a value present I believe?

kontheocharis: Yes, as far as the AST goes it expects values for the explicit parameters and types for the implicit parameters, however as far as the typechecker is concerned they both accept the same thing, since `Type` is a valid type. I am planning on changing the parsing a bit to also allow `(..)` in types and `<..>` in expressions to make it more uniform. Most of the time, `<..>` will be used for types and `(..)` will be used for values, but this is simply because you want types to usually be inferred, and values not. Also by default untyped parameters `<T>` will be assumed to be `<T: Type>` for implicit functions, while this is not the case for explicit functions.

feds01 (approved):

PR #666, opened 2022-12-29
source: Shrink `SourceId` from 12 bytes to 4 bytes

feds01:
This patch simplifies the logic of how sources are identified within the compiler, and significantly cuts down on the memory usage of the representation. Previously, the `SourceId` was defined as:
```rs
enum SourceId {
   Module(ModuleId),
   Interactive(InteractiveId)
}
```
with each sub-id being 8 bytes (which is too big) in size. This design is somewhat wasteful because it forces `SourceId` to be aligned to 12 bytes. Now, `SourceId` is
defined as the following:
```rs
struct SourceId {
    _raw: u32
}
```
This `_raw` value uses the leftmost bit to represent the kind of source it is (either `module` or `interactive`), and the rest of the bits are used to describe the Id of the actual source. This is a reasonable size reduction because the compiler isn't expecting the total amount of sources to exceed $2^{31} - 1$.

This is an effort to reduce the overall memory usage of `SourceLocation` from 20 bytes to 12 bytes. 

This patch also gets rid of a bunch of hashmaps and slotmaps that aren't really needed to represent the `SourceMap` or the `NodeMap`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,9 +635,9 @@ dependencies = [
  "hash-alloc",
  "hash-target",
  "hash-utils",
+ "index_vec",
  "lazy_static",
  "num-bigint",
- "slotmap",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,7 +635,6 @@ dependencies = [
  "hash-alloc",
  "hash-target",
  "hash-utils",
- "index_vec",
  "lazy_static",
  "num-bigint",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,6 +422,7 @@ dependencies = [
  "hash-source",
  "hash-tree-def",
  "hash-utils",
+ "index_vec",
  "num-bigint",
  "replace_with",
 ]
@@ -635,6 +636,7 @@ dependencies = [
  "hash-alloc",
  "hash-target",
  "hash-utils",
+ "index_vec",
  "lazy_static",
  "num-bigint",
 ]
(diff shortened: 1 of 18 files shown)


kontheocharis:
I can understand the advantage of reducing the size of SourceId. However, this loses the type safety of interactive Vs module IDs. Is there a way that this can be kept? For example having newtypes for interactive as well as module IDs, where their constructors verify that the input is valid, and still being a u32 in size. It would remove all those debug asserts.

On another note I wonder if it might be easier to just use an i32 instead of a u32 and if it is negative it is interactive and if it is positive it is a module.. instead of the bit masking.

kontheocharis:
Alternatively https://docs.rs/packed_struct/latest/packed_struct/ might be helpful to not have to resort to bit fiddling, and instead declare the packed members

feds01:
> Alternatively https://docs.rs/packed_struct/latest/packed_struct/ might be helpful to not have to resort to bit fiddling, and instead declare the packed members
 
I think this is a better approach, but it will likely not remove the debug assertions... since it would still have to check in some places that it got the right value.

feds01:
> I can understand the advantage of reducing the size of SourceId. However, this loses the type safety of interactive Vs module IDs. Is there a way that this can be kept? For example having newtypes for interactive as well as module IDs, where their constructors verify that the input is valid, and still being a u32 in size. It would remove all those debug asserts.
> 
> On another note I wonder if it might be easier to just use an i32 instead of a u32 and if it is negative it is interactive and if it is positive it is a module.. instead of the bit masking.

The other reason to get rid of the slot map keys is that using keys involves a computation to actually get a particular value from the map, it is not a pure index.

kontheocharis:
I see, so wait `packed_struct` does not actually change the layout? That's a bit useless...
Anyway, I think this is fine, just have two more types `ModuleId` and `InteractiveId` which can be safely transmuted to `SourceId` (but not vice versa without a check). This would be the best of both worlds no?

feds01:
I think as a compromise, the `SourceId` could have APIs that convert the raw value into an index key like `ModuleId` and `InteractiveId` (which is just a wrapped `u32` value), and all of the functions that expects a particular kind of source will only accept it

kontheocharis:
Yeah, though ModuleId and InteractiveId's memory representation should be the same as SourceId so that it is zero-cost.

kontheocharis (approved):

PR #671, opened 2023-01-04
ir+lower: add more utlities for traversing CFG

feds01:
This patch introduces a common module in `hash-utils` that will store all graph-related algorithms and utilities. This module provides a generic `DirectedGraph` interface that then implements further traits that allow for various traversal and statistic algorithms to be implemented for the general `DirectedGraph`.

This should be reviewed after #670


compiler/hash-utils/src/graph/mod.rs
```diff
@@ -0,0 +1,67 @@
+//! This module contains graph algorithms and utilities that are used by the
+//! compiler.
+
+pub mod dominators;
+pub mod visit;
+
+use index_vec::Idx;
+
+/// This trait can be used to represent a generic directed
+/// graph. The index of each graph node used a [`index_vec::Idx`].
```
kontheocharis, line 10: Why using indexvec?

compiler/hash-utils/src/graph/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+pub trait DirectedGraph {
```
kontheocharis, line 11: Surely one needs more info to call this a directed graph?

kontheocharis: Why is DirectedGraph split into WithPredecessors and WithSuccessors?

compiler/hash-utils/src/graph/mod.rs
```diff
@@ -0,0 +1,67 @@
+//! This module contains graph algorithms and utilities that are used by the
+//! compiler.
+
```
kontheocharis, line 3: `//! Based on rustc: https://github.com/rust-lang/rust/blob/master/compiler/rustc_data_structures/src/graph/mod.rs`?

compiler/hash-utils/src/graph/mod.rs
feds01: Yeah it could be just combined into `DirectedGraph`

compiler/hash-utils/src/graph/mod.rs
feds01: Since it is most likely that node identifiers will use an `index_vec` to store the nodes.

compiler/hash-utils/src/graph/mod.rs
feds01: Also, what additional information is needed to call it a `DirectedGraph`?

kontheocharis: Well a directed graph is a set of nodes N and for each ordered pair of nodes (a, b) in NxN, an optional edge.

For example,

```rust
pub trait DirectedGraph {
   type Node;
   type Edge;
   type SuccessorsIter;
   type PredecessorsIter;

   // Optional method to get an edge between two nodes if it exists
   // By default, it can be implemented as `self.successors(a).find_map(|&x| x == b)`
   fn edge(&self, a: Self::Node, b: Self::Node) -> Optional<Self::Edge>;

   fn successors(&self, a: Self::Node) -> Self::SuccessorsIter;
   fn predecessors(&self, a: Self::Node) -> Self::PredecessorsIter;
}
```

feds01: There isn’t much need to access this information, only successor and predecessor queries are required

kontheocharis: Also it's probably better if `Node` is a generic parameter rather than an assoc type

feds01: Any particular reason?

kontheocharis:
@feds01 status?

feds01:
> @feds01 status?

I think it's sensible to merge `WithSuccessors` and `WithPredecessors` with the main trait, but I think it is better to keep the associated type

kontheocharis:
I just thought that generic makes more sense if DirectedGraph is used a data structure since `Vec<T>` >> `Vec<Item=T>`. But I don't mind either way, so good to merge if you want.

kontheocharis (approved):

feds01:
I think we should re-work it later when the use cases are more clear as there is only one place that is using this (for now).

PR #685, opened 2023-01-11
codegen: implement lowering `RValue`s

feds01:
This patch implements the lowering procedures for all `RValue` kinds.

### DIFF
--- a/compiler/hash-codegen/src/common.rs
+++ b/compiler/hash-codegen/src/common.rs
@@ -228,16 +228,18 @@ pub enum RealComparisonKind {
     True,
 }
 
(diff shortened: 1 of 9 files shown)


compiler/hash-codegen/src/common.rs
```diff
@@ -193,6 +228,20 @@ pub enum RealComparisonKind {
     True,
 }
 
+impl From<ir::BinOp> for RealComparisonKind {
```
kontheocharis, line 231: Maybe `TryFrom`?

kontheocharis (dismissed):

kontheocharis (approved):

PR #690, opened 2023-01-13
Add a declarative way of defining intrinsics

### LINKED ISSUE #672: Add a declarative way of adding intrinsic functions

kontheocharis:
This PR fills out a new crate called `hash-intrinsics`, which contains definitions for intrinsic functions as well as primitive data types. This is done so that intrinsics and primitives can be added in a declarative way, and that intrinsics internally are Rust functions that operate on the TC environment. The goal is for the lowering to treat each intrinsic function call as special and perform the appropriate transformation (i.e. `Intrinsics::add_i32` ~> signed add instruction on target).

This PR also moves some ops from hash-typecheck to hash-types since they do not require typechecking facilities and are useful outside of `hash-typecheck`.

Closes #672


compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -0,0 +1,123 @@
...
+/// corresponding function definition (in order to be able to call it from the
+/// code), and an implementation which is a Rust function that is called when
+/// the intrinsic is called.
+pub struct Intrinsic {
+    /// The ID of the intrinsic.
+    pub id: IntrinsicId,
+    /// The function definition of the intrinsic.
+    pub fn_def: FnDefId,
+    /// The implementation of the intrinsic.
+    pub implementation: fn(&(dyn AccessToPrimitives), &[TermId]) -> TermId,
```
feds01, line 29: who is this intended to be used by, presumably the evaluation part of tc?

kontheocharis: yeah. will add

feds01 (approved):

PR #699, opened 2023-01-13
layout: implement computing the layout for all `IrTy`s

### LINKED ISSUE #646: write a type layout algorithm, and decide on specifcs

feds01:
This PR implements all of the logic responsible for computing a type's layout. 

The basic algorithm for struct-like (univariant) types is to essentially sort the fields by the following rules:

- push all of the ZST-like fields to the start of the struct to avoid dealing with them between other fields
- sort the remaining fields in order of "effective" alignment of each field, essentially the largest fields by
size and alignment are grouped first, and then descending down. 
- try and optimise the ABI of the given type to represent it as a scalar which means it can reach more 
optimisations when code is generated for this kind.

For enum ADTs, the algorithm is the following:
- Compute the layout of the enum tag
- Compute the layout of the inner ADT of each enum variant with the provided enum tag
- Potentially adjust the enum tag if it can be nicely aligned with the rest of the enum variant alignments.
- Attempt to optimise the ABI in order to represent the enum as a scalar or a pair of scalars.

For unions:
- compute the layout of each field
- take the maximum size and alignment of the fields and propagate it to the layout of the union.
- Attempt to optimise the ABI in order to see if the layout can be represented as a scalar.

## Left to do

- Integrate the logic with the general code generation interface to call these methods and query type layouts.
- Add a nice formatter for `Layout` so that layout can be visualised for debugging purposes.
- Add a directive `#layout_of` which will allow the user to print the computed layout of a particular type.

closes #646

### DIFF
--- a/compiler/hash-layout/src/compute.rs
+++ b/compiler/hash-layout/src/compute.rs
@@ -1,0 +1,159 @@
+//! Contains all of the logic that computes the layout of an [IrTy].
+//! This logic is also designed to avoid doing as much duplicate work
+//! as possible, thus using a [LayoutCache] in order to cache all the
+//! previously computed layouts, and re-use them as much as possible
+
+use hash_ir::ty::{AdtId, IrTy, IrTyId, RefKind, VariantIdx};
+use hash_target::{
+    abi::{AbiRepresentation, Integer, Scalar, ScalarKind, ValidScalarRange},
+    layout::TargetDataLayout,
+    primitives::{FloatTy, SIntTy, UIntTy},
+    size::Size,
+};
+use hash_utils::store::Store;
+
+use crate::{Layout, LayoutCtx, LayoutId, LayoutShape, Variants};
+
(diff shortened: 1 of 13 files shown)


compiler/hash-layout/src/compute.rs
```diff
@@ -0,0 +1,772 @@
...
+            // If we can represent all of the variant layouts as a scalar,
+            // we can then use a scalar-pair representation
+
+            let mut common_prim = None;
+            let mut common_prim_initialised_in_all_variants = true;
+
+            for (field_layouts, variant_layout) in field_layouts.iter().zip(variant_layouts) {
+                // All variant layouts must be a struct
+                let LayoutShape::Struct { ref offsets, .. } = variant_layout.shape else {
+                    panic!()
```
kontheocharis, line 643: empty panic

kontheocharis (changes requested):

kontheocharis (approved):

PR #700, opened 2023-01-15
Layout: implement `#layout_of` directive and layout visualisation

feds01:
This patch implements the `#layout_of` directive which allows the user to dump a prettified version of a type's layout. Additionally, this patch implements the actual logic to pretty print a type's layout. Some examples are:

```rs
#layout_of Value := struct (
    k: Kind, // 1 byte
    x: i32, // 4 bytes
    z: i16, // 2 bytes
    y: i32, // 4 bytes
);
```
is displayed as:
```
Layout of `Value` (size=20b align=4b):
┌────────────┬────────────┬─────────────┬─────────────┬────────┐
│  k: Kind   │   x: i32   │   y: i32    │   z: i16    │ 2b pad │
│            │            │             │             │        │
│   size: 8b │   size: 4b │   size: 4b  │   size: 2b  │   ##   │
│ offset: 0b │ offset: 8b │ offset: 12b │ offset: 16b │   ##   │
│  align: 4b │  align: 4b │  align: 4b  │  align: 4b  │   ##   │
└────────────┴────────────┴─────────────┴─────────────┴────────┘
```

The following enum with fields is displayed as:
```rs
#layout_of Kind := enum (
    A(u16, u8), 
    B,
    C(u32),
    D
);
```

```
Layout of `Kind` (size=8b align=4b):
┌────────┬─────────────────────────────┐
│ 1b tag │                             │
├───┬────┴───┬────────────┬────────────┤
│ 0 │ 1b pad │   1: u8    │   0: u16   │
│   │        │            │            │
│   │   ##   │   size: 1b │   size: 2b │
│   │   ##   │ offset: 1b │ offset: 2b │
│   │   ##   │  align: 2b │  align: 2b │
├───┼────────┼────────────┴────────────┤
│ 1 │ 1b pad │                         │
├───┼────────┴───┬─────────────────────┤
│ 2 │   0: u32   │                     │
│   │            │                     │
│   │   size: 4b │                     │
│   │ offset: 4b │                     │
│   │  align: 4b │                     │
├───┼────────┬───┴─────────────────────┤
│ 3 │ 1b pad │                         │
└───┴────────┴─────────────────────────┘
```


compiler/hash-codegen/src/lower/rvalue.rs
```diff
@@ -497,7 +497,7 @@ impl<'b, Builder: BlockBuilderMethods<'b>> FnBuilder<'b, Builder> {
 
                 // check if the type is a ZST, and if so this satisfies the
                 // case that the rvalue creates an operand...
-                self.ctx.layout_of(ty).is_zst(self.ctx.layout_ctx())
+                self.ctx.layout_of(ty).is_zst(self.ctx.layout_computer())
```
kontheocharis, line 500: i dont do computers

compiler/hash-layout/src/write.rs
```diff
@@ -0,0 +1,864 @@
...
+    /// ```ignore
+    /// +-------------+
+    /// | 1byte (pad) |
+    /// +-------------+
+    /// ```
+    pub fn height(&self) -> usize {
+        if let Self::Content { title, content } = self {
+            // If there are no lines, then don't use any space, otherwise
+            // add one line of padding and the rest as the number of lines
+            let lines = if content.is_empty() { 0 } else { content.lines().count() + 1 };
```
kontheocharis, line 204: .lines() creates owned lines, maybe store as a vec of string in the first place?

kontheocharis (commented):
I wonder if some of this code can be simplified if you separate the drawing logic from the logic that figures out the layout.. Cause there seems to be a lot of bookkeeping relating to connectors and formatting alignment etc

kontheocharis:
Regardless, happy to merge, lmk

compiler/hash-layout/src/write.rs
kontheocharis: never mind im dumb

kontheocharis (approved):

PR #716, opened 2023-01-23
Separate typechecking from semantics, and improve diagnostics

### LINKED ISSUE #709: Split semantics from TC

kontheocharis:
The typechecking is now independent of the rest of the semantics pass, and lives in a separate crate.

The diagnostics API has been changed slightly to `Diagnostics` and `AccessToDiagnostics` traits. This is useful to access diagnostics methods from a root object, without having to specifically refer to the diagnostics implementor.

Closes #709


compiler/hash-parser/src/diagnostics/warning.rs
```diff
@@ -60,7 +60,7 @@ impl Display for SubjectKind {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
feds01, line 63: Why is this needing cloning?

compiler/hash-reporting/src/diagnostic.rs
```diff
@@ -2,50 +2,261 @@
...
+        )*
+    };
+}
+
+make_diagnostic_traits!(Diagnostics with &Self, DiagnosticsMut with &mut Self);
+
+pub struct ImmutableDiagnostics<E, W> {
+    pub errors: RefCell<Vec<E>>,
+    pub warnings: RefCell<Vec<W>>,
+}
```
feds01, line 77: `@@Docs`, also would be nice to get rid of refcells somehow?

compiler/hash-semantics/src/new/ops/context.rs
```diff
@@ -1,189 +1 @@
...
-        environment::{
-            context::{Binding, BindingKind, BoundVarOrigin, ScopeKind},
-            env::AccessToEnv,
-        },
-        params::{DefParamIndex, ParamId, ParamsId},
-        scopes::StackMemberId,
-    },
-    ty_as_variant,
-};
-use hash_utils::store::{SequenceStore, Store};
```
feds01, line 1: File not deleted?

compiler/hash-semantics/src/storage/cache.rs
```diff
@@ -61,12 +61,6 @@ impl<K: Copy + Hash + Eq, V: Clone> PartialStore<K, V> for CacheStore<K, V> {
     fn internal_data(&self) -> &RefCell<HashMap<K, V>> {
         self.store.internal_data()
     }
-
-    /// Clear the [CacheStore] and metrics.
-    fn clear(&self) {
-        self.store.clear();
-        self.reset_metrics();
-    }
```
feds01, line 69: Was this does not being used before to reset the cache after each module was being walked?

feds01 (changes requested):

compiler/hash-reporting/src/diagnostic.rs
kontheocharis: Not really possible to get rid of refcells since the access is immutable..

feds01: Oh right, the naming is a little confusing, as in it needs to be within an immutable context to append things to the diagnostic store.

kontheocharis: I'm open to better naming ideas

feds01: Maybe something like `DiagnosticMut` to `Diagnostics`, and `ImmutableDiagnostics` to `CelledDiagnostics`?

compiler/hash-semantics/src/storage/cache.rs
kontheocharis: Idk why that was deleted, added it back.

compiler/hash-reporting/src/diagnostic.rs
kontheocharis: `DiagnosticStore` and `DiagnosticCellStore` ? Cause `Diagnostics` is the name of the trait

feds01 (approved):

PR #718, opened 2023-01-23
tests: add capability to pass arguments and capture `stdout` and `stderr`

feds01:
This patch allows for test cases to pass arguments to the compiler. 

This meant that the argument parsing had to be refactored in order to support arbitrarily specifying keys and values (which is why this patch deletes all of the old `clap.rs` code handling).  In the new implementation, a custom (not fully featured) command-line argument parser supports passing arbitrary key-value pairs and applying them to the `CompilerSettings`.

In addition to allowing for arguments passed, the tests now also capture compiler standard output and standard error streams. This meant that a new compiler streams API had to be introduced in order to abstract away just "writing" to either of these streams. This is now done via getting the desired stream, and then using the handy macro `stream_writeln!`. This means that the code base should not use `eprintln|println` macros when wanting to output content that is user-facing. For cases where it is not required, or doesn't make sense to use, one can use `streamless_writeln!` and streamless_ewriteln!` to write to the standard streams. If it is not possible to use these macros, then it is acceptable to use a `println`, however, these should be avoided at all costs in order to remove confusion between all of these new macros. 

All of this works means that things like AST tree printing, IR generation printing, Layout printing, Type tree printing, etc is now all testable via our testing eco-system 🥳 !


tests/cases/lowering/functions.stdout
```diff
@@ -0,0 +1,23 @@
...
+}
+
+
+IR dump for function `main` defined at $DIR/functions.hash:6:9-8:2
+main() -> () {
+    mut _0: ();
+
+    bb0 {
+        _0 = const <unevaluated> print(const "Hello, world!") -> bb1;
+    }
```
feds01, line 17: IR printing bug

compiler/hash-pipeline/src/args.rs
```diff
@@ -0,0 +1,252 @@
...
+    } else if arg.starts_with("--") {
+        let key = arg.trim_start_matches("--").to_string();
+
+        match key.as_str() {
+            "debug" => {
+                settings.debug = true;
+            }
+            "output-metrics" => {
+                settings.output_metrics = true;
+            }
```
kontheocharis, line 155: Would be nice to have these in some macro form eventually..

compiler/hash-utils/src/printing.rs
```diff
@@ -3,6 +3,41 @@
...
 use std::fmt;
 
+/// This is used a wrapper around [`writeln!`] which integrates writing
+/// to a specified stream.
+///
+/// This macro conveniently unwraps the result of the write operation.
+#[macro_export]
+macro_rules! stream_writeln {
+    ($stream:expr, $($arg:tt)*) => {
+        use std::io::Write;
```
kontheocharis, line 13: These will sometimes produce unused import warnings

tests/testing-macros/src/lib.rs
```diff
@@ -221,19 +221,21 @@ pub fn generate_tests(input: TokenStream) -> TokenStream {
...
     let filenames = entries.iter().map(|entry| entry.path.file_prefix().unwrap().to_str().unwrap());
-    let case_metadata = entries.iter().map(|entry| entry.metadata);
 
     // Create the test names from the provided prefix and the computed `snake_name`
     let test_names = entries
         .iter()
         .map(|entry| format_ident!("{}_test_{}", input.test_prefix, entry.snake_name));
 
+    // @@Copying: we're cloning the metadata here...
+    let case_metadata = entries.iter().map(|entry| entry.metadata.clone());
```
kontheocharis, line 231: Why is this necessaru?

compiler/hash-pipeline/src/args.rs
feds01: Yes, it would be nice, and it could capture the type of argument, some kind of description, and potentially some parsing parameters. Ill add a `@@Todo`...

compiler/hash-utils/src/printing.rs
feds01: No, it doesn't because this actually requires the `Write` trait to be in scope.

kontheocharis: Alright.,

kontheocharis (approved):

PR #719, opened 2023-01-24
Implement traversal utilities for TIR

### LINKED ISSUE #717: Implement traversal for terms, types, and patterns

kontheocharis:
This PR adds `visit_*` and `fmap_*` methods for constructs in the TIR, in order to be able to traverse them without having to write out all the recursive logic every time.

The next PR will use this to implement substitutions.

closes #717


compiler/hash-tir/src/new/tuples.rs
```diff
@@ -41,7 +41,7 @@ pub struct TupleTerm {
 pub struct TuplePat {
     /// The original tuple type, if known or given as part of the literal (might
     /// contain holes).
-    pub original_ty: Option<TupleTy>,
+    // pub original_ty: Option<TupleTy>,
```
feds01, line 44: why are these commented out?

kontheocharis: Yeah, i removed it cause it is not needed anymore (inference will store the result in cache if needed). Added commit to remove commented lines.

feds01 (approved):

PR #756, opened 2023-02-01
Remove DefParams/DefArgs/DefPatArgs in favour of the simple versions

### LINKED ISSUE #755: Remove def params/args/etc

kontheocharis:
Before, data types could contain multiple groups of parameters like `<T: Type> (t: T) ...`. This has been removed to simplify the implementation, since even the simple `Params` etc will allow for dependent referencing (e.g. `(T: Type, t: T)`). Now data types are only annotated with a single`<...>` parameter list. This also should cut down on memory usage, and nicely shaves off almost 1k lines of boilerplate.

Closes #755

### DIFF
--- a/compiler/hash-semantics/src/new/diagnostics/error.rs
+++ b/compiler/hash-semantics/src/new/diagnostics/error.rs
@@ -74,6 +74,7 @@ impl<'tc> WithTcEnv<'tc, &SemanticError> {
     /// Format the error nicely and add it to the given reporter.
     fn add_to_reporter(&self, reporter: &mut Reporter) {
         let locations = self.tc_env().stores().location();
+        // @@ErrorReporting: improve error messages and locations
         match &self.value {
             SemanticError::Signal => {}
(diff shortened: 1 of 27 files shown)


compiler/hash-semantics/src/new/diagnostics/error.rs
```diff
@@ -226,6 +228,16 @@ impl<'tc> WithTcEnv<'tc, &SemanticError> {
...
+            SemanticError::UnexpectedArguments { location } => {
+                let error = reporter
+                    .error()
+                    .code(HashErrorCode::ValueCannotBeUsedAsType)
+                    .title("unexpected arguments given to subject");
+
+                error
+                    .add_span(*location)
+                    .add_info("cannot use these arguments as the subject does not expect them");
+            }
```
kontheocharis, line 240: This error will be improved later

feds01: maybe leave a `@@` `for this?

feds01 (approved):

PR #771, opened 2023-02-04
codegen: emit a wrapper for entry point

### LINKED ISSUE #647: Detect and check for program entry point
Currently, a program is not expected to specify an entry point which is fine until code generation when it needs to know 
where the program will start (disregarding pre-start setup). 

### When is it required that an entry point is specified

For many programs and runs, it isn't necessary that the main function is specified. For example, if it is only to type check or 
to see the result of a lowering operation (for debugging purposes), the compiler should not insist that a `main` function is present. Therefore, the following rules for the need for an entry point function should be as follows:
- The compiler is instructed to generate code, be it compiled or VM, it should ensure that there is an entry point function.
- The compiler is not running in interactive mode, since it is already within a running context.

### When should a check for an entry point happen?

Ideally, it should either happen as semantic analysis or typechecking (preferably typechecking since the signature of the function also matters). However, it is not the responsibility of this stage to "report" that an entry point is missing. What this stage should offer is the ability for later stages to query for the entry point (probably some kind of `DefId` which will correspond to the generated IR). The stage that cares about this will perform a check and fetch a reference to the function. If it cannot find the function, then it will report the error.

However, if the function signature mismatches with the expected entry point signature, the typechecking stage should probably report this.

So, based on this, the scheme for ensuring there exists an entry point is:
```
      Discovery -------------> Signature Verification ------------> Query
(semantic analysis)                (Typechecking)               (IR/Codegen)
```

### The entry point

The entry point should be a function declaration at the root scope of the entry module like so:

```rs
main := () => { ... }
```

It is unclear what the signature of the function should be, for example, should have the classic c-style argument length and arguments, or should it be an argumentless function with no specified return type? We need to decide this...

### How to specify an entry point?

The programmer is expected to specify a `main` function within the file path that they pass to the compiler which is treated as an entry point of the whole program that is to be compiled. However, the programmer should have the ability to specify a "custom" entry point in some way. The only sensible option I see is to add a `#entrypoint` directive, which has the following semantics:
- The argument to the directive must be a declaration like `main := () => ...` 
- The directive must be specified at the top level of the module, so it can't be used within a function body, etc.
- The directive can only be specified once, multiple specifications are considered as a hard error.
- The directive cannot be specified in a non-entry module (i.e. it can't be used in a module that is imported into the initial module.

feds01:
This patch (+ again another round of bugfixes) implements the logic that deals with program entry points. The compiler will now ensure that a program contains an entry point if an executable is expected to be produced from the pipeline.

This patch also adds basic symbol mangling which is required to flatten all function definitions and avoid naming clashes. Whilst the mangling scheme is definitely not final, it is a starting point to get things working in the first place. Some TODOs have been left where needed, and they will be tackled when we need to come up with a proper naming scheme.

Since we don't have a notion of intrinsic/language items yet being fully supported, we can't really deal with the program arguments being passed in and dealing with a return value of the function. For now, the main entry point code generates the following wrapper:
```llvm
i32 @main(i32 argc, ptr argv) {
    call user_main;
    ret 0;
}
```
In the future, what should happen is a function should be called (i.e. `lang_start`) which deals with the function arguments, and conveniently returns a correct integer value for the return code.

closes #647


tests/cases/semantics/directives/entry_point.stderr
```diff
@@ -0,0 +1,17 @@
+error: the `entry_point` directive expects a `declaration` as an argument
+ --> $DIR/entry_point.hash:6:18
+5 |   foo := {
+6 |       #entry_point 1;
+  |                    ^ a `literal` cannot be given to the `entry_point` directive
+7 |   
+
+error: the `entry_point` directive is must be within a module block
```
kontheocharis, line 8: Why? Also typo

kontheocharis (changes requested):

kontheocharis (approved):

PR #772, opened 2023-02-04
Implement matching and reordering of arguments

### LINKED ISSUE #698: Implement parameter matching

kontheocharis:
This PR deals with reordering arguments and pattern arguments to match their parameter order.

The arguments are validated to all exist (if they have names), to have the right length, and then they are rearranged to the order of the parameters. For term arguments, default values are populated if not provided. For pattern arguments, if there is a spread, remaining arguments are captured by it, or otherwise all arguments are validated to exist.

In order to denote arguments that have been captured by a spread, the `Capture` variant is introduced for patterns and pattern lists.

This strategy will be used for function parameters, tuples, constructors.

Closes #698

### DIFF
--- a/compiler/hash-typecheck/src/params.rs
+++ b/compiler/hash-typecheck/src/params.rs
@@ -2,8 +2,9 @@ use std::collections::HashSet;
 
 use derive_more::{Constructor, Deref};
 use hash_tir::new::{
-    args::{ArgData, ArgId, ArgsId},
+    args::{ArgData, ArgsId, PatArgData, PatArgsId, PatOrCapture, SomeArgId, SomeArgsId},
     params::{ParamId, ParamIndex, ParamsId},
+    pats::Spread,
     utils::{common::CommonUtils, AccessToUtils},
 };
 use hash_utils::store::{SequenceStore, SequenceStoreKey};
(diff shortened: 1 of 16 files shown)


compiler/hash-typecheck/src/params.rs
```diff
@@ -0,0 +1,381 @@
...
+    ) -> TcResult<()> {
+        let mut error_state = self.new_error_state();
+
+        // Check for too many arguments
+        if args_id.len() > params_id.len() {
+            error_state.add_error(ParamError::TooManyArgs { expected: params_id, got: args_id });
+        }
+
+        let mut seen = HashSet::new();
+        let mut found_named = None;
```
feds01, line 97: it would be nice to use a bitset or something with just positional indices here, but of course, it might not be worth the complexity...

compiler/hash-typecheck/src/params.rs
```diff
@@ -0,0 +1,381 @@
...
+        &self,
+        args_id: SomeArgsId,
+        params_id: ParamsId,
+    ) -> TcResult<()> {
+        let mut error_state = self.new_error_state();
+
+        // Check for too many arguments
+        if args_id.len() > params_id.len() {
+            error_state.add_error(ParamError::TooManyArgs { expected: params_id, got: args_id });
+        }
```
feds01, line 94: seems that this function doesn't deal with defaults yet?

feds01 (changes requested):

compiler/hash-typecheck/src/params.rs
kontheocharis: I don't think it would be worth the complexity

compiler/hash-typecheck/src/params.rs
kontheocharis: It does.

compiler/hash-typecheck/src/params.rs
kontheocharis: This stores identifiers not indices

feds01 (approved):

feds01 (approved):

PR #776, opened 2023-02-08
Fix some inference bugs and implement constructor type argument inference

kontheocharis:
This PR follows from #731 by fixing a few bugs and implements inference of constructor type arguments. Now, `Option<_>::Some(3)` is typed as `Option<i32>`. This is currently not implemented for arbitrary functions, just constructors.

There are still a few bugs to fix but now the type-checker is able to check and run a lot of programs! (mod the unimplemented stuff)

### DIFF
--- a/compiler/hash-semantics/src/new/passes/inference/mod.rs
+++ b/compiler/hash-semantics/src/new/passes/inference/mod.rs
@@ -29,7 +29,7 @@ impl_access_to_tc_env!(InferencePass<'_>);
 
 /// The maximum number of hole-filling iterations to perform when inferring a
 /// term, before giving up and reporting an error.
-const MAX_INFER_ITERATIONS: usize = 20;
+const MAX_INFER_ITERATIONS: usize = 2;
 
 impl InferencePass<'_> {
     /// Infer the given subject by the provided closure, until it contains no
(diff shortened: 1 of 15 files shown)


compiler/hash-semantics/src/new/passes/inference/mod.rs
```diff
@@ -29,7 +29,7 @@ impl_access_to_tc_env!(InferencePass<'_>);
 
 /// The maximum number of hole-filling iterations to perform when inferring a
 /// term, before giving up and reporting an error.
-const MAX_INFER_ITERATIONS: usize = 20;
+const MAX_INFER_ITERATIONS: usize = 2;
```
feds01, line 32: It would be nice to comment on why exactly "2", and maybe why it's not valid after that?

feds01: And, it could be made configurable through command line argument i.e. `-Cinference-passes=<?>` with it defaulting to `2`.

kontheocharis: Yeah currently it's just for testing, I will add an option for it later.

kontheocharis: also there's nothing special about 2, it's currently just a random value

feds01: Maybe leave it as a `@@Todo`?

feds01 (approved):

PR #780, opened 2023-02-09
Linker: finish implementing the linker and cleanup cmdargs 

feds01:
This patch implements the linker handling for generated hash sources.

Additionally, this patch cleans up some command-line arguments:
- use `--debug` and `--release` to specify whether to build in debug/release mode.
- use `-Cmetrics` to emit metrics about compilation.
- use `-Cdebug` to enable debug logging for the compiler.
- use `-Cdump=link-line` to print the produced linker line for linking the final executable.

Some other fixes and additions:
- don't mangle intrinsic functions (for now).
- introduce  `ItemAttributes` on `Instance`s in order to store attributes such as `#no_mangle` and `#foreign`.

This is quite the landmark for the Hash compiler since we now have a full workflow from source to executable. 🎉🎉🎉


kontheocharis (approved):
🎖️ 🏅 🥉

PR #789, opened 2023-02-11
source: refactor `IntConstant` to use value slots rather than scalar

feds01:
This PR refactors the handling of integer constants within the compiler to just use integer slots rather than storing them as a singular scalar and converting between one and the other.

### DIFF
--- a/compiler/hash-source/src/constant.rs
+++ b/compiler/hash-source/src/constant.rs
@@ -248,7 +248,7 @@ impl IntConstantValue {
     /// Attempt to convert the given value into a [u128] provided that the
     /// value itself is not [`IntConstantValue::Big`]. If the value is a
     /// big integer, then [`None`] is returned.
-    pub fn as_u128(self) -> Option<u128> {
+    pub fn as_u128(&self) -> Option<u128> {
         match self {
             Self::I8(inner) => {
                 let inner = inner.to_be_bytes();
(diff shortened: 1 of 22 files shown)


compiler/hash-source/src/constant.rs
```diff
@@ -267,54 +396,149 @@ pub struct IntConstant {
...
+        Self { value, suffix: None }
     }
 
-    /// Convert the constant into the "small" variant.
-    pub fn as_small(self) -> Option<u128> {
+    /// Get the bytes of a constant value assuming that it is
+    /// a primitive integer type.
+    ///
+    /// N.B. This returns bytes in Big Endian order.
+    pub fn bytes_be(&self) -> [u8; 16] {
```
kontheocharis, line 464: Get rid of this?

compiler/hash-source/src/constant.rs
```diff
@@ -323,76 +547,86 @@ impl IntConstant {
...
+impl From<BigInt> for IntConstant {
+    fn from(value: BigInt) -> Self {
+        let (sign, mut bytes) = value.to_bytes_be();
+        let value = IntConstantValue::from_be_bytes(&mut bytes, sign == Sign::NoSign);
+
+        Self { value, suffix: None }
     }
 }
 
 impl PartialOrd for IntConstant {
```
kontheocharis, line 612: Derived?

feds01: No, unfortunately not.

kontheocharis (approved):

PR #790, opened 2023-02-11
Add intrinsics for operators and use them in normalisation

### LINKED ISSUE #681: Handle AST operators during resolution

kontheocharis:
This PR adds support for binary and unary operators for primitive types, including their handling in normalisation.

There is still the issue of dealing with overflow, which the normaliser currently panics at (the lowering is free to do what it wants with it since it will substitute the intrinsics with its own stuff).

Closes #681


compiler/hash-intrinsics/src/utils.rs
```diff
@@ -53,6 +76,96 @@ pub trait PrimitiveUtils: AccessToPrimitives + Sized {
...
+    /// Get the given term as a float literal if possible.
+    fn create_term_from_integer_lit<L: Into<BigInt>>(&self, lit: L, int_ty: IntTy) -> TermId {
+        self.new_term(Term::Prim(PrimTerm::Lit(Lit::Int(IntLit {
+            underlying: ast::IntLit {
+                kind: ast::IntLitKind::Unsuffixed,
+                value: CONSTANT_MAP.create_int_constant(IntConstant::from_big_int(
+                    lit.into(),
+                    int_ty,
+                    None,
+                )),
```
feds01, line 133: this will have to change into with the new constant stuff since there is no more `ty` on it.

compiler/hash-pipeline/src/lib.rs
```diff
@@ -161,11 +161,11 @@ impl<I: CompilerInterface> Compiler<I> {
 
             // Temporarily swap the settings with a patched settings in order
             // for the prelude bootstrap to run
-            let mut old_settings = std::mem::take(ctx.settings_mut());
+            // let mut old_settings = std::mem::take(ctx.settings_mut());
```
feds01, line 164: Originally, these were here to ensure that the prelude runs in it's own instance of the settings. However, this probably doesn't make anymore sense because it doesn't need run with its own configuration.

But... The prelude pipeline should not run to "build/link" because  the main pipeline instance will deal with this. So, this has to be limited to not being `CompilerStageKind::Codegen`.

compiler/hash-semantics/src/new/passes/resolution/exprs.rs
```diff
@@ -883,26 +890,89 @@ impl<'tc> ResolutionPass<'tc> {
     /// Make a term from an [`ast::BinaryExpr`].
     fn make_term_from_ast_binary_expr(
```
feds01, line 891: Will you deal with `assign_op_expr` in a subsequent PR?

compiler/hash-semantics/src/new/environment/tc_env.rs
```diff
@@ -62,14 +70,47 @@ macro_rules! tc_env {
...
 tc_env! {
     #hide env: Env<'tc>,
     diagnostics: DiagnosticsStore,
     ast_info: AstInfo,
+    prelude_or_unset: PreludeOrUnset,
     primitives_or_unset: DefinedPrimitivesOrUnset,
     intrinsics_or_unset: DefinedIntrinsicsOrUnset,
 }
 
+pub struct SemanticStorage {
```
feds01, line 84: This should have the `EntryPointState` in some state for the codegen and later.

compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -46,52 +96,727 @@ impl Debug for DefinedIntrinsics {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq, IntoPrimitive, TryFromPrimitive)]
+#[repr(u8)]
+pub enum EndoBinOp {
+    /// '|'
+    BitOr,
+    /// '&'
+    BitAnd,
+    /// '^'
+    BitXor,
+    /// '^^'
```
feds01, line 143: `**`

compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -46,52 +96,727 @@ impl Debug for DefinedIntrinsics {
...
+                            let lhs: BigUint = env.try_use_term_as_integer_lit(lhs).unwrap();
+                            let rhs: BigUint = env.try_use_term_as_integer_lit(rhs).unwrap();
+                            Ok(env.new_bool_term(operate_int!(parsed_op, lhs, rhs)))
+                        }
+                        LitTy::F32 => {
+                            // @@Todo: properly handle f32
+                            let lhs: f64 = env.try_use_term_as_float_lit(lhs).unwrap();
+                            let rhs: f64 = env.try_use_term_as_float_lit(rhs).unwrap();
+                            Ok(env.new_bool_term(operate_float!(parsed_op, lhs, rhs)))
+                        }
```
feds01, line 479: why is this a problem at the moment?

compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -46,52 +96,727 @@ impl Debug for DefinedIntrinsics {
...
+                ParamData { default: None, name: b_sym, ty: env.new_ty(t_sym) },
+            ]
+            .into_iter(),
+        );
+        let ret = env.new_ty(t_sym);
+
+        Self::add_intrinsic(
+            env,
+            implementations,
+            "endo_bin_op",
```
feds01, line 537: "endo" meaning what?

compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -46,52 +96,727 @@ impl Debug for DefinedIntrinsics {
...
+                            let lhs: u32 = env.try_use_term_as_integer_lit(lhs).unwrap();
+                            let rhs: u32 = env.try_use_term_as_integer_lit(rhs).unwrap();
+                            Ok(env.new_bool_term(operate_int!(parsed_op, lhs, rhs)))
+                        }
+                        LitTy::I64 => {
+                            let lhs: i64 = env.try_use_term_as_integer_lit(lhs).unwrap();
+                            let rhs: i64 = env.try_use_term_as_integer_lit(rhs).unwrap();
+                            Ok(env.new_bool_term(operate_int!(parsed_op, lhs, rhs)))
+                        }
+                        LitTy::U64 => {
```
feds01, line 449: This looks like low hanging fruit for a macro?

feds01 (changes requested):

compiler/hash-intrinsics/src/intrinsics.rs
kontheocharis: Just that it has the same return type as its arguments

compiler/hash-semantics/src/new/environment/tc_env.rs
kontheocharis: Yes, I will add this in next PR

feds01 (approved):

PR #792, opened 2023-02-12
lower: write out logic to convert TIR types into IR types

feds01:
This patch kicks off the migration of the compiler from the old typechecker to the new typechecker. This patch implements conversion logic between the TIR types into the IR types.


compiler/hash-lower/src/ty/new.rs
```diff
@@ -0,0 +1,263 @@
...
+                                let fields = fields
+                                    .iter()
+                                    .map(|field| {
+                                        let ty = self.ty_id_from_tir_ty(field.ty);
+                                        let name = self
+                                            .get_symbol(field.name)
+                                            .name
+                                            .unwrap_or(IDENTS.underscore);
+                                        // @@Verify: do we need to store default parameter
+                                        // information here?
```
kontheocharis, line 203: no

compiler/hash-lower/src/ty/new.rs
```diff
@@ -0,0 +1,263 @@
...
+                                    }
+                                }
+                                // We don't implement bigints yet
+                                _ => unimplemented!(),
+                            }
+                        }
+                    }
+
+                    // @@Verify: does `str` now imply that it is a `&str`, or should we create a
+                    // `&str` type here.
```
kontheocharis, line 253: Probably the latter, maybe better to do this once TC deals with it?

kontheocharis (approved):

compiler/hash-lower/src/ty/new.rs
```diff
@@ -0,0 +1,263 @@
...
+                let params = self.stores().params().map_fast(*params, |params| {
+                    self.lcx.tls().create_from_iter(
+                        params.iter().map(|param| self.ty_id_from_tir_ty(param.ty)),
+                    )
+                });
+
+                let return_ty = self.ty_id_from_tir_ty(*return_ty);
+
+                // @@Todo: functions should have a name if possible, otherwise they default
+                // to an underscore.
```
kontheocharis, line 127: What, like function types?

feds01: Yes, the function type/definition should store associated attributes, one of them being a name

compiler/hash-lower/src/ty/new.rs
feds01: As in that it verifies that the `str` does not appear in a type position (except generic parameters)?

kontheocharis: Yeah, surely that should be part of the layout logic?

compiler/hash-lower/src/ty/new.rs
kontheocharis: A function type is different from a function definition.

PR #797, opened 2023-02-15
Fix some issues with normalisation and recursive inference

kontheocharis:
This PR traverses functions in two passes, first to resolve their types and then to resolve their bodies. This allows for recursive calls to work. Furthermore, normalisation of typeof should work better.


compiler/hash-typecheck/src/inference.rs
```diff
@@ -41,6 +40,15 @@ use crate::{
     AccessToTypechecking,
 };
 
+/// The mode in which to infer the type of a function.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FnInferMode {
+    /// Infer the type of a function but do not look at its body.
+    FirstPass,
+    /// Infer the type of a function and its body.
+    Normal,
```
feds01, line 49: Maybe `FnHeader` and `FnBody` instead?

feds01 (dismissed):

feds01 (approved):

PR #810, opened 2023-03-03
ffi: add some core intrinsics and support `repr_c` and `foreign` directives

feds01:
This adds the discussed `#foreign` directive which marks a function as originating from a foreign context. In principle, this means that the function is not lowered and compiled, it is merely a symbol that can be referenced by other functions, and finally should be resolved when the program is linked.

Additionally, in order to support some interfacing between Hash and the outside world, `#repr_c` is introduced as a directive which specifies that the layout of a type must be compatible with the C layout.

This PR also re-adds the `prelude` and defines some core items that can be used to read/write and interface with the outside world. For now, they are only meant as a scaffold until accessing string/array data and length is expressible in the language (which would require the definition of `transmute` to work).

Next steps:
- Ensure C layout doesn't affect anything in code generation
- Implement `transmute` intrinsic in code generation (for odd cases)
- Implement and lower other needed intrinsics.


compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -816,6 +826,101 @@ impl DefinedIntrinsics {
...
+            let t_sym = env.new_symbol("item");
+            let params = env.param_utils().create_params(
+                [ParamData { default: None, name: t_sym, ty: env.new_data_ty(prim.str()) }]
+                    .into_iter(),
+            );
+            let ret = env.new_data_ty(prim.usize());
+            add("len", FnTy::builder().params(params).return_ty(ret).build(), |_, _| {
+                unimplemented!("`len` intrinsic evaluation")
+            })
+        };
```
kontheocharis: Remove

compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -816,6 +826,101 @@ impl DefinedIntrinsics {
...
+            let t_sym = env.new_symbol("T");
+            let params = env.param_utils().create_params(
+                [ParamData { default: None, name: t_sym, ty: env.new_data_ty(prim.str()) }]
+                    .into_iter(),
+            );
+            let ret = env.new_ref_ty(env.new_ty(t_sym), RefKind::Raw, false);
+            add("raw_ptr", FnTy::builder().params(params).return_ty(ret).build(), |_, _| {
+                unimplemented!("`raw_ptr` intrinsic evaluation")
+            })
+        };
```
kontheocharis: Remove, instead use ArrayData struct

kontheocharis (approved):

PR #813, opened 2023-03-04
Fix some issues with match bindings

kontheocharis:
This PR fixes the issue where match bindings do not get added to the match case context. It also simplifies how context is kept track of.

### DIFF
--- a/compiler/hash-typecheck/src/inference.rs
+++ b/compiler/hash-typecheck/src/inference.rs
@@ -56,6 +56,25 @@ pub enum FnInferMode {
     Body,
 }
 
+/// The result of inferring the annotation of an item.
+///
+/// I.e. type of a term, type of a pattern, params of args, etc.
+#[derive(Debug, Clone)]
+pub struct Inference<X, T>(
+    // /// A substitution occurring from the inference.
+    // pub sub: Sub,
+    /// The result of the inference, with the substitution applied.
+    pub X,
+    /// The annotation of the result, with the substitution applied.
+    pub T,
+);
+
+impl<X, T> From<Inference<X, T>> for (X, T) {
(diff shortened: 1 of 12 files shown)


compiler/hash-typecheck/src/inference.rs
```diff
@@ -1305,13 +1338,16 @@ impl<T: AccessToTypechecking> InferenceOps<'_, T> {
 
         match result_decl_term {
             Ok(result_decl_term) => {
-                self.context_utils().add_from_decl_term(&result_decl_term);
-                Ok((result_decl_term, self.check_by_unify(self.new_void_ty(), annotation_ty)?))
+                // self.context_utils().add_from_decl_term(&result_decl_term);
```
feds01, line 1341: commented out?

compiler/hash-typecheck/src/inference.rs
```diff
@@ -1305,13 +1338,16 @@ impl<T: AccessToTypechecking> InferenceOps<'_, T> {
...
+                Ok(Inference(
+                    result_decl_term,
+                    self.check_by_unify(self.new_void_ty(), annotation_ty)?,
+                ))
             }
             Err(err) => {
                 // We still want to add the bindings from the decl term to the context, even if
                 // the term is invalid.
-                self.context_utils().add_from_decl_term(decl_term);
+                // self.context_utils().add_from_decl_term(decl_term);
```
feds01, line 1350: commented out?

feds01 (approved):

compiler/hash-typecheck/src/inference.rs
kontheocharis: oops

feds01 (approved):

PR #818, opened 2023-03-05
Fix some panics during inference

kontheocharis:
Mismatching array lengths no longer panic

### DIFF
--- a/compiler/hash-tir/src/utils/context.rs
+++ b/compiler/hash-tir/src/utils/context.rs
@@ -65,9 +65,25 @@ impl<'env> ContextUtils<'env> {
         });
     }
 
+    /// Get a binding from the current scopes.
+    pub fn get_binding(&self, name: Symbol) -> Binding {
+        self.context()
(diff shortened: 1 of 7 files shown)


compiler/hash-tir/src/utils/context.rs
```diff
@@ -65,9 +65,25 @@ impl<'env> ContextUtils<'env> {
         });
     }
 
+    /// Get a binding from the current scopes.
+    pub fn get_binding(&self, name: Symbol) -> Binding {
+        self.context()
+            .try_get_binding(name)
+            .unwrap_or_else(|| panic!("cannot get binding for {}", self.env().with(name),))
```
feds01, line 72: trailing comma

feds01 (approved):

compiler/hash-tir/src/utils/context.rs
kontheocharis: yes

PR #839, opened 2023-03-12
Implement monomorphisation of terms

### LINKED ISSUE #835: Weak head normalisation

kontheocharis:
This PR adds monomorphisation support to TC. Every `pure` function or an expression with `#run` will be evaluated at compile-time. In effect, generic calls are reduced to their monomorphised equivalents, appropriate for lowering.

This PR also:
- Adds the `#pure` directive for functions, to label them as pure without having to use the `<>` syntax.
- Adds the `#run` directive to run any expression at compile time
- Implements weak-head normalisation so that the unification can use it.

There are still a few things to do:
- Assign atom info to all monomorphised terms so that the lowering can actually process them
- Assign nice names to monomorphised functions

The monomorphisation can be turned off with `-Cmono=off` if needed (for debugging).

Closes #835

### DIFF
--- a/compiler/hash-untyped-semantics/src/visitor.rs
+++ b/compiler/hash-untyped-semantics/src/visitor.rs
@@ -65,7 +65,7 @@ impl<'s> SemanticAnalyser<'s> {
                 Expr::FnDef(_) => {}
                 _ => self.invalid_argument(DirectiveArgument::FnDef, directive, value.ast_ref()),
             }
-        } else {
+        } else if !matches!(subject.body(), Expr::FnDef(_)) {
             self.invalid_argument(DirectiveArgument::Declaration, directive, subject);
         }
     }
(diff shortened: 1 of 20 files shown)


compiler/hash-untyped-semantics/src/visitor.rs
```diff
@@ -218,7 +218,11 @@ impl<'s> SemanticAnalyser<'s> {
             // Check that the supplied argument to a function modifying directive
             // is a declaration of a function that the directive will apply to.
             self.validate_data_def_directive(directive, subject)
-        } else if !directive.is(IDENTS.dump_ast) && !directive.is(IDENTS.dump_tir) {
+        } else if !directive.is(IDENTS.dump_ast)
+            && !directive.is(IDENTS.dump_tir)
+            && !directive.is(IDENTS.run)
+            && !directive.is(IDENTS.pure)
```
feds01, line 224: Surely this should check that is set on a function definition, or is this checked later?

kontheocharis: Fixed

feds01 (approved):

PR #844, opened 2023-03-13
Simplify context bindings

kontheocharis:
This PR fixes some issues with unification, and unifies `StackMemberId`/`ArgId` bindings into `Decl`. This makes a lot of things simpler, and now members can be looked up just by symbol.

Some groundwork has been added to make inference and unification modifying rather than copying, which should greatly improve performance but also retain locations automatically.

### DIFF
--- a/compiler/hash-semantics/src/passes/discovery/defs.rs
+++ b/compiler/hash-semantics/src/passes/discovery/defs.rs
@@ -5,9 +5,9 @@ use hash_reporting::macros::panic_on_span;
 use hash_tir::{
     data::{CtorDefData, DataDefId},
     defs::DefId,
-    environment::env::AccessToEnv,
+    environment::{context::Decl, env::AccessToEnv},
     mods::{ModDefData, ModDefId, ModKind, ModMemberData, ModMemberValue},
-    scopes::{StackId, StackMemberData},
+    scopes::StackId,
     symbols::Symbol,
     tys::TyId,
     utils::{common::CommonUtils, AccessToUtils},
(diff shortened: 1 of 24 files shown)


compiler/hash-semantics/src/passes/discovery/defs.rs
```diff
@@ -204,7 +204,7 @@ impl<'tc> DiscoveryPass<'tc> {
                 self.def_state().stack_members.modify_fast(stack_id, |members| {
                     if let Some(members) = members {
                         let members = std::mem::take(members);
-                        let members_len = members.len();
+                        let _members_len = members.len();
```
feds01, line 207: remove?

compiler/hash-semantics/src/passes/discovery/defs.rs
```diff
@@ -438,11 +434,12 @@ impl<'tc> DiscoveryPass<'tc> {
...
                 buf.push((
                     node.id(),
-                    StackMemberData {
+                    Decl {
                         name: self.new_symbol(binding.name.ident),
-                        is_mutable: binding.mutability.as_ref().map(|m| *m.body())
-                            == Some(ast::Mutability::Mutable),
-                        ty: self.new_ty_hole(),
+                        // is_mutable: binding.mutability.as_ref().map(|m| *m.body())
+                        //     == Some(ast::Mutability::Mutable),
```
feds01, line 440: This is commented out?

stdlib/prelude.hash
```diff
@@ -28,6 +28,11 @@ transmute := <T,U> => (item: T) -> U => {
 
 #repr_c SizedPointer := struct(&raw u8, usize);
 
+panic := (value: str) => {
+  print(value);
+  Intrinsics::abort()
+}
+
```
feds01, line 35: this is already present in main, this can be remove here

compiler/hash-semantics/src/passes/discovery/defs.rs
kontheocharis: Yes i will add it back later

feds01 (approved):

PR #848, opened 2023-03-15
Refactor inference, add dependent pattern matching and definitions, optional entry point

### LINKED ISSUE #841: Dependent pattern matching

kontheocharis:
This PR:
- Modifies the way that inference and unification is done so that new terms don't have to be created all the time. This retains most locations automatically.
- Adds dependent pattern matching support (and equality identities in Prelude), as well as the ability to define inductive data types with indices to the parser.
- Fixes a lot of issues with the previous inference.
- The analysis only requires an entry point if the target stage is `> Lowering`.

Closes #841.

Unfortunately there are still a few places where locations aren't set due to copying, hopefully this can be done soon.

### DIFF
--- a/compiler/hash-parser/src/parser/definitions.rs
+++ b/compiler/hash-parser/src/parser/definitions.rs
@@ -69,7 +69,17 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
             self.consume_gen(gen);
         }
 
-        Ok(self.node_with_joined_span(EnumDefEntry { name, fields }, name_span))
+        // Attempt to parse an optional type for the variant
+        // Now try and parse a type if the next token permits it...
+        let ty = match self.parse_token_fast(TokenKind::Colon) {
+            Some(_) => match self.peek() {
+                Some(token) if matches!(token.kind, TokenKind::Eq | TokenKind::Comma) => None,
+                _ => Some(self.parse_ty()?),
+            },
+            None => None,
+        };
+
+        Ok(self.node_with_joined_span(EnumDefEntry { name, fields, ty }, name_span))
     }
 
     /// Parses an nominal definition type field, which could either be a named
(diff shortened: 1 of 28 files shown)


compiler/hash-parser/src/parser/definitions.rs
```diff
@@ -69,7 +69,17 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
             self.consume_gen(gen);
         }
 
-        Ok(self.node_with_joined_span(EnumDefEntry { name, fields }, name_span))
+        // Attempt to parse an optional type for the variant
+        // Now try and parse a type if the next token permits it...
+        let ty = match self.parse_token_fast(TokenKind::Colon) {
+            Some(_) => match self.peek() {
+                Some(token) if matches!(token.kind, TokenKind::Eq | TokenKind::Comma) => None,
```
feds01, line 76: what should happen on `:=`, this seems to ignore this case?

compiler/hash-typecheck/src/substitution.rs
```diff
@@ -323,7 +476,8 @@ impl<T: AccessToTypechecking> SubstitutionOps<'_, T> {
     /// Invariant: the arguments are ordered to match the
     /// parameters.
     pub fn create_sub_from_args_of_params(&self, args_id: ArgsId, params_id: ParamsId) -> Sub {
-        assert!(params_id.len() == args_id.len(), "called with mismatched args and params");
+        // assert!(params_id.len() == args_id.len(), "called with mismatched args and
+        // params");
```
feds01, line 480: commented out?

compiler/hash-typecheck/src/inference.rs
```diff
@@ -954,138 +935,88 @@ impl<T: AccessToTypechecking> InferenceOps<'_, T> {
...
-
-        // Atom is already registered but not inferred, it means we are in a
-        // recursive call.
-        if self.atom_is_registered(fn_def_id) && fn_mode == FnInferMode::Body {
-            return Ok(Inference(fn_def_id, fn_def.ty));
+        self.infer_fn_annotation_ty(fn_def_id, annotation_ty)?;
+        let fn_def = self.get_fn_def(fn_def_id);
+
+        if fn_mode == FnInferMode::Header {
+            // If we are only inferring the header, then also want to check for
```
feds01, line 950: typo

feds01 (approved):

PR #849, opened 2023-03-16
exe mode

feds01:
- compiler: add `exe` mode to enable running the built executable
- link+target: basic support for x86 Windows PC platform
- link: add timing for discovery and execution
- link: fix linking problems for MSVC linker

### DIFF
--- a/compiler/hash/src/main.rs
+++ b/compiler/hash/src/main.rs
@@ -86,15 +86,15 @@ fn main() {
 
     match entry_point {
         Some(path) => {
-            let new_state = compiler.run_on_filename(path, ModuleKind::EntryPoint, compiler_state);
-            let workspace = new_state.workspace();
-            let settings = new_state.settings();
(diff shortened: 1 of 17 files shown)


compiler/hash/src/main.rs
```diff
@@ -83,7 +86,25 @@ fn main() {
...
+                let path = workspace.executable_path(settings);
+
+                // We need to convert the path to a string so that we can pass it
+                // to the `Command` struct.
+                let path = path.to_str().unwrap();
+
+                // @@Todo: ideally, we should be able to parse the arguments that are specified
+                // after `--` into the spawned process.
+                let stream = Command::new(path).spawn().unwrap().wait_with_output().unwrap();
+                print!("{}", String::from_utf8_lossy(&stream.stdout));
```
kontheocharis, line 106: stderr?

kontheocharis: Also this prints it all at once.

kontheocharis (approved):

PR #858, opened 2023-03-24
Move out exhaustiveness, remove the old typechecker sources 

kontheocharis:
- Move out exhaustiveness, remove the old typechecker sources and disable TC tests
- Add a quick fix for tuple types not being monomorphised (need to investigate why this is happening)


compiler/hash-typecheck/src/inference.rs
```diff
@@ -383,6 +383,9 @@ impl<T: AccessToTypechecking> InferenceOps<'_, T> {
         let tuple_ty =
             self.new_expected_ty_of(original_term_id, self.new_ty(TupleTy { data: params }));
         self.check_by_unify(tuple_ty, annotation_ty)?;
+        // @@Review: why is this needed? Shouldn't the substitution be applied during
+        // `check_by_unify`?
+        self.sub_ops().apply_sub_to_atom_from_context(annotation_ty);
```
feds01, line 388: Does this application cause the large performance regression?

kontheocharis: Hm, it might actually now that you say that. Updated.

feds01 (approved):

PR #863, opened 2023-05-23
source: introduce `LineRanges` to represent line ranges of a module source

feds01:
A new data structure is introduced to represent a mapping from a range of bytes to lines in a source file. This is useful for converting `Span`s and `SourceLocation`s into exact line and column numbers by a simple lookup. This is intended to be used by the compiler when generating debug information during code generation, exact computing line and column numbers for errors, and figuring out line arrangements when formatting the code. 

Consequently, this led to a slight code cleanup of how information about modules was being stored in the `SourceMap`, now it is all grouped together under a `Module` (which includes lazily computed line ranges).

This also fixes a build issue with codegen LLVM attributes having mismatched IDs after bumping a version.


compiler/hash-utils/src/range_map.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        for (item, _) in self.store.iter() {
+            if overlaps(item, &key) {
+                panic!("keys are not allowed to overlap in a range map")
+            }
+
+            if key.start > item.end {
+                continue;
+            }
```
kontheocharis, line 139: why continue here?

kontheocharis: doesn't this mean that sometimes insert will silently fail?

feds01: thanks for the catch, fixed!

kontheocharis (approved):

PR #868, opened 2023-06-24
tc: Remove the usage of `Context` in `hash-semantics`

### LINKED ISSUE #867: Refactor the typechecking context system
Currently, the `Context` structure and its corresponding `ContextUtils` are being shared across `hash-semantics` and `hash-typecheck`, residing in `hash-tir`. Ideally, we would keep this context (albeit removing most `ScopeKind`s) for `hash-typecheck`, and separate out the `ScopeKind` stuff into a custom solution for `hash-semantics`

- [x] Duplicate `Context` inside `hash-semantics` and refactor it to fit the specific use case (remove unnecessary stuff)
- [x] Stop using `Context` from `hash-typecheck` inside semantics, replace with new `Context`.
- [x] Consider the separation between `hash-tir` and `hash-typecheck` in relation to `Context`.

kontheocharis:
This PR cleans up the currently widespread usage of `Context` throughout `hash-semantics`, `hash-typecheck` and `hash-lower`. Specifically, the `Scoping` module inside `hash-semantics` used to use `Context` internally to keep track of scope items. Now, `Scoping` keeps track of everything itself, not depending on `Context`. This means that `Context` no longer needs to know about different "binding kinds" other than declarations. `Scoping` deals with and resolves constructor names, data definition names, and module names so that by the time that TIR is generated, all these are resolved to their according terms, rather than remaining as `Var`s.

Closes #867


compiler/hash-semantics/src/passes/mod.rs
```diff
@@ -38,7 +38,7 @@ impl<'tc> Visitor<'tc> {
 
     /// Visits the source passed in as an argument to [Self::new_in_source]
     pub fn visit_source(&self) -> SemanticResult<()> {
-        self.context().clear_all();
+        // self.context().clear_all();
```
feds01, line 41: commented out

compiler/hash-semantics/src/passes/resolution/scoping.rs
```diff
@@ -155,29 +174,22 @@ impl<'tc> Scoping<'tc> {
...
+                b.pop();
+            },
+            f,
+        )
     }
 
     /// Add a new scope
-    pub(super) fn add_scope(&self, kind: ScopeKind, context_kind: ContextKind) {
-        self.context().add_scope(kind);
+    pub(super) fn add_scope(&self, _kind: ScopeKind, context_kind: ContextKind) {
```
feds01, line 191: remove the argument if it is unneeded.

feds01 (changes requested):

feds01 (approved):

feds01 (approved):

PR #873, opened 2023-06-25
Make stores multithreaded by RwLock and DashMap

kontheocharis:
This PR replaces usages of `RefCell` by parking lot's `RwLock` and usages of hash maps by `DashMap`. This is done so that the stores can be accessed by multiple threads, and more relevantly to be accessible from global static variables.

There is no observed performance difference before and after this change.


compiler/hash-tir/src/locations.rs
```diff
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, collections::HashMap};
+use std::{collections::HashMap, sync::RwLock};
```
kontheocharis, line 1: use parking lot here

feds01 (dismissed):

feds01 (approved):

PR #878, opened 2023-06-27
Introduce `hash_driver` as public facing compiler API

feds01:
- remove interactive shell sources and cleanup compiler session interface
- session: cleanup compiler session and compiler creation
- pipeline: rename `hash_session` to `hash_driver` and re-organise interface


compiler/hash-driver/src/driver.rs
```diff
@@ -0,0 +1,358 @@
...
+    settings::CompilerStageKind,
+};
+use hash_reporting::writer::ReportWriter;
+use hash_source::{ModuleKind, SourceId};
+use hash_utils::{log, stream_writeln, timing::timed};
+
+/// The Hash Compiler interface. This interface allows a caller to create a
+/// [Compiler] with the specified components. This allows external tinkerers
+/// to add their own implementations of each compiler stage with relative ease
+/// instead of having to scratch their heads.
```
kontheocharis, line 16: not clear how to do this yet?

kontheocharis (approved):

compiler/hash-driver/src/driver.rs
feds01: Yeah, this is an old comment, which should eventually be realised, should we keep this or no?

kontheocharis: Maube add a `@@Future` to it?

PR #882, opened 2023-06-29
Implement `fmt::Display` for TIR structures without requiring Env

### LINKED ISSUE #816: Consider turning some stores into statics rather than passing them around within contexts
Things to do:

- [x] Add a static STORES variable (#872)
- [x] Traits for accessing store data through IDs without a special context (#872)
- [x] Implement display for TIR IDs without a special context (#882)
- [ ] Remove `stores` from the TIR environment and solely use the static version. (will be done separately)

kontheocharis:
Pretty self explanatory. Closes #816. There is only one problem: when `ModDef`s are printed, ones which correspond to source files, the formatter needs access to the `SourceMap` in order to resolve the path from the `SourceId`. However, the map is not static, and since the `Display` implementations no longer have access to `Env`, this is a problem. @feds01 thoughts on how to deal with this? One option is to make the SourceMap static, not sure if it is a good idea or not.


feds01 (dismissed):

compiler/hash-layout/src/compute.rs
```diff
@@ -852,7 +852,8 @@ impl<'l> LayoutComputer<'l> {
 
             for (field_layouts, variant_layout) in field_layouts.iter().zip(variant_layouts) {
                 // All variant layouts must be a struct
-                let LayoutShape::Aggregate { fields: ref offsets, .. } = variant_layout.shape else {
+                let LayoutShape::Aggregate { fields: ref offsets, .. } = variant_layout.shape
+                else {
                     panic!("layout of enum variant is non-aggregate");
```
feds01, line 857: this is pretty weird formatting, is this from `rustfmt`?

compiler/hash-ast/src/tree.rs
```diff
@@ -53,7 +53,7 @@ impl AstVisitor for AstTreeGenerator {
             "entry",
             name.map(|t| TreeNode::branch("name", vec![t]))
                 .into_iter()
-                .chain(ty.map(|t| TreeNode::branch("type", vec![t])).into_iter())
+                .chain(ty.map(|t| TreeNode::branch("type", vec![t])))
```
feds01, line 56: I think that changes like this should be in their own PR in the future. If there is some PR unrelated fmt or clippy change, then I think the policy should be to make a separate PR for it and merge it before your main PR.

feds01:
In regards to clippy barfing, there seems to be some work being done to fix it (https://github.com/rust-lang/rust-clippy/pull/11098). For now, we can disable the `unnecessary_literal_unwrap` lint for `hash-codegen-llvm` until the issue is fixed. This should be done probably in a separate PR with the other fmt and clippy changes.

compiler/hash-ast/src/tree.rs
feds01: I've made #883 to fix all the clippy and fmt issues, maybe we should merge this first and then rebase?

y21:
👋 coming from clippy. Interesting ICE you found. Just wanted to say that the linked PR doesn't fix this, but I've created an issue with a small repro: https://github.com/rust-lang/rust-clippy/issues/11099

feds01:
Also, this PR doesn't update `tc_panic!` macros in `hash_semantics::panic`. Should they be removed or moved elsewhere since they aren't really being used?

kontheocharis:
I wonder if this issue is also relevant to the clippy crash https://github.com/rust-lang/rust-clippy/issues/11064

kontheocharis:
> Also, this PR doesn't update `tc_panic!` macros in `hash_semantics::panic`. Should they be removed or moved elsewhere since they aren't really being used?

Will tackle this in this PR and then we can merge?

compiler/hash-ast/src/tree.rs
kontheocharis: > I think that changes like this should be in their own PR in the future. If there is some PR unrelated fmt or clippy change, then I think the policy should be to make a separate PR for it and merge it before your main PR.

Yeah sounds reasonable but sometimes can be really annoying..

compiler/hash-layout/src/compute.rs
kontheocharis: Yes

feds01:
Rebase and we're good to go

feds01 (approved):

PR #886, opened 2023-07-08
Remove old println line in inference

kontheocharis:
Oops..


feds01 (approved):

feds01:
If only there could be some kind of hook to fail CI if this happens

PR #887, opened 2023-07-12
Re-add exhaustiveness into compiler pipeline

### LINKED ISSUE #729: Hook up exhaustiveness

feds01:
This PR updates exhaustiveness to use the new TIR representation when processing items, and re-integrates it into the typechecking infrastructure. Whilst there are some rough edges about how this is done, this is a good first step towards having exhaustiveness run. In the future, we want to move away from directly calling exhaustiveness, and more appropriately queue requests through a channel and let them be processed. However, this work deserves a PR of its own.

- exhaustiveness: re-add sources and update to new representation
- exhaustiveness: use new TIR API to read data types
- semantics: remove un-needed `Flags` struct
- exhaustiveness: re-implement `fmt::Debug` methods for various data types
- exhaustiveness: implement converting `ExhaustivenessError` and `ExhaustivenessWarning` into `Reports`
- tir: add `MatchOrigin` to `MatchTerm`
- semantics: allow for exhaustiveness errors and warnings to be added into semantic environment
- tc: hookup exhaustiveness to inference pass
- exhaustiveness: fix bug with float wildcards panicking
- exhaustiveness: fix bug where spread pats are attempted to be deconstructed
- tir: fix some visual discrepancies with data constructor and constant printing
- range_map+tc: remove `println!` calls
- semantics: return semantic warnings to pipeline
- tc: adjust printing of ctor patterns and check match exhaustiveness
- lower: fix bug where assert operands weren't being updated when eliminated
- tests: add back exhaustiveness tests

closes #729


kontheocharis:
Going off of what you said last time, I wonder if some of these unrelated changes can be made into separate PRs.

compiler/hash-exhaustiveness/src/lib.rs
```diff
@@ -1 +1,273 @@
...
+//! ```ignore
+//! x := Some(3); // ty: Option<i32>
+//! match x {
+//!     Some(_) => print("there is a number");
+//!     None => print("there is no number");
+//! };
+//! ```
+//!
+//! So in this example, for `x` which is of type `Option<i32>`, will the
+//! patterns: [`Some(_)`, `None`] cover all cases of `Option<i32>`. In this
```
kontheocharis, line 25: `?`

compiler/hash-exhaustiveness/src/lib.rs
```diff
@@ -1 +1,273 @@
...
+    /// primitive types with the TIR.
+    primitives: &'tc DefinedPrimitives,
+
+    /// The exhaustiveness store keeps track of created de-constructed patterns
+    /// and usefulness constructors, basically representations used by
+    /// exhaustiveness to evaluate whether something is exhaustive. Since
+    /// every exhaustive instance is independent from one another, we can keep
+    /// the [ExhaustivenessCtx] local to each checker, or in other words, local
+    /// to the thread that is processing the job.
+    ecx: ExhaustivenessCtx,
```
kontheocharis, line 120: Can exhaustiveness stores become thread local static?

compiler/hash-tir/src/pats.rs
```diff
@@ -33,32 +33,55 @@ pub struct Spread {
...
     /// If the range includes the `end` or not.
-    pub range_end: RangeEnd,
+    pub end: RangeEnd,
 }
 
 /// Represents a pattern.
 ///
 /// Check the documentation of each member for more information.
 #[derive(Copy, Clone, Debug, From)]
 pub enum Pat {
```
kontheocharis, line 56: Duplicate documentation with the member structs?

compiler/hash-typecheck/src/inference.rs
```diff
@@ -1471,6 +1506,14 @@ impl<T: AccessToTypechecking> InferenceOps<'_, T> {
         }
 
         self.check_by_unify(unified_ty, annotation_ty)?;
+
+        // @@Todo: If it hasn't, we can use/make a new ExhaustivenessChecker and then
```
kontheocharis, line 1510: If what hasn't?

kontheocharis:
Does this mean that enums work in exhaustiveness too?

feds01:
> Does this mean that enums work in exhaustiveness too?

Yeah, prelude has some enums in match cases, but I will also add some tests

compiler/hash-typecheck/src/inference.rs
feds01: I’ll remove

compiler/hash-exhaustiveness/src/lib.rs
feds01: They can but I don’t see any need to share them across several exhaustiveness instances. The idea is in the future when they run on their own threads, they are independent and don’t interfere with one another.

feds01:
> Going off of what you said last time, I wonder if some of these unrelated changes can be made into separate PRs.

Yeah I can cherrypick some things out. It was weird, something unrelated broke in lowering whilst I implemented exhaustiveness.

compiler/hash-exhaustiveness/src/lib.rs
kontheocharis: Yeah that's what I mean, thread locals. Just so that you don't need to carry them around in the context.

feds01: Ah yes, they can. Ideally, they should be cleared after TC so they don't take up excess memory. I guess this can be done in the `cleanup()` call in the `SemanticAnalysis`? Can I make the stores thread local by just using `thread_local!` and use the same technique as in TIR?

compiler/hash-tir/src/pats.rs
feds01: Which part do you mean?

compiler/hash-exhaustiveness/src/lib.rs
kontheocharis: Take a look at `hash_tir:: environment::stores`. Probably have to do something similar. A lot of it might be able to be factored out into `hash_utils`.

compiler/hash-tir/src/pats.rs
feds01: Indeed, there is a bit of duplication. I still think it's useful to summarise all of the content in one place, but it can be removed.

kontheocharis: Maybe just a shortened version? otherwise they will eventually start getting out of sync

compiler/hash-exhaustiveness/src/lib.rs
feds01: Decided that we'll do this later and migrate exhaustiveness when its ready.

kontheocharis (approved):

kontheocharis (approved):

PR #892, opened 2023-07-16
tir: Replace some functions in `common.rs` using `Env` by static functions using static stores

kontheocharis:
This is part of a greater effort to stop relying on `Stores` existing on `Env` and instead use the static stores.

This PR also uses the newly added `Symbol::fresh()` to create a positional parameter when a positional index is given to `make_param_name_from_arg_index` (which is now called `ParamIndex::into_symbol`)


compiler/hash-tir/src/params.rs
```diff
@@ -70,19 +70,12 @@ impl From<ParamId> for ParamIndex {
...
-                format!(" = {}", default)
-            } else {
-                "".to_string()
-            }
-        )
+impl ParamIndex {
+    pub fn into_symbol(&self) -> Symbol {
+        match self {
+            ParamIndex::Name(name) => Symbol::from_name(*name),
+            ParamIndex::Position(i) => Symbol::from_name(*i),
```
feds01, line 77: `Symbol::fresh()` instead?

feds01 (approved):

PR #894, opened 2023-07-17
support open ranges in patterns, i.e. `4..`, `..<5` `..`, etc

feds01:
- target: avoid awkward API for querying sizes/max/mins when handling bigints
- parser+ast: support open ranges, i.e. omit start and or end
- tir: specify `RangePat` `lo` and `hi` values as optionals
- exhaustiveness: support open pattern ranges
- tc: support open ranges in patterns
- lower: support open ranges in lowering
- tests: add tests for open ranges in patterns and remove outdated tests

- [x] #893 should be merged first.

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -876,9 +876,11 @@ define_tree! {
     #[node]
     pub struct RangePat {
         /// Initial bound of the range
-        pub lo: Child!(Lit),
+        pub lo: OptionalChild!(Lit),
+
         /// Upper bound of the range
-        pub hi: Child!(Lit),
+        pub hi: OptionalChild!(Lit),
+
         /// Whether the `end` is included or not
         pub end: RangeEnd,
     }
--- a/compiler/hash-tir/src/pats.rs
+++ b/compiler/hash-tir/src/pats.rs
@@ -40,10 +40,10 @@ pub struct Spread {
 #[derive(Copy, Clone, Debug)]
 pub struct RangePat {
     /// The beginning of the range.
-    pub lo: LitPat,
+    pub lo: Option<LitPat>,
 
     /// The end of the range.
-    pub hi: LitPat,
+    pub hi: Option<LitPat>,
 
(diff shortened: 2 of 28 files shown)


compiler/hash-ast/src/ast.rs
```diff
@@ -876,9 +876,11 @@ define_tree! {
     #[node]
     pub struct RangePat {
```
kontheocharis, line 877: Maybe add a comment "Invariant : must have at least one of lo, hi"

feds01: But you don't actually need this

kontheocharis: What's the point of omitting both?

feds01: Well it is the same as just specifying a wildcard, but I don’t see much point in disallowing it.

compiler/hash-tir/src/pats.rs
```diff
@@ -40,10 +40,10 @@ pub struct Spread {
 #[derive(Copy, Clone, Debug)]
 pub struct RangePat {
     /// The beginning of the range.
-    pub lo: LitPat,
+    pub lo: Option<LitPat>,
```
kontheocharis, line 43: Why is this an option, why not inline min/max during resolution?

feds01: Yeah I thought this would be the best thing to do, it would avoid quite alot of code duplication. I just wasn’t sure if it made sense to do it in resolution. Ideally it should happen when the type of each operand is known. This would avoid the changes in lowering and exhaustiveness. What do you think?

kontheocharis: Since the type should be known it should happen during inference as you said. However is there any time the type will not be known because it will be a generic?

feds01: Well, for now it is only possible for the range to be composed of numeric literals, hence inly `char` and sized `{i|u}xxx` types are supported.

PR #897, opened 2023-07-17
ir+lower: remove old TIR `stores()` API and migrate to new methods

feds01:


compiler/hash-lower/src/build/into.rs
```diff
@@ -193,16 +192,13 @@ impl<'tcx> BodyBuilder<'tcx> {
...
-                    FnCallTermKind::LogicalBinOp(op, _, _) => {
+                    FnCallTermKind::LogicalBinOp(op, lhs_term, rhs_term) => {
                         let (short_circuiting_block, mut else_block, join_block) = (
                             self.control_flow_graph.start_new_block(),
                             self.control_flow_graph.start_new_block(),
                             self.control_flow_graph.start_new_block(),
                         );
 
-                        let lhs_term = self.stores().args().get_at_index(*args, 1).value;
-                        let rhs_term = self.stores().args().get_at_index(*args, 2).value;
```
feds01, line 204: Might be responsible for different IR gen?

compiler/hash-lower/src/build/rvalue.rs
```diff
@@ -23,41 +23,39 @@ use crate::build::category::RValueKind;
 
 impl<'tcx> BodyBuilder<'tcx> {
     /// Construct an [RValue] from the given [ast::Expr].
-    pub(crate) fn as_rvalue(&mut self, mut block: BasicBlock, term_id: TermId) -> BlockAnd<RValue> {
-        // @@Temporary: replace with get_ref();
-        let term = self.stores().term().get(term_id);
-        let span = self.span_of_term(term_id);
+    pub(crate) fn as_rvalue(&mut self, mut block: BasicBlock, term: TermId) -> BlockAnd<RValue> {
```
feds01, line 26: rename `term` to `rvalue` for clarity

compiler/hash-lower/src/build/into.rs
kontheocharis: Not sure what you mean

feds01: Im investigating why ir changed, its a self comment

feds01: Yeah, this is actually a bug in lowering that I accidentally discovered, it should be that I need to get the `2` and `3` arguments.

tests/cases/lowering/early_returns.stdout
```diff
@@ -9,9 +9,10 @@ boo := (_1: i32, _2: i32) -> i32 {
     mut _3: bool;
     mut _4: bool;
     mut _5: bool;
-    mut _6: i32;
+    mut _6: bool;
```
kontheocharis, line 12: I'm guessing this is the difference?

kontheocharis (approved):

PR #903, opened 2023-07-18
Move out static store implementation from TIR to utils

### LINKED ISSUE #900: utils: Factor out "static" store logic to utils so that other parts of the compiler can use it
The TIR currently does its own magic to have access to static stores and access values through IDs. This can be factored out into `hash-utils` so that more parts of the compiler can use static stores.

feds01:
- tir: parameterise store macros with `store_source`
- tir: rename `global_stores()` to `tir_stores()`
- tir: make `stores!` macro use a parameterised name
- utils: remove `append-only-store` variant
- utils(store): split up various stores into their own modules
- tir: rename `tir_sequence_*` macros to `static_sequence_*`
- utils(store): move static store implementation into `hash_utils::store`

Whilst we're at it, we could move everything here into `hash-storage` or rename `hash-alloc` into `hash-storage` and move things there?

closes #900


kontheocharis (dismissed):
GG, looks great. I think moving to alloc and renaming to storage is a good idea.

feds01:
> GG, looks great. I think moving to alloc and renaming to storage is a good idea.

should I do this in the current PR?

kontheocharis:
Yes I think so

feds01:
> Yes I think so
Done!

kontheocharis (approved):

PR #905, opened 2023-07-19
Make layouts static

feds01:
And some code cleanup with it.

### DIFF
--- a/compiler/hash-codegen/src/lower/intrinsics.rs
+++ b/compiler/hash-codegen/src/lower/intrinsics.rs
@@ -36,7 +36,7 @@ impl<'a, 'b, Builder: BlockBuilderMethods<'a, 'b>> FnBuilder<'a, 'b, Builder> {
         args: &[OperandRef<Builder::Value>],
         result: Builder::Value,
     ) {
-        let result = PlaceRef::new(builder, result, ret_abi.info);
+        let result = PlaceRef::new(result, ret_abi.info);
 
         let value = match intrinsic {
             Intrinsic::Abort => {
@@ -114,7 +114,7 @@ impl<'a, 'b, Builder: BlockBuilderMethods<'a, 'b>> FnBuilder<'a, 'b, Builder> {
     ) {
         let src = self.codegen_operand(builder, src);
 
-        self.ctx.layouts().map_many_fast([src.info.layout, dest.info.layout], |layouts| {
+        self.ctx.layouts().store().map_many_fast([src.info.layout, dest.info.layout], |layouts| {
             let (src_layout, dest_layout) = (layouts[0], layouts[1]);
 
             // Special case for transmuting between pointers and integers.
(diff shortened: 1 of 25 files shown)


compiler/hash-codegen/src/lower/intrinsics.rs
```diff
@@ -114,7 +114,7 @@ impl<'a, 'b, Builder: BlockBuilderMethods<'a, 'b>> FnBuilder<'a, 'b, Builder> {
     ) {
         let src = self.codegen_operand(builder, src);
 
-        self.ctx.layouts().map_many_fast([src.info.layout, dest.info.layout], |layouts| {
+        self.ctx.layouts().store().map_many_fast([src.info.layout, dest.info.layout], |layouts| {
```
kontheocharis, line 117: Is it needed to still have `.store()` on layouts? You can just use `layout_stores()` right?

compiler/hash-layout/src/lib.rs
```diff
@@ -56,18 +62,25 @@ pub struct PointeeInfo {
...
+    layouts: LayoutStore
+);
+
+/// The global [`LayoutStores`] instance.
+static STORES: OnceLock<LayoutStores> = OnceLock::new();
+
+/// Access the global [`LayoutStores`] instance.
+pub(crate) fn layout_store() -> &'static LayoutStores {
+    STORES.get_or_init(LayoutStores::new)
+}
```
kontheocharis, line 76: Might as well suffix with `.layouts()` since that's the only store?

compiler/hash-layout/src/lib.rs
```diff
@@ -568,3 +573,36 @@ pub enum Variants {
...
+    /// Compute the [Size] of a given [LayoutId].
+    pub fn size(&self) -> Size {
+        self.borrow().size
+    }
+
+    /// Compute the [Alignments] of a given [LayoutId].
+    pub fn alignments(&self) -> Alignments {
+        self.borrow().alignment
+    }
+}
```
kontheocharis, line 608: If only rust had class inheritance...

feds01: Indeed

compiler/hash-layout/src/lib.rs
feds01: Do you mean to just call it `Layouts`?

kontheocharis, line 76:
```suggestion
pub(crate) fn layout_store() -> &'static LayoutStore {
    STORES.get_or_init(LayoutStores::new).layouts()
}
```

kontheocharis: See suggestion below

feds01: Oh, but you cannot do this since the `static_single_store` requires it to return `LayoutStores` so that it can index into layouts.

kontheocharis: Oh yeah you're right.. I guess it is fine then. It is just potentially annoying to have to do `layout_store().layouts()` all the time

feds01: Yeah its only in one place anyway so its not too bad.

kontheocharis (approved):

PR #907, opened 2023-07-21
spanless ast

feds01:
- ast: Make AST spanless, and store `Span`s within `SPAN_MAP`
- ast: Use `RwLock` and `once_cell:sync::Lazy` for SpanMap

### DIFF
--- a/compiler/hash-ast/src/ast.rs
+++ b/compiler/hash-ast/src/ast.rs
@@ -5,6 +5,7 @@ use std::{
     hash::Hash,
     ops::{Deref, DerefMut},
     path::PathBuf,
+    sync::OnceLock,
 };
 
 use hash_source::{
@@ -13,7 +14,11 @@ use hash_source::{
     location::Span,
 };
 use hash_tree_def::define_tree;
-use hash_utils::{counter, smallvec::SmallVec};
+use hash_utils::{
+    counter,
+    index_vec::{Idx, IndexVec},
+    smallvec::SmallVec,
+};
 use replace_with::replace_with_or_abort;
 
 counter! {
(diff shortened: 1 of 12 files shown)


compiler/hash-ast/src/ast.rs
```diff
@@ -23,15 +29,53 @@ counter! {
...
+    fn add_span(span: Span) -> AstNodeId {
+        SPAN_MAP.write().push(span)
+    }
+
+    /// Update the span of a node by [AstNodeId].
+    fn update_span(id: AstNodeId, span: Span) {
+        SPAN_MAP.write()[id] = span;
+    }
+}
+
```
kontheocharis, line 71: Why not use the partial_store macro for this?

kontheocharis (dismissed):

compiler/hash-ast/src/ast.rs
feds01: Well its not a partial store, but the single store could be used here. I also wanted to make AstNodeId specifically 4bytes and not 8.

compiler/hash-ast/src/ast.rs
```diff
@@ -13,7 +13,13 @@ use hash_source::{
...
+use hash_utils::{
+    counter,
+    index_vec::{Idx, IndexVec},
+    parking_lot::RwLock,
+    smallvec::SmallVec,
+};
+use once_cell::sync::Lazy;
 use replace_with::replace_with_or_abort;
 
 counter! {
```
kontheocharis, line 25: Since you are using indexvec then does it make sense to create this ID through https://doc.rust-lang.org/beta/nightly-rustc/rustc_index/macro.newtype_index.html?

compiler/hash-ast/src/ast.rs
kontheocharis: Right cause its a total mapping. See comment below on the ID.

compiler/hash-ast/src/ast.rs
feds01: Yeah I was thinking about doing this, probably don’t need the `counter!` anymore.

kontheocharis: Will this retain the invariant of increasing `AstNodeIds` across the AST for ordering error messages? Also maybe that should be in a comment somewhere here.

feds01: Yes, it will because only `AstNode::new()` can create a new AstNodeId and if they are made in a particular order, then that order should be preserved (as shown by the tests passing which rely on this behaviour to be true).

kontheocharis (approved):

PR #909, opened 2023-07-21
core: Use `SourceLocation` (new name `Span`) over `Span` (new name `ByteRange`) to refer to locations

feds01:
- ast: Make AST spanless, and store `Span`s within `SPAN_MAP`
- ast: Use `RwLock` and `once_cell:sync::Lazy` for SpanMap
- lower: fix panic when trying to access numbered ADT fields
- ast: Add `AstNodeId` to `AstNodes`, guaranteeing node lists have spans
- ast+source: use `SourceLocation` over `Span`
- ir+lower: switch to using `AstNodeId`s for spans instead of directly using spans
- ir: add static size assertions
- source: rename `RowColSpan` to `RowColRange`
- source: rename `Span` to `ByteRange`
- source: rename `SourceLocation` to `Span`
- tir: allow `AstInfo` to have multiple items refer to the same `AstNodeId`

Todo:
- [x] review & merge #907

### DIFF
--- a/compiler/hash-ir/src/ir.rs
+++ b/compiler/hash-ir/src/ir.rs
@@ -1524,7 +1524,6 @@ mod tests {
     }
 }
 
-
 #[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
 mod size_asserts {
     use hash_utils::assert::static_assert_size;
--- a/compiler/hash-ir/src/ir.rs
+++ b/compiler/hash-ir/src/ir.rs
@@ -1523,3 +1523,15 @@ mod tests {
         assert_eq!(format!("{}", place), "(*(*(*_0)))");
     }
 }
+
+
+#[cfg(all(target_arch = "x86_64", target_pointer_width = "64"))]
+mod size_asserts {
+    use hash_utils::assert::static_assert_size;
+
+    use super::*;
+
+    static_assert_size!(Statement, 64);
+    static_assert_size!(Terminator, 104);
+    static_assert_size!(RValue, 40);
+}
--- a/compiler/hash-ir/src/ir.rs
+++ b/compiler/hash-ir/src/ir.rs
@@ -840,7 +840,7 @@ pub struct Statement {
     /// The location of the statement. This is mostly used for error reporting
     /// or generating debug information at later stages of lowering
     /// beyond the IR.
-    pub span: AstNodeId,
+    pub origin: AstNodeId,
 }
 
 /// The kind of assert terminator that it is.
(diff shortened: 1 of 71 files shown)


compiler/hash-ir/src/ir.rs
```diff
@@ -836,11 +837,10 @@ pub struct Statement {
...
     pub kind: StatementKind,
 
-    /// The [Span] of the statement, relative to the [Body]
-    /// `source-id`. This is mostly used for error reporting or
-    /// generating debug information at later stages of lowering
+    /// The location of the statement. This is mostly used for error reporting
+    /// or generating debug information at later stages of lowering
     /// beyond the IR.
-    pub span: Span,
+    pub span: AstNodeId,
```
kontheocharis, line 843: Hmm, should these kinds of fields now be called `origin`?

compiler/hash-lexer/src/lib.rs
```diff
@@ -964,7 +963,7 @@ impl<'a> Lexer<'a> {
             return self.emit_fatal_error(
                 None,
                 LexerErrorKind::UnclosedStringLit,
-                Span::new(start, self.offset.get()),
+                ByteRange::new(start, self.offset.get()),
```
kontheocharis, line 966: Is this truly a byte range or a codepoint range?

compiler/hash-ir/src/ir.rs
feds01: yeah this is probably a better name for the fields.

compiler/hash-lexer/src/lib.rs
feds01: It is kind of both.

kontheocharis: Maybe it is better to call it `CharacterRange` or something like that?

feds01: hmmm but it does specifically refer to bytes, the `RowColRange` is more of a character range.

compiler/hash-lexer/src/lib.rs
```diff
@@ -306,7 +305,7 @@ impl<'a> Lexer<'a> {
             return None;
         }
 
-        let location = Span::new(offset, self.len_consumed());
+        let location = ByteRange::new(offset, self.len_consumed());
```
kontheocharis, line 308: This counts characters no?

feds01: Bytes:
```rs
    fn len_consumed(&self) -> usize {
        self.offset.get()
    }
```

feds01: However, it shouldn't ever be a value that is invalid to index into (in terms of bytes). So you shouldn't be able to end up in the middle of a unicode codepoint.

kontheocharis: Okay I see. Then this is fine

kontheocharis (approved):

PR #910, opened 2023-07-22
Speedup lowering by 20-25%

feds01:
- lower: avoid `.value()` on `FnDef`
- lower: optimise type lowering to avoid needlessly creating `()` which removes creating a lot of unneeded ADTs for just units. In some extreme cases, this led to a speedup of 20-25% and significantly less memory usage (in order of magnitude of MB).


kontheocharis:
Wow nice

kontheocharis (approved):

feds01:
> Wow nice

the power of profiling lol

PR #912, opened 2023-07-24
tir: Remove remaining usages of non-static `stores()`

### LINKED ISSUE #899: tc: Add `tir_partial_store` macro for partial stores in TIR

### LINKED ISSUE #911: tc: Fully migrate to static stores

kontheocharis:
This PR fully migrates `hash-tir`, `hash-typecheck`, `hash-intrinsics` and `hash-semantics` to the new static-based store API. As a result most `CommonUtils` have been factored out and placed into appropriate `impl` blocks for various TIR constructs. This is also the case for `ParamUtils`, `DataUtils` etc.

This PR also adds a static version of `PartialStore`, though this is not currently used.

Closes #899, closes #911


compiler/hash-exhaustiveness/src/diagnostics.rs
```diff
@@ -89,7 +89,7 @@ impl ExhaustivenessError {
                     .code(HashErrorCode::RefutablePat)
                     .title(format!("refutable pattern in {origin} binding: {pats} not covered"))
                     .add_labelled_span(
-                        env.get_location(pat).unwrap(),
+                        get_location(pat).unwrap(),
```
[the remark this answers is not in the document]
kontheocharis: Yes. I am also planning on wrapping them in a generic struct instead of hardcoding the fields `id, span` for each one. (Like with the AST)

compiler/hash-lower/src/lower_ty.rs
```diff
@@ -65,22 +66,21 @@ impl<'ir> BuilderCtx<'ir> {
...
-                    Ty::Data(data_ty) => self.ty_from_tir_data(*data_ty),
-
-                    // Hot path for unit types.
-                    Ty::Tuple(tuple) if tuple.data.is_empty() => COMMON_IR_TYS.unit,
-                    _ => self.uncached_ty_from_tir_ty(id, ty),
-                };
-
-                (result, false)
-            })
+            let ty = id.borrow();
```
feds01, line 69: I meant to just do an `id.map(...)` here

feds01 (changes requested):

compiler/hash-lower/src/lower_ty.rs
kontheocharis: It's the same thing

feds01: But surely the deref does the same thing as `value()` right?

kontheocharis: &* is a syntactic no op, it just casts from ref to &

feds01 (approved):

PR #914, opened 2023-07-25
docs: add entry about new macro syntax

feds01:


docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
+# Macros
+ 
+This section describes the syntax for macros in Hash. Macros are a way to write code that writes other code. There are two kind of macro invocations: one macro works on AST items, and the other works on 
+tokens.
+
+## AST macros 
+
+AST level macros are written with the syntax `#marco_name <subject>`
```
kontheocharis, line 8: marco

kontheocharis, line 8: AST-level

docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
...
+or `#[macro_name] <subject>`. The first form is a used as a shorthand
```
kontheocharis, line 9: Maybe add some args on that for clarity?

docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
...
+
+For example, the `#dump_ast` macro will accept any AST item as the subject and print the parsed AST to the console. 
+    
+```rust
+#dump_ast main := () => {
+    println("Hello, world!");
+}
+```
+
+An example of the AST macro being used to set some attributes on a function:
```
kontheocharis, line 20: an AST macro

docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
...
+```rust
+#[attr(foreign(c), no_mangle, link_name = "jpeg_read_header")]
+jpeg_read_header := (&raw cinfo, bool require_image) -> i32;
+```
+
+## Token macros
+
+Token macros follow a similar syntax to AST macros, but instead of working on AST items, they work on tokens. The syntax for token macros is `@macro_name <subject>` or `@[macro_name] <subject>`. The first form is a used as a shorthand for token macros that have no arguments. However, one significant difference between token macros and AST macros is that the token macro only accepts a token tree
```
kontheocharis, line 28: again maybe add some args

docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
...
+Formally, the macro syntax invocation can be written as follows:
+
+```bnf
+token_macro_invocation ::= "@" { macro_name | macro_args } token_tree;
+
+token_tree ::= "{" any "}" 
+             | "[" any "]" 
+             | "(" any ")";
+
+ast_macro_invocation ::= '#' (macro_name | macro_args ) macro_subject;
```
kontheocharis, line 127: Decide on grammar separators either () or {}

docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
...
+                  | param 
+                  | arg 
+                  | type_arg
+                  | pat_arg
+                  | match_case 
+                  | enum_variant;  
+
+macro_args ::= "[" { ∅ | macro_invocation ("," macro_invocation)* ","? } "]";
+
+macro_invocation ::= macro_name ( "("  ∅ | expr ("," expr )* ","?  ")" )?;
```
kontheocharis, line 143: Maybe just write args here? Cause named arguments..

docs/src/features/macros.md
```diff
@@ -0,0 +1,147 @@
...
+
+macro_name ::= ident;
```
kontheocharis, line 145: No namespacing?

kontheocharis (changes requested):

docs/src/features/macros.md
feds01: I was thinking about that, and I suppose ast path is better here.

docs/src/features/macros.md
feds01: Yeah, do they follow just the function argument rules?

docs/src/features/macros.md
feds01: ?

kontheocharis: There is a typo

docs/src/features/macros.md
kontheocharis: I think that's probably the simplest thing to do

docs/src/features/macros.md
feds01: Woah I didn't even see that.

kontheocharis (approved):

PR #921, opened 2023-08-21
parser: support parsing new AST macro syntax, and temporarily remove directive handling code

### LINKED ISSUE #916: Implement parsing AST-level macros `#foo` and `#[foo]`

feds01:
- ast: add support for macro invocations in various nodes
- parser+desugar: amend parser to new AST changes
- expand: disable directive checking (for now)
- fmt: implement pretty printing for new macro invocations
- tir+ir: remove handling of directives from TIR
- parser: support parsing macros and macro arguments
- parser: improve error message for encountering an unexpected keyword
- storage: don't re-export `fxhash::FxHashMap` and replace it with `hash_utils::fxhash`
- parser: fix parsing bug with newly introduced directives, and fixup tests

closes #916


compiler/hash-ast/src/ast.rs
```diff
@@ -401,6 +402,111 @@ define_tree! {
         }
     }
 
+    /// Function call argument.
```
kontheocharis, line 405: Wrong doc

compiler/hash-intrinsics/src/intrinsics.rs
```diff
@@ -749,10 +746,12 @@ impl DefinedIntrinsics {
                 "print_fn_directives",
                 FnTy::builder().params(params).return_ty(ret).build(),
                 |_, args| {
-                    if let Term::FnRef(fn_def_id) = args[1].value() {
-                        let directives =
-                            tir_stores().directives().get(fn_def_id.into()).unwrap_or_default();
-                        stream_less_writeln!("{:?}", directives.directives);
+                    if let Term::FnRef(_fn_def_id) = args[1].value() {
+                        todo!()
```
kontheocharis, line 750: left as Todo

compiler/hash-parser/src/parser/macros.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+            MacroKind::Ast if self.parse_token_fast(TokenKind::Pound).is_some() => {
+                self.skip_token();
+                Some(())
+            }
+            _ => None,
+        }
+    }
+
+    fn parse_macro_arg(&mut self) -> ParseResult<AstNode<MacroInvocationArg>> {
```
kontheocharis, line 28: what's the point of this being separate from function call args

compiler/hash-intrinsics/src/intrinsics.rs
feds01: Yeah, this will be re-implemented once attributes are registered on respective nodes, and the terms have associated `AstNodeId`s

compiler/hash-parser/src/parser/macros.rs
feds01: If the two are to be merged, then we should probably rename `ConstructorCallArg` to something more general. Any ideas?

kontheocharis: Maybe just Arg? It can even be unified with TyArg

feds01: Actually, `ConstructorCallArg` is defined as https://github.com/hash-org/hashc/blob/7a4f43d1c3d695d62e2e831699edd4b48ebb0dbd/compiler/hash-ast/src/ast.rs#L1864-L1873 which contains a macro invocation in the `ConstructorCallArg` itself whilst `MacroInvocationArg` does not. I guess the distinction is that a macro arg is not allowed to have inner args? But, we could still use the same variant... I don't know which is better.

compiler/hash-intrinsics/src/intrinsics.rs
feds01: I'll add a `@@ReAddDirectives` note so that this can be fixed when attributes are finally re-added.

compiler/hash-parser/src/parser/macros.rs
feds01: Also, does this mean that `TyArg` needs `macro_args` as well?

kontheocharis: Maybe there can be a EnrichedArg which is an Arg with an optional macro. And the former is what is used everywhere other than macros

kontheocharis: And yes, which is another reason to make this change

feds01: However, `TyArg` has a value as `Ty` not `Expr` so it isn't possible to unify them like that.

feds01: I do think that `ConstructorCallArg` should be renamed to `Arg` tho.

kontheocharis: Indeed, though it could trivially be wrapped in Expr::Ty. It is a cost of an extra ast node though, so perhaps settle for ExprArg/TyArg

feds01: This means that we still need a `MacroArg` variant.

kontheocharis: Looks good, though I think `s/macro_args/macros/`

compiler/hash-ast/src/ast.rs
```diff
@@ -401,6 +402,111 @@ define_tree! {
...
+    }
+
+    /// A macro invocation on an expression.
+    #[derive(PartialEq, Debug, Clone)]
+    #[node]
+    pub struct ExprMacroInvocation {
+        /// The directives that apply on the subject expression.
+        pub macro_args: Child!(MacroInvocations),
+
+        /// The subjection expression of the invocation.
```
kontheocharis, line 472: Shouldn't this be either an expression or tokens, depending on the kind?

compiler/hash-intrinsics/src/intrinsics.rs
kontheocharis: Okay, but maybe remove the todo!() so that it doesn't panic?

compiler/hash-ast/src/ast.rs
feds01: Yeah, I'd like to make this change after the original change, consider some additional constrains about ast/token macros, and then write them down.

kontheocharis (approved):

PR #923, opened 2023-08-22
TC: factor out parameter/argument validation into `hash-tir`

### LINKED ISSUE #922: Move out param/args checking logic from `hash-typecheck` to `hash-tir` so that it can be reused.

feds01:
- reporting: re-export `hash-error-codes` and update references across compiler
- tc+tir: factor out parameter/argument validation into `hash-tir` from `hash-typecheck`

closes #922 

## Todos:
- Possibly refactor `TcErrorState` to just use the `hash_reporting::diagnostics::ErrorState`

### DIFF
--- a/compiler/hash-reporting/src/diagnostic.rs
+++ b/compiler/hash-reporting/src/diagnostic.rs
@@ -298,3 +298,75 @@ pub trait AccessToDiagnosticsMut {
         self.diagnostics().merge_diagnostics(other)
     }
 }
+
+/// A convinient trait which specifies that an error type can be converted
+/// into a compound error version.
+pub trait IntoCompound: Sized {
+    fn into_compound(items: Vec<Self>) -> Self;
+}
+
+/// Accumulates errors that occur during typechecking in a local scope.
+///
+/// This is used for error recovery, so that multiple errors can be reported
+/// at once.
+#[derive(Debug)]
+pub struct ErrorState<E> {
+    pub errors: Vec<E>,
+}
+
+impl<E: IntoCompound> ErrorState<E> {
+    /// Create a new [ErrorState].
+    pub fn new() -> Self {
+        Self { errors: vec![] }
+    }
+
+    /// Add an error to the error state.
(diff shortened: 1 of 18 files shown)


compiler/hash-reporting/src/diagnostic.rs
```diff
@@ -298,3 +298,75 @@ pub trait AccessToDiagnosticsMut {
...
+    }
+
+    /// Take the errors from the error state.
+    pub fn take_errors(&mut self) -> Vec<E> {
+        std::mem::take(&mut self.errors)
+    }
+
+    /// Convert the accumulated [ErrorState] into a single error. This is
+    /// possible since [ErrorState] implements [IntoCompound].
+    pub fn into_error<T>(&mut self, t: impl FnOnce() -> Result<T, E>) -> Result<T, E> {
```
kontheocharis, line 358: Would it be better to factor this out into its own impl block with the IntoCompound bound?

compiler/hash-tir/src/utils/params.rs
```diff
@@ -1,42 +1,238 @@
...
+        ParamError::Compound { errors }
+    }
+}
+
+pub type ParamResult<T> = Result<T, ParamError>;
+
+/// Operations related to module definitions.
+#[derive(Constructor)]
+pub struct ParamUtils<'tc> {
+    env: &'tc Env<'tc>,
```
kontheocharis, line 223: Is env needed here?

compiler/hash-reporting/src/diagnostic.rs
feds01: Yeah, this is probably better.

compiler/hash-tir/src/utils/params.rs
feds01: I think not, I'll remove

feds01:
Also, what do you think about replacing `TcErrorState` with just an `ErrorState`

kontheocharis:
I think it should be fine to do that

kontheocharis (approved):

PR #930, opened 2023-08-25
attributes: implement checking and handling of attributes

feds01:
- attrs: introduce new crate and add Attribute representations
- attrs: add `AttrTyMap`
- ast: add initial code for running macro/attribute checks in `hash-expand`
- core: more dependency cleanups
- parser: don't include macros in parent span of macro-able items
- tir: use `From<...>` for implementing conversion to Lit kinds
- ir: use `attr_store()` for attributes and add `Instance::has_attr`
- core: rename `CompilerStageKind::DeSugar` to `CompilerStageKind::Expand`
- untyped-semantics: cleanup old code, and remove commented out directive checking code
- tir: register more `AstNodeId`s to items, specifically sequences of ids
- hash-tir: improve `ParamError` diagnostic messages
- expand: implement attribute checking
- expand: correctly apply attrs on attribute subject
- attrs: implement attribute specific error/warning checking and hookup to expand
- tir: move `primitives.rs` into `hash-tir` from `hash-intrinsics
- tir: introduce `HasNodeId` and re-implement directive handling
- parser: fix improper span on grouped macro invocations
- target: introduce `HasTarget` trait to query current `target()` from any implementor
- lower: re-implement `#layout_of`
- lower: implement applying `repr` and add `discriminant` option
- lower: re-implement `#dump_ir`


compiler/hash-attrs/src/target.rs
```diff
@@ -0,0 +1,292 @@
...
+            ast::Expr::Block(ast::BlockExpr { data: block }) => match block.body() {
+                ast::Block::Loop(_) | ast::Block::While(_) | ast::Block::For(_) => AttrTarget::Loop,
+                ast::Block::Match(_) | ast::Block::If(_) => AttrTarget::Match,
+                ast::Block::Body(_) => AttrTarget::Block,
+            },
+            ast::Expr::Import(_) => AttrTarget::Import,
+            ast::Expr::StructDef(_) => AttrTarget::StructDef,
+            ast::Expr::EnumDef(_) => AttrTarget::EnumDef,
+            ast::Expr::TyFnDef(_) => AttrTarget::TyFnDef,
+            ast::Expr::TraitDef(_) => AttrTarget::Item,
```
kontheocharis, line 121: Should other defs also be items?

compiler/hash-attrs/src/target.rs
```diff
@@ -0,0 +1,292 @@
...
+            ast::Expr::ImplDef(_) => AttrTarget::ImplDef,
+            ast::Expr::ModDef(_) => AttrTarget::ModDef,
+            ast::Expr::FnDef(_) => AttrTarget::FnDef,
+            ast::Expr::Ty(_) => AttrTarget::Ty,
+
+            // If this is a declaration, we have to recurse into the subject...
+            ast::Expr::Declaration(ast::Declaration { value: Some(value), .. }) => {
+                AttrTarget::classify_expr(value.body())
```
kontheocharis, line 129: What if someone wants to put an attribute on the declaration itself?

compiler/hash-attrs/src/target.rs
```diff
@@ -0,0 +1,292 @@
...
+}
+
+impl fmt::Display for AttrTarget {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let item_count = self.bits().count_ones();
+
+        // We will collect all of the allowed argument kinds into a vector
+        let mut allowed_argument_kinds = Vec::with_capacity(item_count as usize);
+
+        for kind in (*self).iter() {
```
kontheocharis, line 143: Map?

compiler/hash-attrs/src/target.rs
```diff
@@ -0,0 +1,292 @@
...
+                AttrTarget::Loop => allowed_argument_kinds.push("loop block"),
+                AttrTarget::Match => allowed_argument_kinds.push("match block"),
+                AttrTarget::ImplDef => allowed_argument_kinds.push("impl block"),
+                AttrTarget::Mod => allowed_argument_kinds.push("module"),
+                AttrTarget::ModDef => allowed_argument_kinds.push("mod block"),
+                AttrTarget::Block => allowed_argument_kinds.push("body block"),
+                AttrTarget::Import => allowed_argument_kinds.push("import"),
+                AttrTarget::StructDef => allowed_argument_kinds.push("`struct` definition"),
+                AttrTarget::EnumDef => allowed_argument_kinds.push("`enum` definition"),
+                AttrTarget::TyFnDef => allowed_argument_kinds.push("type function definition"),
```
kontheocharis, line 158: This is a bit unrelated to this PR, but type functions are called implicit functions in the TIR. Maybe we should unify these two names?

compiler/hash-attrs/src/target.rs
```diff
@@ -0,0 +1,292 @@
...
+    PatArg(ast::AstNodeRef<'ast, ast::PatArg>),
+
+    /// A type.
+    Ty(ast::AstNodeRef<'ast, ast::Ty>),
+
+    /// A type argument.
+    TyArg(ast::AstNodeRef<'ast, ast::TyArg>),
+}
+
+impl<'ast> AttrNode<'ast> {
```
kontheocharis, line 231: I guess this is the repetition you were trying to avoid...

compiler/hash-attrs/src/ty.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+impl AttrTy {
+    /// Create a new [AttrTy] with the given name, parameters and subject.
+    pub fn new(name: impl Into<Identifier>, params: ParamsId, subject: AttrTarget) -> Self {
+        Self { name: name.into(), params, subject }
+    }
+}
+
+macro_rules! make_ty {
```
kontheocharis, line 95: Why not just keep the original names and avoid repetition?

compiler/hash-attrs/src/ty.rs
```diff
@@ -0,0 +1,181 @@
...
+        let name: Identifier = stringify!($name).into();
+        let index = $table.map.push(AttrTy::new(name, Param::empty_seq(), $subject));
+
+        if $table.name_map.insert(name, index).is_some() {
+            panic!("duplicate attribute name: `{}`", name);
+        }
+    }
+}
+
+lazy_static::lazy_static! {
```
kontheocharis, line 139: I don't think you need lazy_static with LazyLock

compiler/hash-attrs/src/ty.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub static ref ATTR_MAP: LazyLock<AttrTyMap> = {
```
kontheocharis, line 140: Do you think the attribute definitions themselves should happen in a different module?

compiler/hash-codegen/src/symbols/mangle.rs
```diff
@@ -68,9 +68,7 @@ pub fn compute_symbol_name(instance_id: InstanceId) -> String {
     let m = &mut Mangler { out: String::new() };
 
     instance_id.map(|instance| {
-        if !instance.attributes.contains(IDENTS.no_mangle)
-            && !instance.attributes.contains(IDENTS.foreign)
-        {
+        if !instance.has_attr(IDENTS.no_mangle) && !instance.has_attr(IDENTS.foreign) {
```
kontheocharis, line 71: Would be cool if the attr map exposes the IDs of the defined attributes so they don't have to be looked up by name

kontheocharis, line 71: Would be cool if the attr map exposes the IDs of the defined attributes so they don't have to be looked up by name

compiler/hash-attrs/src/target.rs
feds01: This is a mistake, it should be just trait def.

compiler/hash-attrs/src/target.rs
feds01: Im unsure how to solve this problem.

compiler/hash-attrs/src/target.rs
feds01: I don't think that is a thing

compiler/hash-codegen/src/symbols/mangle.rs
feds01: Yeah, I wanted to do this, but I'm not sure how to go about this.

compiler/hash-attrs/src/target.rs
kontheocharis: I mean you can just do .iter().map() and not have to worry about dealing with vecs/pushing/reserving, but I guess it's fine either way

compiler/hash-attrs/src/target.rs
kontheocharis: Maybe it should always apply on the declaration, and macros should handle this accordingly?

feds01: This causes problems because the `AstNodeId` that the attributes are then registered on isn't used to determine whether it has attributes on it. E.g. when you write:

```rs
#[link_name("foo")]
foo := () => { ... }
```

Lowering and codegen use the `AstNodeId` of the `() => { ... }` rather than the declaration itself whether it specifies any attributes.

compiler/hash-attrs/src/ty.rs
feds01: I was thinking of using this somehow to allow it to specify more complex types, but yeah I guess for now we don't need to.

compiler/hash-attrs/src/target.rs
feds01: I think its kind of unavoidable here...

compiler/hash-attrs/src/target.rs
feds01: Yeah I see now, but since this isn't quite like an `enum` so the match will always

compiler/hash-attrs/src/target.rs
kontheocharis: That's possible, but it's a bit of a hack.. I guess we could argue that declarations are not "real" expressions so macros never apply to them, they always apply to the inner expressions.. Anyway I think it should be fine to keep it like this for now and if we encounter a use case of putting a macro on a declaration we can reconsider. Thoughts?

compiler/hash-attrs/src/ty.rs
kontheocharis: Yeah I just mean like in a different file cause this one is called ty.rs

compiler/hash-codegen/src/symbols/mangle.rs
kontheocharis: I guess there could be a surrounding macro which generates the table static, and accepts a bunch of entries, for which it involves define_attribute as well as a public symbol for the attribute.

compiler/hash-attrs/src/ty.rs
feds01: Actually, yes this makes sense.

compiler/hash-attrs/src/target.rs
feds01: Yeah, I'm gonna make this thread into an issue/discussion so that we can come back to this later and properly decide.

kontheocharis (dismissed):

kontheocharis (approved):

PR #931, opened 2023-08-28
ast: use `TyParams` to represent parameters in types

### LINKED ISSUE #928: Use `TyParam` instead of `TyArg` in `TupleTy` and `FnTy`
There is some confusion about what a type parameter/argument is when it comes to `TupleTy`, `FnTy` and `TyFnTy`. 

To disambiguate the parsing and supported syntax, we introduce a `TyParam` node which is essentially a `<name> (: <ty>)? (= <ty>)? | <ty>` syntax.

feds01:
closes #928


compiler/hash-ast/src/ast.rs
```diff
@@ -567,7 +567,7 @@ define_tree! {
     #[node]
     pub struct TupleTy {
         /// inner types of the tuple type
-        pub entries: Children!(TyArg),
+        pub entries: Child!(TyParams),
```
kontheocharis, line 570: This should be Params no? `<ident> : <ty> = <expr>`

compiler/hash-ast/src/ast.rs
```diff
@@ -1411,7 +1414,7 @@ define_tree! {
     #[node]
     pub struct StructDef {
         /// Type parameters that are attached to the definition.
-        pub ty_params: Children!(Param),
+        pub ty_params: OptionalChild!(TyParams),
```
kontheocharis, line 1417: Shouldn't all of these be just Params??

kontheocharis (changes requested):

compiler/hash-ast/src/ast.rs
feds01: But this is a tuple type right, this can only feature in a type position. Also, whilst it may use the `TyParam`, it doesn't allow for a default value to be specified as seen in the grammar: https://hash-org.github.io/hashc/features/types.html

compiler/hash-ast/src/ast.rs
feds01: Well, that would mean it uses `<name: ty? = <expr>?` which for type parameters isn't quite right. Also, for something like:
```rs
Foo := <T> => struct(...)
```
would mean the same as:
```rs
Foo := struct<T>(...)
```
So would it not make sense to use a type parameter here?

compiler/hash-ast/src/ast.rs
kontheocharis: Yeah well, I guess it is not supported, but then it doesn't matter if Params or TyParams is used. However, if we ever decide to support it, it should be Params, not TyParams.

compiler/hash-ast/src/ast.rs
kontheocharis: You are right I am just stupid.. But I think TupleTy should use Params like FnTy

kontheocharis: Also for the second thing you said, should
the first syntax be supported?

compiler/hash-ast/src/ast.rs
feds01: Yeah you're right, perhaps we should change `FnTy` and `TupleTy` to use `Params`? However, we would need to introduce a `Params` node (but I think that is good for consistency).

compiler/hash-ast/src/ast.rs
feds01: Whilst this concern is directly related, the question of to which items apply attributes and macros. 

If we decide to simplify the attribute application rules (as in they just apply on whatever item you specified them on), then it would be nice to avoid having something like `<T> => struct(..)` because it kind of complicates what happens in say this case:

```rs
Baz := #foo <T> => struct(..)
```
In this case, the user clearly means to apply `foo` on the struct definition, not the type function itself.

kontheocharis: True, though this is already an issue with functions anyway. There is a potential solution to this. What if each parameter also allows type arguments? E.g. we have some kind of single Params node, which appears in structs, enums, mods, functions.

And the syntax is

```
params ::= <ident: ty (= ty)?...> (ident: ty (= expr)?...)
```

Both parts () and <> are optional (recovering the original type functions and normal functions). And then the syntax for other items becomes:

```
struct ::= "struct" params fields
enum ::= "enum" params variants
function ::= params "->" ty "=>" expr
function_ty ::= 
...
```

Then there is only one kind of function and you can do something like

```
empty_list := <T> () => List<T>(data=[])

identity := <T> (t: T) -> T => t
```

What do you think about this? It would solve the attribute issue because now generics are part of the same node as the function, not an entirely separate function.

kontheocharis: It would require some changes to the parser though, idk if you think it is worth it. and I don't know if there are any ambiguities with it.

feds01: I think we should keep it as it is, and maybe later we can re-visit the subject.

kontheocharis: > For things like `enum` and `struct` and others, it doesn't make sense for them to have the normal parameter bit though, you'd end up having to write:
> ```rs
> Foo := <T>()(..)
> ```

Yeah in general for simple types it wouldn't be too useful (of course you can omit the empty () the base syntax would be unchanged. You can still write `Foo := struct <T> (field: i32)`).

But this is useful for, e.g. the equality type:

```
Equal := enum <T> (a: T, b: T) (
    Refl(x: T): Equal<T>(a, b)
)

/// Here the T is inferred
transitivity : <T, a: T, b: T, c: T> -> (Equal(a, b), Equal(b, c)) -> Equal(a, c)
                  = ...
```

In general for dependent types with indices it is useful because otherwise you have to pack all the arguments in the <> and then inference is not that nice.

But at the same time this specific thing is not strictly necessary, basically is just an added benefit. The main point is that attributes on functions would not be problematic anymore.

kontheocharis: Yeah I mean, I'm not really suggesting to do this now, I'm just trying to find a solution to the attribute issue (plus it simplifies the visual noise a bit)

kontheocharis (approved):

PR #934, opened 2023-08-31
parser+source: improvements and cleanups

feds01:
## Changes

This PR is an aggregation of various improvements and cleanups to the parser and constants API in the compiler.

Improvements to constant storage:
- Move out methods from `CONSTANT_MAP` and just place them onto each `Interned<ConstKind>` - this improves code readability in quite a few places.
- Don't use a `HashMap` for storing floats/ints as they don't need to be looked up in that manner.
- use `fxhash` to look up identifiers and strings instead of the default dashmap hasher.

Improvements to the parser:
- Use a more expressive `ExpectedItem` type to represent what the parser might be expecting when parsing items, which gives it more flexibility to specify expected items other than tokens. Additionally, it saves on quite a bit of memory around the place: the size of `ParseResult<T>` went down from a minimum bound of **70 bytes** to **40 bytes**.
-  Use a perfect hash function to lookup if an identifier is a keyword, there should be a 5% improvement in speed to lexing (there is more that could be done here, maybe use `gprerf`).
- Collect tokenisation and generation timings, and report them as other stages do.
- remove `is_compound_expr` flag on `AstGen`.
- Use `ThinVec` for diagnostics across compiler, which should reduce several items that used to store `Vec`s inline (which seemed wasteful). It also optimises for the common case of no diagnostics which effectively means that it will be a `std::ptr::nullptr()` and only takes up **8bytes** instead of **24bytes** as a `Vec` would.
- Share diagnostics among all `AstGen`s which again reduces the size of an `AstGen` from **112 bytes** to **64 bytes**.
- parser: fix bug where errors were being entirely ignored - due to not consuming errors when they should be.
- parser: fix span of `ParseWarning::RedundantParenthesees` to be more accurate.

## Todo
- [x] #931 needs to be merged before review.


compiler/hash-ast/src/ast.rs
```diff
@@ -1849,6 +1851,90 @@ define_tree! {
...
+
+        /// The definition is a `mod` block.
+        Mod,
+
+        /// Funtion type.
+        FnTy,
+
+        /// Tuple type.
+        TupleTy,
+    }
```
kontheocharis, line 1914: These are not used right?

kontheocharis (dismissed):

kontheocharis (approved):

PR #936, opened 2023-09-01
Rework the TIR structure to contain `Node<T>` nodes, which store their AST origins

### LINKED ISSUE #929: Wrap all TIR nodes in a location rather than storing them in an individual HashMap

kontheocharis:
This PR changes the TIR structure so that each node is wrapped in the `Node` type. This allows each node to carry information about its AST origin inline, rather than having to store this in an auxiliary store. For sequences of nodes, there is an outer `Node` which contains the whole sequence, and each element of the sequence is also a node. This allows storing the locations of sequences which might be empty.

This PR does not fill the correct origins for each node yet; they are currently all set to `NodeOrigin::Generated`. A subsequent PR will fill these origins correctly and then remove `AstInfo` and `LocationStore`, which will become redundant.
This should hopefully also result in speedups for semantic analysis, by dramatically reducing the number of HashMap operations that need to happen.

closes #929


compiler/hash-intrinsics/src/utils.rs
```diff
@@ -229,51 +239,60 @@ pub trait PrimitiveUtils: AccessToEnv {
...
     fn create_term_from_char_lit(&self, lit: char) -> TermId {
-        Term::create(Term::Lit(Lit::Char(CharLit { underlying: ast::CharLit { data: lit } })))
+        Node::create_at(
+            Term::Lit(Lit::Char(CharLit { underlying: ast::CharLit { data: lit } })),
+            NodeOrigin::Generated,
+        )
     }
 
-    /// Get the given term as an integer literal if possible.
+    /// Get the given terman integer literal if possible.
```
feds01, line 293: Typo

compiler/hash-lower/src/build/matches/declarations.rs
```diff
@@ -106,7 +106,7 @@ impl<'tcx> BodyBuilder<'tcx> {
             }
             Pat::Ctor(CtorPat { ctor_pat_args: fields, ctor_pat_args_spread: spread, .. })
             | Pat::Tuple(TuplePat { data: fields, data_spread: spread }) => {
-                fields.borrow().iter().for_each(|field| {
+                fields.borrow().borrow().iter().for_each(|field| {
```
feds01, line 109: Whats the second `borrow()` for? What does `fields.borrow()` return?

compiler/hash-lower/src/build/matches/test.rs
```diff
@@ -15,7 +15,7 @@ use hash_ir::{
 };
 use hash_reporting::macros::panic_on_span;
 use hash_source::constant::{IntConstant, IntConstantValue, InternedInt};
-use hash_storage::store::statics::StoreId;
+use hash_storage::store::statics::CoreStoreId;
```
feds01, line 18: There is alot PR noise here, maybe this should be in a different PR?

compiler/hash-tir/Cargo.toml
```diff
@@ -12,6 +12,8 @@ num-bigint = "0.4"
 textwrap = "0.16"
 utility-types = "0.0.2"
 typed-builder = "0.11"
+paste = "1.0"
+derive_more = "0.99"
```
feds01, line 16: This is exported from `hash_utils` already

compiler/hash-tir/src/environment/stores.rs
```diff
@@ -66,33 +85,265 @@ macro_rules! tir_debug_value_of_sequence_store_element_id {
...
+/// `hash_storage::static_sequence_store_direct!` to define the store and its ID
+/// type; it creates a single store from key `$id` to value `Node<$id_seq>`, and
+/// a direct sequence store from key `$id_seq` to values in `$value`.
+///
+/// This can be used to define TIR nodes like:
+///
+/// ```ignore
+/// tir_node_sequence_store_direct!(Param);
+///
+/// /// These now exist:
```
feds01, line 181: Maybe add docs for what each one is responsible for and its purpose? Specifically `ParamsSeqId` and `ParamsId`

compiler/hash-tir/src/node.rs
```diff
@@ -0,0 +1,111 @@
...
+        Self { data, origin }
+    }
+
+    /// Create a new node with the given data and a generated origin.
+    pub fn gen(data: Data) -> Self {
+        Self { data, origin: NodeOrigin::Generated }
+    }
+
+    /// Get the node ID of this node.
+    pub fn node(&self) -> Option<AstNodeId> {
```
feds01, line 58: I wonder if we can get away with a node always having a `AstNodeId`

compiler/hash-tir/src/node.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum NodeOrigin {
+    /// The node was given by the user.
+    Given(AstNodeId),
+    /// The node was created through type inference of the given origin.
+    InferredFrom(AstNodeId),
+    /// The node was generated by the compiler, and has no origin.
+    Generated,
```
feds01, line 101: Generated things should come from intrinsics right? So, what if define the defs in source, parse them and then use those as the locations?

tests/cases/typecheck/property_access/out_of_bounds_access.stderr
```diff
@@ -1,6 +1,6 @@
-error[0016]: property `2` not found on type `(s517: i32, s518: i32)`
+error[0016]: property `2` not found on type `(s517: i32, s519: i32)`
```
feds01, line 1: Unrelated to this PR, but perhaps we should change tuple type printing to omit name field if no name exists.

compiler/hash-lower/src/build/matches/declarations.rs
kontheocharis: `fields: ParamsId`, so `fields.borrow(): &Node<ParamsSeqId>`, so `fields.borrow().borrow(): &[Node<ParamId>]`. Maybe I will change this to `.members().borrow()` since that does the same thing, but is more clear.

compiler/hash-lower/src/build/matches/test.rs
kontheocharis: You mean the rename? I also was thinking of undoing this change tbh, instead having `StoreId -> SingleStoreId` and `CoreStoreId -> StoreId` which should restore most of the old names...

compiler/hash-tir/src/node.rs
kontheocharis: I think sometimes this is not possible, for example the primitives are made up of nodes but are not defined in any source file.

compiler/hash-tir/src/node.rs
kontheocharis: It's possible to do this, but I think it should be done as a separate task

compiler/hash-lower/src/build/matches/declarations.rs
kontheocharis: Yeah this already exists, i just need to change these .borrow().borrow() call sites.

compiler/hash-lower/src/build/matches/test.rs
feds01: Yeah, I think keeping old names would be better. In terms of users of the stores, the `CoreStoreId` name seems a bit strange

compiler/hash-tir/src/node.rs
feds01: Oh yeah of course

compiler/hash-tir/src/node.rs
feds01: Also, is there a need to keep track of what is inferred or generated?

kontheocharis: Yes, because it can lead to better error messages. For example, "type X given here` vs "type X inferred from this expression"

feds01 (approved):

PR #937, opened 2023-09-02
defer literal parsing

### LINKED ISSUE #888: Defer literal parsing until type is known concretely
Currently, the parser immediately tries to parse float/integer literals into some form. It does this by assuming that the user specifies the largest literal. However, later on at typechecking, it may be the case that a smaller integer literal was actually meant and then the constant must be re-mapped into the smaller variant. This isn't ideal because it causes more work for constant remapping and overall feels hacky. To solve this, the parser should defer integer/float literal parsing until a point in time when the type is concretely known (whether by annotation or context).

This will do two things:
- remove the weird constant remapping hack.
- shrink the size of tokens and literal AST since they don't need to store or parse the literal (until TC).

However, this means that we have to move literal parsing logic (to an extent) out of the lexer/parser or isolate it so it can be used in typechecking contexts. It might also require some retooling to occur in `hash-source`, such that anyone in the compiler can query "snippets" of source code based on text in a more ergonomic way.

feds01:
closes #888


compiler/hash-parser/src/parser/mod.rs
```diff
@@ -137,13 +152,13 @@ impl<'stream, 'resolver> AstGen<'stream, 'resolver> {
...
     pub(crate) fn span(&self) -> Span {
-        Span { span: self.parent_span, id: self.resolver.source() }
+        Span { range: self.parent_span, id: self.resolver.source() }
     }
 
     /// Function to create a [Span] from a [ByteRange] by using the
     /// provided resolver
     pub(crate) fn make_span(&self, span: ByteRange) -> Span {
-        Span { span, id: self.resolver.source() }
+        Span { range: span, id: self.resolver.source() }
```
feds01, line 161: rename `span` to `range`

compiler/hash-intrinsics/src/utils.rs
```diff
@@ -239,61 +239,51 @@ pub trait PrimitiveUtils: AccessToEnv {
...
-                        IntConstantValue::Big(Box::new(lit.into())),
-                        None,
-                    )),
-                },
-            })),
-            NodeOrigin::Generated,
-        )
+    fn create_term_from_integer_lit<L: Into<IntConstantValue>>(&self, lit: L) -> TermId {
+        let lit = Lit::Int(InternedInt::create(IntConstant::new(lit.into(), None)).into());
+        Node::create_at(Term::Lit(Node::create_gen(lit)), NodeOrigin::Generated)
```
kontheocharis, line 260: Can use `create_gen` instead of `create_at(.., Generated)`

kontheocharis (approved):

PR #944, opened 2023-09-03
Fill the `origin` fields of all TIR nodes

### LINKED ISSUE #941: Add `Node<..>` to exhaustiveness structures

### LINKED ISSUE #943: Fill out AST IDs in the TIR `origin` fields

kontheocharis:
This PR:
- fills the `origin` field of each TIR node to its appropriate `AstNodeId`,
- ~adds `Node<..>` wrappers to the exhaustiveness data structures and appropriately fills their origins,~
- and adds a new `NodeOrigin::Computed` variant for results from CTE.

When the `NodeOrigin::Generated` variant is used, it is accompanied by a `##GeneratedOrigin` note in order to justify why this is okay. Additionally, there are some `@@MissingOrigin` action items which require some more thought about what the origin should be.

This PR does not yet remove the old location store/AST info map (#942). This will be done in the next PR.
Closes #941, closes #943


compiler/hash-exhaustiveness/src/storage.rs
```diff
@@ -1,13 +1,14 @@
...
 
 use crate::{construct::DeconstructedCtor, deconstruct::DeconstructedPat};
 
 new_store_key!(pub DeconstructedPatId, derives = Debug);
-pub type DeconstructedPatStore = DefaultStore<DeconstructedPatId, DeconstructedPat>;
+pub type DeconstructedPatStore = DefaultStore<DeconstructedPatId, Node<DeconstructedPat>>;
 
 new_store_key!(pub DeconstructedCtorId, derives = Debug);
-pub type DeconstructedCtorStore = DefaultStore<DeconstructedCtorId, DeconstructedCtor>;
+pub type DeconstructedCtorStore = DefaultStore<DeconstructedCtorId, Node<DeconstructedCtor>>;
```
feds01, line 11: Not sure this is right to do, there isn't really s need for them to be in nodes.

compiler/hash-semantics/src/passes/discovery/defs.rs
```diff
@@ -485,9 +500,10 @@ impl<'tc> DiscoveryPass<'tc> {
                 buf.push((
                     node.id(),
                     Decl {
-                        name: sym(binding.name.ident),
-                        // is_mutable: binding.mutability.as_ref().map(|m| *m.body())
```
feds01, line 489: Maybe a leave a comment that mutability should be added here in `@@Future`?

compiler/hash-semantics/src/passes/resolution/paths.rs
```diff
@@ -51,8 +51,8 @@ use crate::diagnostics::error::{SemanticError, SemanticResult};
 #[derive(Clone, Debug)]
 pub struct AstPathComponent<'a> {
     pub name: Identifier,
-    pub name_span: Span,
-    pub args: Vec<AstArgGroup<'a>>,
+    pub name_node_id: AstNodeId,
+    pub args: Node<Vec<AstArgGroup<'a>>>,
     pub node_id: ast::AstNodeId,
```
feds01, line 56: Nice!

compiler/hash-semantics/src/passes/resolution/scoping.rs
```diff
@@ -146,13 +147,16 @@ impl<'tc> Scoping<'tc> {
     pub(super) fn lookup_symbol_by_name_or_error(
         &self,
         name: impl Into<Identifier>,
-        span: Span,
+        origin_node_id: AstNodeId,
```
feds01, line 150: Overly verbose name, perhaps `origin` or `node`

feds01 (changes requested):

compiler/hash-semantics/src/passes/discovery/defs.rs
kontheocharis: I don't think that's necessary tbh, we have to figure out mutability/memory as a whole anyway, which will probably require additions in lots of places..

compiler/hash-exhaustiveness/src/storage.rs
kontheocharis: It is so that the recovered `TIR` nodes can have their original origins

feds01: But they don’t need an origin, its only ever used in a position so that it can be printed, the locations are irrelevant

kontheocharis: I guess, though it might be useful for richer errors, for example to justify why there are/aren't more constructors it can point to the constructors definition.. I could also remove it if you think that is better..

feds01: that is impossible since the ones that will be generated will always have an un-trackable span

feds01: I think we can remove it for now until there is a more explicit intent on using the information. If it is later deemed that the information is useful, we can re-add it. Perhaps we should save those changes as a patch somewhere?

feds01 (approved):

PR #957, opened 2023-09-04
re-implement `#dump_ast` and re-organise AST printing code

feds01:
- ast+expand: imlement `#dump_ast` directive
- ast: move `tree.rs` into `hash-ast-utils`
- move AST pretty printing from `hash-fmt` into `hash-ast-utils`


compiler/hash-ast-utils/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+//! Defines various utilties that can be used to hash the AST. This crate
```
kontheocharis, line 1: `hash the AST`?

kontheocharis (approved):

PR #961, opened 2023-09-08
TIR: Merge `Ty` and `Term` enums

### LINKED ISSUE #947: Merge the `Term` and `Ty` enums, removing lots of redundancy in the process

kontheocharis:
Due to the presence of `Term::Ty` and `Ty::Eval`, the `Term` and `Ty` enums are mutually recursive. Additionally, `Var` and `Hole` exist both on `Term` and `Ty`, introducing the annoying equivalences `Ty::Eval(Term::Var(x)) == Ty::Var(x)`, `Ty::Eval(Term::Ty(x)) = x` etc. In order to eliminate all of this redundancy, this PR merges the two enums into `Term`, by moving the rest of the `Ty` variants into `Term`. This greatly simplifies a lot of code while keeping the size of `Term` to 32 bytes.

This is the first in a series of PRs to simplify the TIR representation structure. The next PR will focus on removing the `FnDefStore` and storing functions inline inside `Term` as well.

Closes #947.


compiler/hash-tir/src/terms.rs
```diff
@@ -33,6 +35,12 @@ pub struct UnsafeTerm {
     pub inner: TermId,
 }
 
+/// Infer the type of the given term, returning its type.
+#[derive(Debug, Clone, Copy)]
+pub struct TypeOfTerm {
```
feds01, line 40: `TyOfTerm`

compiler/hash-tir/src/terms.rs
```diff
@@ -85,28 +95,45 @@ pub enum Term {
...
     Hole(Hole),
 }
 
 tir_node_single_store!(Term);
 tir_node_sequence_store_indirect!(TermList[TermId]);
 
+pub type Ty = Term;
+pub type TyId = TermId;
+pub type TyListId = TermListId;
+
```
feds01, line 130: Interesting hack, I was wondering how it was possible to have `Term` and `Ty` in the same match block.

compiler/hash-typecheck/src/inference.rs
```diff
@@ -1606,7 +1550,7 @@ impl<T: AccessToTypechecking> InferenceOps<'_, T> {
                 self.infer_block_term(&block_term, annotation_ty, term_id)?
             }
             Term::TypeOf(ty_of_term) => {
-                self.infer_type_of_term(ty_of_term, annotation_ty, term_id)?
+                self.infer_termpe_of_term(ty_of_term, annotation_ty, term_id)?
```
feds01, line 1553: `termpe_of_term`?

compiler/hash-lower/src/discover.rs
```diff
@@ -160,7 +160,6 @@ impl FnDiscoverer<'_> {
         self.traversing_utils()
             .visit_term::<!, _>(term, &mut |atom: Atom| match atom {
                 Atom::Term(_) => Ok(ControlFlow::Continue(())),
-                Atom::Ty(_) => Ok(ControlFlow::Break(())),
```
feds01, line 163: It's not strictly necessary to change this, but should the updated version also break on type-like terms?

kontheocharis: Technically the correct thing to do is to recurse in case that there are indeed some functions in there. For example `(a: i32, b: i32 -> i32 = (x) => x)` is a valid type.

feds01: But in that example, it is in the `default` expression position - not inside the type itself.

kontheocharis: Yes but the root node is a `TupleTy` so if `Break` is given for that then the function will never be seen.

kontheocharis: I know that technically default values are not supported yet but this is a side point

feds01: Yeah ok, this makes sense. 

Side note: In the current implementation, this wouldn't quite work anyway (I think).

kontheocharis: Yeah it wouldn't, but I hope soon it will

feds01 (approved):

PR #967, opened 2023-09-12
lower+codegen: Introduce and implement handling for new `Const` representation

feds01:
- codegen: simplify handling of zero-sized operands
- codegen(llvm): use a patched version of LLVM for constant lookups
- ir: simplify `ir::Const` by removing `Unevaluated` variant
- ir: introduce new `Const` representation and implement printing/conversions
- lexer: implement `todo!()` in `eat_float_lit`
- ir: move printing logic into `hash-ir-utils`
- ir: move out `compare_constant_values` into `hash-lower`
- lower: implement `compare_const_values` for new `Const` representation
- lower: remove constant folding for now
- ir: rename `Const::from_int` to `Const::from_scalar_like`
- lower: implement converting from  TIR literals into new `Const` representation
- codegen: implement translating new IR `Const` representation
- lower: switch to using new constant representation
- lower: re-add constant folding for constant binary operators
- ir: remove conversion impl from `Scalar` to `IntConst`
- codegen(llvm): re-work code generation for `transmute` intrinsic calls
- codegen: switch to using opaque pointers


compiler/hash-ir/src/constant.rs
```diff
@@ -0,0 +1,537 @@
...
+            IntConstantValue::U16(val) => Scalar::from_uint(val, size),
+            IntConstantValue::U32(val) => Scalar::from_uint(val, size),
+            IntConstantValue::U64(val) => Scalar::from_uint(val, size),
+            IntConstantValue::U128(val) => Scalar::from_uint(val, size),
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Constructor)]
+pub struct ScalarInt {
```
kontheocharis, line 363: What is the purpose of this struct? Is it to "parse" a scalar into an int?

compiler/hash-ir/src/constant.rs
```diff
@@ -0,0 +1,537 @@
...
+        value
+            .to_bits(Size::from_bytes(std::mem::size_of::<f64>()))
+            .map(|u| f64::from_bits(u as u64))
+    }
+}
+
+/// An [Alloc] represents a single allocation slot which contains the actual
+/// data, alignment, and mutability of the allocation.
+#[derive(Clone, Debug)]
+pub struct Alloc<Buf = Box<[u8]>> {
```
kontheocharis, line 426: One potential thing to do is to use this for some TIR arrays. E.g. maybe 
```
enum ArrayTerm {
   Concrete(Alloc),
   Symbolic(TermListId)
}
```
Some more thought is needed for this.

compiler/hash-ir/src/constant.rs
```diff
@@ -0,0 +1,537 @@
...
+
+impl AllocRange {
+    /// Get the end of the [AllocRange].
+    pub fn end(&self) -> Size {
+        self.start + self.size
+    }
+}
+
+pub trait AllocBuf:
+    Clone + fmt::Debug + Eq + PartialEq + Deref<Target = [u8]> + DerefMut<Target = [u8]>
```
kontheocharis, line 459: Why needing all of these bounds?

compiler/hash-ir/src/constant.rs
```diff
@@ -0,0 +1,537 @@
...
+        }
+    }
+}
+
+/// A scalar value. [Scalar]s are used to represent all integer, characters, and
+/// floating point values, as well as integers. The largest scalar value is
+/// 128bits, i.e. a `u128` or `i128`.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[repr(packed)]
+pub struct Scalar {
```
kontheocharis, line 184: This can definitely be used as a variant of `Term`, and the rest of the integer types can be data types defined using it

kontheocharis:
I'm just wondering how to deal with the representation overlap of alloc and scalar. Ideally at the TIR level there should be a `byte` type which can be placed into arrays `[byte; 4]` etc. And values of type `[byte; 4]` can be `ArrayTerm::Concrete` using `Alloc`. However this means that `Scalar` will only be used for `byte`. So perhaps `ArrayTerm::Small` could be a `Scalar` if the array is an array of bytes less than 16. But then this introduces equivalent representations which is not ideal..

compiler/hash-ir/src/constant.rs
feds01: Its only really used for actually printing the `Scalar` since the scalar itself has no type info on it.

compiler/hash-ir/src/constant.rs
feds01: Yeah perhaps it could use the `Alloc`. However, this would mean that the Alloc/Scalar representations (which sometimes use type info) would have to be type agnostic. 

Since it would need to understand both `IrTy`s and `Ty`s. The simple way would be to convert the TIR type into an IR type? If we do this, we'll need to do some minor restructuring of the crates to avoid cyclic deps.

feds01: Ideally, `hash-ir` shouldn't depend on `hash-tir`, and the alloc/scalar stuff is moved into something like `hash-source`. Then, `hash-ir-utils` could provide the tools to convert between type representations, and interface with the alloc API (as it does now)?

compiler/hash-ir/src/constant.rs
```diff
@@ -0,0 +1,537 @@
...
+        debug_assert_eq!(
+            self.size().truncate(self.value),
+            { self.value },
+            "Scalar value is too large for the specified size"
+        )
+    }
+
+    #[inline]
+    pub fn to_bits(self, target_size: Size) -> Result<u128, Size> {
+        assert_ne!(target_size.bytes(), 0, "you should never look at the bits of a ZST");
```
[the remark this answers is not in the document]
feds01: Well it is assumed that the scalar cannot be of length zero, so if an attempt is made to read it as a ZST, then something has gone quite wrong.

compiler/hash-ir/src/constant.rs
feds01: - `Clone`, `Debug` are implied by the store
- `Eq` and `PartialEq` could be removed.
- `Deref` and `DerefMut` so the bytes can be directly accessed.

feds01:
> I'm just wondering how to deal with the representation overlap of alloc and scalar. Ideally at the TIR level there should be a `byte` type which can be placed into arrays `[byte; 4]` etc. And values of type `[byte; 4]` can be `ArrayTerm::Concrete` using `Alloc`. However this means that `Scalar` will only be used for `byte`. So perhaps `ArrayTerm::Small` could be a `Scalar` if the array is an array of bytes less than 16. But then this introduces equivalent representations which is not ideal..

Ideally, you should offload the concerns of the actual representation behind the const, and then simply interface with the constant without having explicit knowledge about the underlying representation (whether small or large). This problem is kind of solved by lowering/codegen by letting the `AbiRepresentation` (computed by layouts) dictate if something can be a scalar or not.

kontheocharis (approved):

PR #969, opened 2023-09-12
TIR: Intrinsic overhaul, removing `hash-intrinsics` and making everything more declarative

kontheocharis:
- Both intrinsics and primitives are now declared in `hash_tir::intrinsics::definitions` in a declarative syntax that mirrors Hash.
- The macro which accepts this DSL generates various convenient items for using and referencing these definitions.
- To represent intrinsics in the TIR, a new term `Term::Intrinsic` is added, which accepts a single enum argument that represents the chosen intrinsic. 
- There is no longer dynamic dispatch for calling intrinsics, which should improve compile-time performance.
- Intrinsics are now allowed to return `Ok(None)` to signal that the call to the intrinsic cannot be simplified.


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 members = [
-  "compiler/*",
+  "compiler/hash*",
```
feds01, line 3: ?

compiler/hash-ir/src/ir.rs
```diff
@@ -16,6 +15,10 @@ use hash_storage::{
         SequenceStore, SequenceStoreKey,
     },
 };
+use hash_tir::intrinsics::definitions::{
+    BinOp as TirBinOp, CondBinOp as TirCondBinOp,
+    ShortCircuitingBoolOp as TirShortCircuitingBoolOp, UnOp as TirUnOp,
+};
```
feds01, line 21: Haha, verbose naming 💯  

I feel like we should unify operators in `hash-source`, especially since we will kind of need to do that if we want to share using the `Const` representation across the compiler. Perhaps, even define it in `hash-token`?

It would eliminate:
- `ir::BinOp` `ir::UnOp`
- `ast::BinOp` `ast::UnOp`
- `tir::BinOp`  `tir::UnOp`

But maybe the separation makes sense?

compiler/hash-semantics/src/diagnostics/error.rs
```diff
@@ -254,6 +257,16 @@ impl<'tc> WithSemEnv<'tc, &SemanticError> {
...
+            SemanticError::CannotUseIntrinsicInPatternPosition { location } => {
+                let error = reporter
+                    .error()
+                    .code(HashErrorCode::ValueCannotBeUsedAsType)
+                    .title("cannot use an intrinsic in pattern position");
+
+                error.add_span(*location).add_info(
+                    "cannot use this in pattern position as it refers to a compiler intrinsic",
+                );
+            }
```
feds01, line 269: Can we get a test case for this?

compiler/hash-tir/src/intrinsics/utils.rs
feds01, line 1: With the new constant representation, this impl should mostly vanish right?

compiler/hash-tir/src/intrinsics/mod.rs
```diff
@@ -0,0 +1,4 @@
+//! Contains definitions related to intrinsics
```
feds01, line 1: nit:
```suggestion
//! Contains definitions related to intrinsics.
```

compiler/hash-lower/src/build/ty.rs
```diff
@@ -102,63 +112,83 @@ impl<'tcx> BodyBuilder<'tcx> {
...
+                            try_use_term_as_integer_lit::<_, u8>(self.env(), op).unwrap(),
+                        )
+                        .unwrap();
+
+                        FnCallTermKind::UnaryOp(parsed_op.into(), subject)
+                    }
+                    TirIntrinsic::SizeOf
+                    | TirIntrinsic::AlignOf
+                    | TirIntrinsic::PtrOffset
+                    | TirIntrinsic::Eval
```
feds01, line 186: Perhaps `Eval`, `UserError` and `DebugPrint` should panic here since they aren't meant to be "lowered"?

compiler/hash-lower/src/discover.rs
```diff
@@ -144,6 +137,10 @@ impl FnDiscoverer<'_> {
                     ModMemberValue::Fn(_) => {
                         // We've already found this one.
                     }
+                    ModMemberValue::Intrinsic(_) => {
+                        // @@Todo: we should probably add all the intrinsic
+                        // functions?
+                    }
```
feds01, line 143: It shouldn't really need to do anything here, after TC, it is up to the backend to emit code for the intrinsic. For example take a look at what the backends do for the various intrinsics:
https://github.com/hash-org/hashc/blob/b17713d5deb9ce802fdaf8f527dcf06d27c56c3b/compiler/hash-codegen/src/lower/intrinsics.rs#L31-L68

Cargo.toml
kontheocharis: Sometimes it is helpful to temporarily remove a crate from the workspace by not prefixing it with `hash-`, otherwise it has to be moved out entirely

compiler/hash-tir/src/intrinsics/utils.rs
kontheocharis: Yes, but I will do that in another PR

compiler/hash-ir/src/ir.rs
kontheocharis: I think it's a bit different because the operators defined in the AST do not necessarily correlate with the "intrinsic" operations that can be defined on TIR terms. For example, with the new constant stuff a lot of the intrinsics will greatly simplify/become redundant. I think I will first adapt to `Const` and then we can consider this separation more.

compiler/hash-tir/src/intrinsics/utils.rs
feds01: Yeah it's just a self note, no action is required.

compiler/hash-ir/src/ir.rs
feds01: When we do integrate the new constant stuff, we will still have to define some kind of common operator kinds, I guess for now the IR can just rely on them and the TIR can define conversions for them.

kontheocharis: Okay. I agree that there should be some unification here..

feds01 (approved):

PR #973, opened 2023-09-13
Lexer+parser improvements

feds01:
This PR improves performance in the lexer and parser (especially in the AST generation part). 

## Parser

Previously, when allocating a new node through `AstNode::new()`, we would make a call to `SPAN_MAP` in order to store the `Span` of the node and receive a corresponding `AstNodeId`. Whilst in practice this is fine, it means that we are constantly contending on the `SPAN_MAP` lock, which isn't ideal, especially with many `parser` workers. In fact, it often led to far poorer performance with more threads than fewer threads.

To fix this issue, we introduce a `LocalSpanMap` which is used by a `parser` thread to allocate `Span`s and `AstNodeId`s for new nodes at the local level. Once the current parser job is done, the `LocalSpanMap` is consumed and merged into the
global `SpanMap`. This means that writes to the global `SpanMap` only occur in between jobs, and perform big bulk writes, instead of many small writes.

Overall, this is a massive performance boost (especially with a higher number of threads). We're looking at a 3.5x performance improvement with 4+ parser threads.

Here are some comparison metrics on a `project` which consists of 1.2m loc with 1000+ files:

| Before | After |
| ------ | ----- |
| parse: `1.678s` |  parse: `356.748ms` |

a 4x performance improvement with a high number of threads.

### Other improvements
- Avoid copying over `AstGen` for each new token tree.
- Don't use `RefCell` for diagnostics in the parser.
- Introduce `AstGenFrame` which manages data related to the parsing "frame" i.e. a token tree.
- Remove useless `'resolver` lifetime.

## Lexer
- Experiment with `gperf` gnu.org/software/gperf/ to avoid using `phf` hashmap which seemed to take a longer amount of time when hashing keys/keywords.  (2% speedup).
- Eat newlines in line comments faster (3% speedup).
- Don't overzealously eat a shebang, when in actuality it is a module attribute declaration.
- De-duplicate string literals at the local level, and sync later. (Perhaps there could be some other way of doing the same thing with `LocalSpanMap` rather than blocking in `parse_source`).


compiler/hash-token/keywords/hash.c
```diff
@@ -0,0 +1,174 @@
+/* ANSI-C code produced by gperf version 3.1 */
```
kontheocharis, line 1: Oh goodness.. Never thought I'd C the day...

compiler/hash-token/keywords/hash.gperf
```diff
@@ -0,0 +1,35 @@
+%{
```
kontheocharis, line 1: Could this file be generated by a macro similar to phf?

compiler/hash-token/src/keyword.rs
```diff
@@ -42,40 +44,34 @@ impl fmt::Display for Keyword {
     }
 }
 
-/// A static map of keywords to their enum variants using a
-/// perfect hashing function to quickly lookup the keyword.
-///
-/// @@Future: look into somehow using `gperf` instead?
-static KEYWORDS: phf::Map<&'static str, Keyword> = phf_map! {
```
kontheocharis, line 49: I wonder if all of this extra stuff is worth it for a 2% gain...

compiler/hash-token/keywords/hash.gperf
feds01: Yeah, we could do it directly from rust, run the `gperf`, and output the `keywords.c` in the build file

compiler/hash-token/src/keyword.rs
feds01: Yeah, I was wondering about this, it might not be entirely worth it. I need to do some more direct benchmarking to see how much of a win this is actually.

compiler/hash-token/keywords/hash.c
feds01: Removed `gperf` get we won't C the day just yet...

kontheocharis (approved):

PR #977, opened 2023-09-14
Organise TIR into folders

### LINKED ISSUE #945: Reorganise the TIR into folders

kontheocharis:
Closes #945


compiler/hash-tir/src/nodes/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod args;
+pub mod data;
+pub mod lits;
+pub mod mods;
+pub mod node;
```
feds01, line 5: Why not put the definitions of this file into `mod.rs`

compiler/hash-tir/src/nodes/mod.rs
feds01, line 1: Perhaps it would make sense to re-export all these modules from this file to avoid having to write `nodes::<item>::<sub-item>` everywhere.

compiler/hash-tir/src/nodes/terms/mod.rs
```diff
@@ -9,26 +9,41 @@ use hash_storage::store::{
...
 };
 
+pub mod access;
+pub mod arrays;
+pub mod casting;
+pub mod control;
+pub mod fns;
+pub mod holes;
+pub mod refs;
+pub mod tuples;
```
feds01, line 45: Same here, it seems a bit pedantic to have to import `hash_tir::nodes::terms::tuples::TupleTerm`, I think a couple of these modules should be re-exported.

compiler/hash-tir/src/nodes/mod.rs
feds01, line 1: I think this module could be called `tir` instead of `nodes`. If you look at the other compiler representations, the core AST and IR definitions are located in files called `ast.rs` and `ir.rs`. This has the benefit that you can write `ast::<SomeItem>`, `ir::<SomeItem>` which quite clearly expresses that we're handling data from a specific representation.

compiler/hash-tir/src/nodes/mod.rs
kontheocharis: What definitions?

compiler/hash-tir/src/nodes/mod.rs
kontheocharis: I guess, though I don't see how that is so necessary since we rarely have to deal with imports directly; language server can deal with that..

kontheocharis: Yeah okay, I will change that

compiler/hash-tir/src/nodes/mod.rs
feds01: The ones in `node.rs`

kontheocharis: Because it is specifically stuff relevant to the `Node` wrapper that is present around the TIR atoms

compiler/hash-tir/src/nodes/mod.rs
kontheocharis: Anyway I will change it

compiler/hash-tir/src/nodes/mod.rs
feds01: Yeah, but it is the base for everything in this sub-module - it doesn’t seem useful to put it into its own file.

kontheocharis: I don't know, it seems cleaner to me to have a separate file for this. It makes it more clear where the "general" node stuff is; the `mod.rs` is just there to declare everything. I will add re-exports so the imports in other places should remain unchanged.

feds01: Fair enough.

feds01 (dismissed):

feds01 (approved):
LGTM, -220 lines is a benefit too

PR #979, opened 2023-09-14
semantics: implement pattern bind checking

### LINKED ISSUE #548: typechecker: incorrect error report regarding un-uniform pattern binds in compound patterns
The following snippet:

```rust
>>> (d | c) := 2;
```

Should produce this behaviour: This pattern is invalid in the declaration position because all children of the compound `|`(or) pattern do not bind `d` and `c`, as in the left-hand-side doesn't bind `c`, and the right-hand-side not binding `d`. The error reporting does not point at the violating children and instead points at the right-hand-side of the `|`  pattern

Instead, this happened: the error report that is printed is not concise.

```
error[0081]: variables `d`, and `c` are not declared in all patterns
 --> <interactive>:1:6
1 |   (d | c) := 2;
  |        ^ pattern doesn't bind `d`, and `c`
```

feds01:
closes #548


kontheocharis (approved):
Looks great!

tests/cases/typecheck/pat_binds/duplicate_binds.hash
```diff
@@ -0,0 +1,10 @@
+// run=fail, stage=typecheck
+
+main := () => {
+    k := (1, 2);
+
+    match k {
+       (a, a) => {}
+    }
+}
+}
```
feds01, line 10: Test case isn't actually testing the functionality

kontheocharis (approved):

PR #984, opened 2023-09-18
TIR: Add `Map` and `Visit` traits to the `tir::Visitor` in order to reduce code duplication

kontheocharis:
Instead of having `fmap_*` and `visit_*` functions for each TIR node, this PR replaces those with implementations of `Visit<*>` and `Map<*>`, so that functions can be implemented for any `T` that has `Map<T>` or `Visit<T>`. This reduces a lot of duplication in TC substitutions, and allows specialised functions to be created generically, for example to copy nodes, or to visit infallibly.

### DIFF
--- a/compiler/hash-tir/src/visitor.rs
+++ b/compiler/hash-tir/src/visitor.rs
@@ -21,12 +21,6 @@ use crate::{
     },
 };
 
-/// Contains methods to traverse the Hash TIR structure.
-pub struct Visitor {
-    visited: RefCell<HashSet<Atom>>,
-    visit_fns_once: bool,
-}
-
 /// An atom in the TIR.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, From, TryInto)]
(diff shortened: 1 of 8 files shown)


compiler/hash-tir/src/visitor.rs
```diff
@@ -65,46 +59,170 @@ impl fmt::Display for Atom {
...
+pub struct Visitor {
+    // @@Todo @@Performance: remove this in favour of referencing functions symbolically.
+    visited: RefCell<HashSet<Atom>>,
+    visit_fns_once: bool,
+}
+
+/// Contains the implementation of `fmap` and `visit` for each atom, as well as
+/// secondary components such as arguments and parameters.
+impl Visitor {
+    /// Create a new `TraversingUtils`.
```
feds01, line 72: New `Visitor`

feds01 (dismissed):

feds01 (approved):

PR #985, opened 2023-09-18
lexer+parser: many improvements and cleanups

feds01:
- token: remove `DelimiterVariant`
- lexer: flatten token stream
- parser: remove `AstGenFrame::token_at()`
- parser: cleanup accesses to `offset` in `AstGenFrame`
- token: introduce `->` and `=>` tokens to simplify parsing & error reporting
- token: introduce `::` tokens to simplify parsing access expr/ty/pats
- parser: cleanup `begins_pat()` implementation
- parser: avoid using `peek_nth()` in binary expression parsing
- token: introduce `..`, `..<`, `...` tokens to simplify spread/range parsing
- parser: fix typo in diagnostics
- parser: several cleanups, and stricter use of token stream API
- parser: Integrated new lexer into the parser
- parser: directly use `TokenCursor` API
- parser: avoid using confusing `next_pos()` function
- analysis: use indexmap in pattern bind analysis to enforce stable error order
- parser: name errors consistently, and remove a bunch of old un-used variants
- parser: remove use of confusing `next_pos()` and replace with `eof_pos()` or `expected_pos()`
- source: Change `ByteRange` to be inclusive on both ends
- lexer: cleanup + greatly improve lexer errors


compiler/hash-parser/src/parser/pat.rs
```diff
@@ -112,9 +108,9 @@ impl<'s> AstGen<'s> {
                 }
                 // An access pattern which accesses the `subject` with a particular `property`
                 // denotes with a name.
-                TokenKind::Colon if matches!(self.peek_second(), Some(token) if token.has_kind(TokenKind::Colon)) =>
+                TokenKind::Access =>
                 {
-                    self.frame.skip(2);
+                    self.skip_fast(); // `::`
```
kontheocharis, line 113: Would it make sense to have some kind of `self.skip_fast(TokenKind::Access)` just for a sanity check?

compiler/hash-parser/src/parser/pat.rs
```diff
@@ -292,9 +261,21 @@ impl<'s> AstGen<'s> {
     /// Parse a [ModulePatEntry]. The [ModulePatEntry] refers to
     /// destructuring either a struct or a namespace to extract fields,
     /// exported members. The function takes in a token atom because both
-    /// syntaxes use different operators as pattern assigners.
+    /// syntax's use different operators as pattern assigners.
```
kontheocharis, line 264: I think the previous line is correct..

kontheocharis (dismissed):

compiler/hash-parser/src/parser/pat.rs
feds01: I thought so too, ill change it back.

compiler/hash-parser/src/parser/pat.rs
feds01: I guess it can have a `debug_assert!()` in `skip_fast()`?

feds01: Although this might be annoying to do when we skip literals since you can't just pass a pattern.

kontheocharis: Could use a macro for that: `debug_assert!(matches!(x, $token_pat))`. But anyway I don't know if this is worth it or not

kontheocharis (approved):

PR #987, opened 2023-09-19
exhaustiveness: speedup checking 3x-5x

feds01:
This PR brings speedups of exhaustiveness from 3x-5x by the following changes:

This PR switches the internal exhaustiveness stores to just use `Vec`s over the `Store` interface. This avoids the locking mechanism for stores which never need to be locked. 

Additionally, this also switches `Fields` to use a `ThinVec` over a `Vec` which leads to better performance and less memory use overall.

Finally, we don't need to store the `bias` on an `IntRange` directly, only when we need to convert back into normal values, so remove it from the `IntRange` struct.

In addition to performance improvements, the analysis phase now records the time taken for exhaustiveness to complete its operations, and reports them through the standard compiler metrics plumbing.

### DIFF
--- a/compiler/hash-parser/src/lib.rs
+++ b/compiler/hash-parser/src/lib.rs
@@ -7,7 +7,7 @@ mod import_resolver;
 pub mod parser;
 mod source;
 
-use std::{env, ops::AddAssign, time::Duration};
+use std::env;
 
 use hash_ast::{
     ast::{self, LocalSpanMap, SpanMap},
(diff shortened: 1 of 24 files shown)


compiler/hash-parser/src/lib.rs
```diff
@@ -291,6 +260,13 @@ fn parse_source(source: ParseSource, sender: Sender<ParserAction>) {
         return;
     }
 
+    // if !source.id().is_prelude() {
+    //     SourceMapUtils::set_module_source(id, contents.clone());
+    //     for token in &tokens {
+    //         println!("{:?}, {}, {:?}", token.kind, token.span,
+    // Span::new(token.span, id).fmt_range());     }
+    // }
+
```
feds01, line 269: remove

kontheocharis:
What is ever the point of using `Vec` when `ThinVec` exists..?

kontheocharis (approved):

feds01:
> What is ever the point of using `Vec` when `ThinVec` exists..?

I'm not too sure, it's superior for memory use, but could it potentially have performance implications?

PR #992, opened 2023-09-23
implement repeat expressions

feds01:
- tir+lower: remember whether a numeric constructor is a platform dependent integer ornot
- ast: introduce repeat expression
- parser: implement parsing repeat expressions
- analysis: implement type checking for repeat expressions
- lower: implement lowering for repeat expressions
- codegen+llvm: implement code generation for repeat expressions
- tests: move entry point tests into semantics
- tests: support running and testing executables
- tests: capture and compare `stdout` and `stderr` for executables that are invoked by tests


compiler/hash-lower/src/lower_ty.rs
```diff
@@ -347,7 +347,7 @@ impl<'ir> BuilderCtx<'ir> {
 
         // We created our own cache entry, so we don't need to update the
         // cache.
-        (reserved_ty, true)
+        (reserved_ty, /* should_cache: */ true)
```
kontheocharis, line 350: Maybe this is a good indicator to use a named struct?

kontheocharis (approved):
Nice one, TC implementation looks good.

compiler/hash-lower/src/lower_ty.rs
kontheocharis: or `ShouldCache::{Yes, No}`

compiler/hash-lower/src/lower_ty.rs
```diff
@@ -35,11 +35,19 @@ use hash_utils::{index_vec::index_vec, itertools::Itertools};
 
 use crate::ctx::BuilderCtx;
 
+/// A flag specifying whether a type lowering from TIR types to IR types
```
kontheocharis, line 38: Stonks

kontheocharis (approved):

PR #994, opened 2023-09-26
codegen: implement string comparisons (Part 1)

### LINKED ISSUE #983: codegen: Implement string comparisons
Currently, this code crashes the compiler code generation:
```rs
main := () => {
    t := "hello"

    match t {
        "hello" => println("hello"),
        "world" => println("world"),
        _ => println("other")
    }
}
```

This crashes on code generation for scalar pairs, i.e. static strings.

feds01:
Partially tackles #983


compiler/hash-codegen/src/lower/rvalue.rs
```diff
@@ -568,6 +571,60 @@ impl<'a, 'b, Builder: BlockBuilderMethods<'a, 'b>> FnBuilder<'a, 'b, Builder> {
         OperandValue::Pair(value, overflow)
     }
 
+    /// Emit code for a [ir::BinOp] which is being applied to two
+    /// [OperandValue::Pair]s. This is usually for `SizedPointer`s, and or
+    /// strings (which is the same representation).
+    fn codegen_pair_binop(
```
kontheocharis, line 577: So these are pointer comparsions right? I am guessing it won't work if the two strings have different allocations but contain the same data? Or am i wrong?

feds01: No you are right, this is just a pointer comparison. The full version would need to emit some additional code at lowering to essentially imitate a `==` operator. However, yes this code isn’t tripped yet because all of the strings are in the same module (thus having the same allocations). I can implement the full version here too

kontheocharis: I think it makes sense to do it in another PR

feds01: Alright, there will be a Part 2 to this PR!

kontheocharis (approved):

PR #999, opened 2023-09-28
lang: implement byte literals `b'a'`

feds01:
- language: add byte literals
- lexer: emit error for empty character literals


compiler/hash-semantics/src/passes/resolution/pats.rs
```diff
@@ -273,6 +273,10 @@ impl<E: SemanticEnv> ResolutionPass<'_, E> {
                 Pat::Lit(LitPat(Node::create_at(Lit::Int((*int_lit).into()), origin))),
                 origin,
             ),
+            ast::Lit::Byte(byte_lit) => Node::create_at(
+                Pat::Lit(LitPat(Node::create_at(Lit::Int((*byte_lit).into()), origin))),
```
kontheocharis, line 277: I wonder how we can remember that this was a ASCII literal rather than an integer in TIR...

kontheocharis (approved):

compiler/hash-semantics/src/passes/resolution/pats.rs
kontheocharis: Perhaps for nicer formatting

feds01: Yeah this would be nice, we can add a flag to int lit, and then use the node origin to just print the raw literal?

kontheocharis: Maybe make an issue for this?

feds01: https://github.com/hash-org/hashc/issues/1001

PR #1004, opened 2023-09-29
lexer: better handling of character and byte literals

### LINKED ISSUE #1003: reporting: wrap newlines correctly in reporting view
Currently, if an error message within span labels contains newline (`\n`) character, the reporting doesn't appropriately wrap them 
around and just prints the label as a raw. For example:
```rs
error: non-ascii character in byte literal
  --> ~/examples/main.hash:78:3
77 |   b'👍',
78 |   b'😘'
   |     ^ must be ASCII
this multibyte character does not fit into a single byte
```
The second line `"this multibyte ..."` is put onto the next line without properly incorporating it into the reporting view.  

This should be fixed by checking if the span label has multiple lines of content, and appropriately formatting it to be aligned with 
the rest of the text.

feds01:
This PR deals with several issues arising from Unicode handling in the compiler. Firstly, the compiler would completely bork when it encountered combining marks. Column offsets and reporting rendering were completely incorrect, and the lexer error message was somewhat confusing. So, this PR does the following:

- lexer: emit error for non-ASCII byte literals
- source: correctly compute column offsets (accounting for Unicode)
- reporting: account for combining marks when drawing span annotations
- lexer: improve errors for combining marks in byte and character literals
- fix #1003

### DIFF
--- a/compiler/hash-reporting/src/render.rs
+++ b/compiler/hash-reporting/src/render.rs
@@ -173,6 +173,56 @@ impl ReportCodeBlock {
         (end - start).saturating_sub(combining_marks)
     }
 
+    /// Function which is responsible for rendering the span note which is
+    /// present at the end of the span view. This will print the initial
+    /// prefix to the span view, usually the highlighted part of the span,
+    /// and then the message.
+    ///
+    /// This function will automatically format the message to correctly align
+    /// all of the lines of the span message with the correct indentation.
+    fn render_span_label(
+        &self,
+        f: &mut fmt::Formatter,
+        offset: usize,
+        initial_prefix: String,
+        longest_indent_width: usize,
+        kind: ReportKind,
+    ) -> fmt::Result {
+        let mut write_line = |line: &str| {
+            writeln!(
+                f,
+                "{} {} {}",
+                " ".repeat(longest_indent_width),
+                highlight(Colour::Blue, "|"),
+                highlight(kind.as_colour(), line)
+            )
+        };
+
+        let line_count = self.code_message.lines().count();
+
+        // If the number of lines is zero, then we still want to print
+        // the initial prefix.
+        if line_count == 0 {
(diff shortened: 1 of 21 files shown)


compiler/hash-reporting/src/render.rs
```diff
@@ -111,6 +111,118 @@ impl ReportCodeBlock {
...
+
+            // If it is a combining mark, its not going to be rendered within the
+            // span, so we need to subtract it from the total width.
+            if unicode_normalization::char::is_combining_mark(ch) {
+                combining_marks += 1;
+            }
+        }
+
+        (end - start).saturating_sub(combining_marks)
+    }
```
kontheocharis, line 174: Interesting approach. Instead of the `unicode_normalization` crate, `unicode_segmentation` could be used which offers a `.graphemes()` iterator so that this function could be replaced by `.graphemes().count()`. Not necessary though, just a potential simplification.

compiler/hash-reporting/src/report.rs
```diff
@@ -151,6 +151,13 @@ pub macro help {
     }
 }
 
+/// Create a `note` note with the given message.
+pub macro note {
```
kontheocharis, line 155: Why are we using first-class macros in this crate?

kontheocharis (approved):

compiler/hash-reporting/src/render.rs
feds01: There is also the `unicode-width` crate, which should compute the true column width of a given Unicode string. Perhaps it is worth leaving a todo/investigate note?

compiler/hash-reporting/src/report.rs
feds01: I guess it just started as a trend, and now it continues

kontheocharis: I guess so

kontheocharis (approved):

PR #1008, opened 2023-10-03
tc: Add some documentation to module items (Part 1)

kontheocharis:
This PR adds some documentation to the typechecking. Specifically it tackles the `lib`, `env`, `traits`, `tc` and `utils` modules. Next up, I will add docs to `operations` and `errors`

This also renames some items to a better suited name, and removes unused code.

### DIFF
--- a/compiler/hash-typecheck/src/options/normalisation.rs
+++ b/compiler/hash-typecheck/src/options/normalisation.rs
@@ -10,9 +10,11 @@ use crate::errors::TcError;
 /// A signal which can be emitted during normalisation.
 #[derive(Debug, Clone, From)]
 pub enum NormaliseSignal {
-    /// A `LoopControlTerm::Break` was encountered in a loop.
+    /// A [`LoopControlTerm::Break`](hash_tir::tir::LoopControlTerm::Break) was
+    /// encountered in a loop.
     Break,
-    /// A `LoopControlTerm::Continue` was encountered in a loop.
+    /// A [`LoopControlTerm::Continue`](`hash_tir::tir::LoopControlTerm::Continue`) was encountered in a
+    /// loop.
     Continue,
     /// A `ReturnTerm` was encountered in a function, with the given return
     /// value.
@@ -179,13 +181,14 @@ impl NormalisationState {
 }
 
 /// The mode in which to normalise terms.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
 pub enum NormalisationMode {
     /// Normalise the term as much as possible.
     Full,
-    /// Normalise the term to a single step (default).
+    /// Normalise the term to a single step.
     ///
     /// This will not execute any impure code.
+    #[default]
     Weak,
 }
 
@@ -204,6 +207,6 @@ impl Default for NormalisationOptions {
 
 impl NormalisationOptions {
     pub fn new() -> Self {
-        Self { mode: LightState::new(NormalisationMode::Weak) }
+        Self { mode: LightState::new(NormalisationMode::default()) }
     }
 }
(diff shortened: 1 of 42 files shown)


compiler/hash-typecheck/src/options/normalisation.rs
```diff
@@ -8,13 +10,36 @@ use crate::errors::TcError;
 /// A signal which can be emitted during normalisation.
 #[derive(Debug, Clone, From)]
 pub enum NormaliseSignal {
+    /// A `LoopControlTerm::Break` was encountered in a loop.
```
feds01, line 13:
```suggestion
    /// A [`LoopControlTerm::Break`] was encountered in a loop.
```

compiler/hash-typecheck/src/options/normalisation.rs
```diff
@@ -8,13 +10,36 @@ use crate::errors::TcError;
...
     Break,
+    /// A `LoopControlTerm::Continue` was encountered in a loop.
```
feds01, line 15:
```suggestion
    /// A [`LoopControlTerm::Continue`] was encountered in a loop.
```

compiler/hash-typecheck/src/options/normalisation.rs
```diff
@@ -8,13 +10,36 @@ use crate::errors::TcError;
...
     Continue,
+    /// A `ReturnTerm` was encountered in a function, with the given return
+    /// value.
     Return(TermId),
+    /// An typechecking error occurred during normalisation.
     Err(TcError),
 }
 
 /// The result of a normalisation operation.
```
feds01, line 24: It would be nice to perhaps just make a type to represent all these outcomes rather than a newtype wrapping it. But I guess its not really feasible since you can't use `?` on self defined types right?

compiler/hash-typecheck/src/options/normalisation.rs
```diff
@@ -152,7 +183,7 @@ impl NormalisationState {
 pub enum NormalisationMode {
     /// Normalise the term as much as possible.
     Full,
-    /// Normalise the term to a single step.
+    /// Normalise the term to a single step (default).
```
feds01, line 186: perhaps, this can be self documents with just:
```suggestion
    /// Normalise the term to a single step
    #[default]
```

compiler/hash-typecheck/src/traits.rs
```diff
@@ -1,19 +1,47 @@
...
+//! node.
 use std::ops::ControlFlow;
 
 use crate::{env::HasTcEnv, errors::TcResult, options::normalisation::NormaliseResult};
 
+/// Main trait for typechecking on TIR atoms.
+///
+/// `X` is the TIR atom type.
 pub trait Operations<X>: HasTcEnv {
-    type TyNode;
```
feds01, line 6: What if this trait was called `OperationsOn<Atom>` ?

compiler/hash-typecheck/src/options/normalisation.rs
kontheocharis: This is why i wrote `@@Improvement: maybe we can make this into a custom enum and implement `std::ops::Try` for it?`

compiler/hash-typecheck/src/traits.rs
kontheocharis: Why need `On`?

kontheocharis: I don't want to use `Atom` because that is a defined type

compiler/hash-typecheck/src/options/normalisation.rs
feds01: Ah right, I didn't quite see it

compiler/hash-typecheck/src/traits.rs
feds01: I see, alright

feds01 (approved):

PR #1010, opened 2023-10-13
Update LLVM installation instructions

kontheocharis:
The previous LLVM installation instructions did not work (at least on macOS due to a homebrew issue relating to GitHub authorisation during installation). It is easier to request people to install the specific LLVM version from the releases page, and set the appropriate environment variables for `inkwell` to find it.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,6 +15,8 @@ specific version that `hashc` uses. We want to move away from this and bundle
 LLVM with `hashc` as a submodule, but for now these are the steps you need to
 follow.
 
+## Linux, macOS
+
 You can download this version of LLVM from
 [here](https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.6) for
 your specific OS. Additionally, you need to install `zstd`  (which can be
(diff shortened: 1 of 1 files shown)


feds01:
I think this is fine, but the windows section should be kept because the process will be fundamentally different on Windows, and it works the last time I tested it. Albeit, we should add a similar section to how to install it directly on Windows (but much harder because of Microsoft dumbness)

kontheocharis:
Is the version `15.0.0` correct? I thought we need `15.0.6`.

feds01:
It is currently version `15.0.6`, although nothing should really break between `15.0.0-15.0.6`, the 'problematic' version is `15.0.7`. 

The only "real" blocker to upgrading the LLVM version to something newer like `15.0.7` or `16.x` or even `17/18.x` is that the current CI runner doesn't support these versions. We could either wait for the versions to be added to the runner, or we could fork the runner and just add them ourselves?

INSTALL.md
```diff
@@ -2,52 +2,38 @@
...
-**Note**: If you are unsure of where LLVM has been installed, you can query 
-the installation path using `brew --prefix llvm@15`.
-
-
-### Windows
-
-To install on Windows, the simplest way to install it is using [Chocolatey](https://chocolatey.org/):
-
-```pwsh
-choco install llvm --version 15.0.0
```
feds01, line 29: Don't remove this section since the build process will be slightly different, and adding things to the environment path is also different. It should indeed however be 15.0.6 here.

kontheocharis: Yes I was just about to add it back, I was just wondering about the versions

feds01 (approved):

PR #1012, opened 2023-10-16
ir: Make projections local to a `Body`

### LINKED ISSUE #996: ir+lower: consider localising `Projections` to each body
## Description

Currently, projections are interned in the global IR projection store. However, there is no need for the items to be all in a global store since the projections have no relation to one another beyond the scope of a `Body`. A potential change could be introduced to make all local projections be stored on the `Body`.

## Why do this?

In the future, we want to parallelise IR building and optimising as much as possible. The thing that hurts the most is the contention on locking (even if we are just reading it, there is still overhead). In order to completely avoid locking when reading projections, we could just store them local to a body, and read them with the owning `Body` in scope.

## Required changes
- Change the IR to store local projections on the `Body`
- Change lowering to use and create projections on the built body. 
- Change the IR visitors to access projections from the owning `Body`

feds01:
This PR makes all of the projections that are applied to a body local to the `Body` rather than being externally stored in a static store. The implementation isn't as clean as I'd hoped it to be (especially in the visitor parts).

## Changes

- ir: localise `Projections` to each `Body`
- ir: cleanup `IrVisitorMut` with `VisitorCtx`
- ir: cleanup `ModiyfyingIrVisitor` with `IrVisitorCtxMut`

## Todo

- [x] Merge #1009 before reviewing this PR

closes #996

### DIFF
--- a/compiler/hash-storage/src/store/sequence/local.rs
+++ b/compiler/hash-storage/src/store/sequence/local.rs
@@ -1,0 +1,71 @@
+use std::marker::PhantomData;
+
+use super::SequenceStoreKey;
+
+/// An implementation of [SequenceStore] which doesn't require being shared
+/// across thread boundaries.
+pub struct LocalSequenceStore<Key, V> {
+    data: Vec<V>,
+    _phantom: PhantomData<Key>,
+}
+
+impl<Key, V> Default for LocalSequenceStore<Key, V> {
+    fn default() -> Self {
+        Self { data: Vec::new(), _phantom: PhantomData }
+    }
+}
+
+impl<Key: SequenceStoreKey, V: Clone> LocalSequenceStore<Key, V> {
+    pub fn new() -> Self {
+        Self { data: Vec::new(), _phantom: PhantomData }
+    }
+
+    /// Create a sequence of values inside the store from a slice, returning its
+    /// key.
+    pub fn create_from_slice(&mut self, values: &[V]) -> Key {
+        let starting_index = self.data.len();
+        self.data.extend_from_slice(values);
+        unsafe { Key::from_raw_parts(starting_index, values.len()) }
+    }
+
+    /// Create an empty sequence of values inside the store, returning its key.
+    pub fn create_empty(&self) -> Key {
+        unsafe { Key::from_raw_parts(self.data.len(), 0) }
+    }
+
+    /// Create a sequence of values inside the store from an iterator-like
+    /// object, returning its key.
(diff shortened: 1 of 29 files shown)


compiler/hash-storage/src/store/sequence/local.rs
```diff
@@ -0,0 +1,71 @@
+use std::marker::PhantomData;
+
+use super::SequenceStoreKey;
+
+/// An implementation of [SequenceStore] which doesn't require being shared
+/// across thread boundaries.
+pub struct LocalSequenceStore<Key, V> {
```
kontheocharis, line 7: How is this different from `IndexVec`?

feds01: It can use the `SequenceStoreKey` API out of the box

kontheocharis: Okay, I see. Perhaps we should slowly start using this over `IndexVec` then

compiler/hash-storage/src/store/sequence/local.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// Get the value sequence for the given key as an owned vector.
+    pub fn get_vec(&self, key: Key) -> Vec<V> {
+        let (index, len) = key.to_index_and_len();
+        self.data[index..index + len].to_vec()
+    }
+
+    /// Borrow a collection of items.
+    pub fn borrow(&self, key: Key) -> &[V] {
+        let (index, len) = key.to_index_and_len();
+        &self.data[index..index + len]
```
kontheocharis, line 57: Can do `unchecked` here too

compiler/hash-storage/src/store/sequence/local.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /// Borrow a mutable collection of items.
+    pub fn borrow_mut(&mut self, key: Key) -> &mut [V] {
+        let (index, len) = key.to_index_and_len();
+        &mut self.data[index..index + len]
```
kontheocharis, line 63: And here

compiler/hash-storage/src/store/sequence/local.rs
feds01: Definitely in areas that lend them selves to using sequence stores

kontheocharis:
LGTM after above

kontheocharis (approved):

PR #1032, opened 2023-10-20
storage: shrink direct sequence element ids from `16bytes` to `12bytes`

feds01:
Any thoughts on how to remove a bunch of type casts everywhere?

### DIFF
--- a/compiler/hash-exhaustiveness/src/deconstruct.rs
+++ b/compiler/hash-exhaustiveness/src/deconstruct.rs
@@ -214,7 +214,7 @@ impl<E: ExhaustivenessEnv> fmt::Debug for ExhaustivenessFmtCtx<'_, Deconstructed
                         // currently active.
                         if let DeconstructedCtor::Variant(index) = ctor {
                             let ctors = data_def.borrow().ctors.assert_defined();
-                            let ctor_name = CtorDefId(ctors.elements(), *index).borrow().name;
+                            let ctor_name = CtorDefId::new(ctors.elements(), *index).borrow().name;
                             write!(f, "::{ctor_name}")?;
                         }
                     }
--- a/compiler/hash-lower/src/build/into.rs
+++ b/compiler/hash-lower/src/build/into.rs
@@ -60,7 +60,7 @@ impl<'tcx> BodyBuilder<'tcx> {
                     .map(|element| {
                         let name = match element.target {
                             ParamIndex::Name(name) => name,
-                            ParamIndex::Position(pos) => pos.into(),
+                            ParamIndex::Position(pos) => Identifier::num(pos),
                         };
 
                         (name, element.value)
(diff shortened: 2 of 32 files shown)


kontheocharis:
We should never be able to provide an index explicitly; I believe doing so is unsafe behaviour. We need to find a way to mark it unsafe and change the relevant code that explicitly constructs direct sequence element IDs so that it doesn't handle indices directly.

feds01 (commented):
I guess this needs to be decided within the specification, what is the maximum index that can be specified. `usize` seems like the logical choice because it then depends on the platform, however it is always unrealistic (if not impossible to actually create a structure that ever reaches any where near such an index). 

My proposal would be that it is decided by this algorithm:
```rs
min(machine_integer_size(), u32)
```

feds01 (commented):
The identifier conversion here is different for `usize` and for `u32`.

This is because internally an `Identifier` is a `u32`, perhaps we should have an explicit function to convert from u32 into a Identifier equivalent?

feds01:
> We should never be able to provide an index explicitly; I believe doing so is unsafe behaviour. We need to find a way to mark it unsafe and change the relevant code that explicitly constructs direct sequence element IDs so that it doesn't handle indices directly.

They have to be constructed somewhere, and sometimes its just mapping things from one type to another. I agree in general that it should be avoided, but without more annoyances. Any suggestions?

compiler/hash-lower/src/build/into.rs
```diff
@@ -477,7 +477,7 @@ impl<'tcx> BodyBuilder<'tcx> {
             .map(|arg| {
                 let name = match arg.target {
                     ParamIndex::Name(name) => name,
-                    ParamIndex::Position(pos) => pos.into(),
+                    ParamIndex::Position(pos) => (pos as usize).into(),
```
feds01, line 480: same problem here with `Identifier` conversions

compiler/hash-exhaustiveness/src/deconstruct.rs
```diff
@@ -214,7 +214,7 @@ impl<E: ExhaustivenessEnv> fmt::Debug for ExhaustivenessFmtCtx<'_, Deconstructed
                         // currently active.
                         if let DeconstructedCtor::Variant(index) = ctor {
                             let ctors = data_def.borrow().ctors.assert_defined();
-                            let ctor_name = CtorDefId(ctors.elements(), *index).borrow().name;
+                            let ctor_name = CtorDefId::new(ctors.elements(), *index).borrow().name;
```
kontheocharis, line 217: Maybe we can provide a method on sequence stores `from_raw_parts` that actually does the assertion that the index is within bounds?

feds01: We could do a `debug_assert!()` in the `new()` constructor?

kontheocharis (approved):

PR #1033, opened 2023-10-25
Update formatting and fix clippy issues

feds01:


kontheocharis (approved):
Finally they added support for this

PR #1039, opened 2024-01-01
Move `IrTy` to `ReprTy`

feds01:
- layout: add `ReprTy` to replace `IrTy`
- layout: rename `represention_stores` -> `repr_stores`
- ir: use `ReprTy` in places over `IrTy` and remove `IrTy`


compiler/hash-attrs/src/builtin.rs
```diff
@@ -30,7 +30,7 @@ macro_rules! define_attr {
     ($table:expr, $name:ident, { ($($arg:ident : $ty:ident),*), $subject:expr }) => {
         let name: Identifier = stringify!($name).into();
 
-        let params = if ${count(arg)} == 0 {
+        let params = if ${count($arg)} == 0 {
```
kontheocharis, line 33: Should `stringify!(arg)` below also be `$arg`?

compiler/hash-layout/src/ty.rs
```diff
@@ -0,0 +1,1136 @@
...
+        params: ReprTyListId,
+
+        /// The return type of the function.
+        return_ty: ReprTyId,
+    },
+
+    /// A function definition, it has an associated instance which denotes
+    /// information about the function, such as the name, defining module,
+    /// ABI, etc.
+    FnDef {
```
kontheocharis, line 248: Seems really strange to me to have this as part of ReprTy.

compiler/hash-layout/src/ty.rs
```diff
@@ -0,0 +1,1136 @@
...
+    u32: ReprTy::UInt(UIntTy::U32),
+    u64: ReprTy::UInt(UIntTy::U64),
+    u128: ReprTy::UInt(UIntTy::U128),
+
+    // ------------------------------------------
+    // Unit types, and unit ptr types
+    // ------------------------------------------
+    usize: ReprTy::UInt(UIntTy::USize),
+    unit: ReprTy::Adt(AdtId::UNIT),
+
```
kontheocharis, line 581: What is the point of having all these special variants in `ReprTy` if they are created as instances of `ReprTy` here anyway? Why not reduce the variants of `ReprTy` to just N * some number of bytes instead, and these common types can remain as they are now?

compiler/hash-layout/src/ty.rs
kontheocharis: Surely all the `Instance` stuff should be IR-specific?

compiler/hash-layout/src/ty.rs
kontheocharis: Perhaps I am not realising something

compiler/hash-layout/src/ty.rs
feds01: Well yes it ideally should be part of IR, but it still has to be accounted for during layout and other things. It is a special variant specific to IR+ stages

kontheocharis: But is the layout of a FnDef not the same as Fn? Shouldn't IrTy wrap ReprTy and potentially add some more variants

compiler/hash-attrs/src/builtin.rs
feds01: This was just an issue that came up in the new Nightly release which complained about it, I don't think we need to change anything else. From what I understand, it is a metavariable expression issue.

compiler/hash-layout/src/ty.rs
feds01: Well, the `Bytes` idea isn't so bad to add to the variants, however we will still need things like a machine specific size variant.

feds01: Also, `bool` and `u8` aren't gonna be treated in the same manner, and hence we'll need a way to distinguish them.

I think we could for the most part remove the integer types and add the following variant:

```rs

enum ReprTy {
  Bool,

  /// Floats are always specific sizes anyway
  Float(..),


  /// chars can also be put here
  Bytes {
    signed: bool,
    size: usize
  },

  MachineInt { 
    signed: bool
  },

  ..
}
```

So yeah this is an option, it is probably better than the current and allows for more flexibility in the future.

compiler/hash-layout/src/ty.rs
feds01: FnDef is effectively a `unit` type because it doesn't really need any storage (mostly because you can't really "assign" a fn-def as an LValue), hence in the layout computation logic it is treated as a `()`.

compiler/hash-layout/src/ty.rs
feds01: I also think right now is probably not the right time to make this representation change. Perhaps when we finish the Const in TIR refactor and the various changes that you are making, then we can consider a representational change.

kontheocharis: I am curious what the difference in handling between bool and u8 is

compiler/hash-layout/src/ty.rs
kontheocharis: Okay. Do you think in the future it is worth investigating to remove FnDef entirely? It still seems to me that it is not really a "type" in the usual sense, but something else entirely.

kontheocharis (dismissed):

kontheocharis (approved):

PR #1041, opened 2024-01-12
Constant representation in TIR

feds01:
- layout: add `ReprTy` to replace `IrTy`
- layout: rename `represention_stores` -> `repr_stores`
- ir: use `ReprTy` in places over `IrTy` and remove `IrTy`
- chore: fix clippy and macro-metavar compilation issues
- wip
- chore: fix clippy and macro-metavar compilation issues
- target: store `TargetDataLayout` on `Target`
- target: don't use `Option` for `TargetDataLayout`
- no more errors lol


compiler/hash-exhaustiveness/src/range.rs
```diff
@@ -256,19 +255,27 @@ impl SplitIntRange {
...
-    pub fn make_range_from_constant(&self, constant: Constant) -> IntRange {
-        let bias: u128 = self.signed_bias(constant.ty);
+    pub fn make_range_from_constant(&self, constant: Const) -> IntRange {
+        let bias = self.signed_bias(constant.ty());
 
         // read from the constant the actual bits and apply bias
-        let val = constant.data() ^ bias;
+        //
+        // @@Correctness: this isn't quite the right conversion, check rustc!
+        let size = constant.as_scalar().size();
```
feds01, line 264: check

compiler/hash-pipeline/src/fs.rs
```diff
@@ -50,7 +49,7 @@ impl Display for ImportErrorKind {
 #[derive(Debug, Clone)]
 pub struct ImportError {
     /// The path that was attempted to be imported.
-    pub path: InternedStr,
+    pub path: String,
```
feds01, line 52: use `AllocId`?

compiler/hash-attrs/src/ty.rs
```diff
@@ -98,3 +102,82 @@ impl AttrTy {
         self.params.at_index(index).map(|param| param.borrow().ty)
     }
 }
+
+fn repr_ty_from_primitive_ctor(ctor: PrimitiveCtorInfo) -> Option<ReprTyId> {
```
kontheocharis, line 106: These two functions seem a bit out of place here

compiler/hash-const-eval/src/eval.rs
```diff
@@ -212,35 +270,97 @@ impl<'ctx> ConstFolder<'ctx> {
...
+            BinOp::Lt => Const::bool(lhs < rhs),
+            BinOp::LtEq => Const::bool(lhs <= rhs),
+            BinOp::Eq => Const::bool(lhs == rhs),
+            BinOp::Neq => Const::bool(lhs != rhs),
+            BinOp::Add => (lhs + rhs).into(),
+            BinOp::Sub => (lhs - rhs).into(),
+            BinOp::Mul => (lhs * rhs).into(),
+            BinOp::Div => (lhs / rhs).into(),
+            BinOp::Mod => (lhs % rhs).into(),
+            BinOp::Exp => (lhs.powf(rhs)).into(),
```
kontheocharis, line 298: What happens on overflow?

compiler/hash-layout/src/constant.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    /// An allocated constant value, this is used to represent everything
+    /// that is non-scalar, including slices, ADTs and arrays.
+    Alloc {
+        /// The offset into the given allocation that is being used.
+        offset: Size,
+
+        /// The allocation itself.
+        alloc: AllocId,
+    },
```
kontheocharis, line 223: What is the difference between these two?

compiler/hash-pipeline/src/fs.rs
kontheocharis: Would be nice to have a newtype wrapper around AllocId for strings

compiler/hash-utils/src/temp_writer.rs
```diff
@@ -0,0 +1,24 @@
+//! Utility for temporarily writing something to a buffer.
+use std::io;
+
+use derive_more::Constructor;
+
+#[derive(Constructor, Default)]
+pub struct TempWriter(Vec<u8>);
+
+impl TempWriter {
```
kontheocharis, line 9: Why not just use `BufWriter`?

compiler/hash-attrs/src/ty.rs
feds01: I wanted to put them in `tir-utils` but it causes a dependency cycle, aby suggestions where to put it

compiler/hash-const-eval/src/eval.rs
feds01: I think it goes to `inf`?

compiler/hash-utils/src/temp_writer.rs
feds01: Where is that from?

compiler/hash-layout/src/constant.rs
feds01: So the only difference between `Pair` and `Alloc` is essentially that a `Pair` is used for an allocation that requires some metadata (i.e. the length of the allocation). Realistically they could be combined into one, however there is some additional logic in codegen (which involves ABI correctness and optimisations) that I don't know how to separate well without having this variant to separate.

I think leaving a `@@Todo` to figure out how to combine these two neatly is a good shout. 

For now, it should be used for strings and slices. Implementation detail from non-codegen parts will be internal - it shouldn't be concerned whether it is a pair or not.

What do you think of this?

compiler/hash-utils/src/temp_writer.rs
feds01: It feels like the [`BufWritter`](https://doc.rust-lang.org/std/io/struct.BufWriter.html#:~:text=For%20example%2C%20every%20call%20to,same%20file%20or%20network%20socket.) API is a bit more awkward to use than this.

compiler/hash-pipeline/src/fs.rs
kontheocharis: like `struct StringId(AllocId)`

compiler/hash-attrs/src/ty.rs
kontheocharis: what is the dependency cycle

compiler/hash-layout/src/constant.rs
kontheocharis: That sounds good; ideally strings should be able to be implemented in the language so they are all just u8 slices

compiler/hash-utils/src/temp_writer.rs
kontheocharis: Okay fair. Why not just use a String and preallocate capacity?

feds01: From the looks of it, [Write](https://doc.rust-lang.org/std/io/trait.Write.html) is implemented for a few types (one of them being `Vec<u8>`, which means we are limited to a number of in-built implementations (foreign implentations restriction) it means we can't simply just use a `String` as a buffer or we would still have to wrap it in a `TempWriter`.

compiler/hash-layout/src/constant.rs
feds01: Yeah that is indeed an ideal, however the compiler at some stages will still need to treat `SizedPointer` like data specially because it can sometimes affect the ABI/codegen (in order to optimise it out).

compiler/hash-utils/src/temp_writer.rs
feds01: The alternative is to directly bake an `str` type and then represent it as an `Alloc`, i.e. `<size> <ptr_to_another_alloc>` and then the `ptr_to_another_alloc` is the underlying bytes buffer for the string.

compiler/hash-layout/src/constant.rs
feds01: The alternative is to directly bake an `str` type and then represent it as an `Alloc`, i.e. `<size> <ptr_to_another_alloc>` and then the `ptr_to_another_alloc` is the underlying bytes buffer for the string.

compiler/hash-pipeline/src/fs.rs
feds01: Now using `StringId`

compiler/hash-attrs/src/ty.rs
kontheocharis: okay.. still quite annoying, i guess hash-repr has the same issue

compiler/hash-utils/src/temp_writer.rs
kontheocharis: it does not implement Write, but why not use String for the purposes of TempWriter. As I understand it it is only used for strings at the end of the day?

kontheocharis (approved):

compiler/hash-attrs/src/ty.rs
feds01: It is the other way around

compiler/hash-utils/src/temp_writer.rs
feds01: I mean a `String` and `Vec<u8>` are the same under the hood, so I think it doesn't matter too much.

PR #1050, opened 2024-03-01
fix miscompilations on Windows

feds01:


compiler/hash-utils/Cargo.toml
```diff
@@ -24,10 +24,21 @@ itertools = "0.10.5"
...
 thin-vec = "0.2.10"
+
+[target.'cfg(windows)'.dependencies.windows]
+version = "0.52.0"
+features = [
+    "Win32_Foundation",
+    "Win32_Storage_FileSystem",
+    "Win32_System_IO",
+    "Win32_System_ProcessStatus",
+    "Win32_System_Threading",
```
kontheocharis, line 42: wtf

kontheocharis (dismissed):

compiler/hash-utils/Cargo.toml
feds01: Yeh its like Microsoft doing it their way

kontheocharis (approved):

Issue #1038, opened 2023-12-09
Introduce a prototype for compiler messages in a separate crate


kontheocharis:
- Type info about a node
   - Where it is defined
   - What kind of node it is
   - Where are its references
- Autocomplete in some span and given characters
- Running the compiler in structured mode to get diagnostics in a machine-readable format
- Lookup by symbol
- Code tree
- Autofixes ...

PR #1067, opened 2024-05-21
messaging

feds01:
- **setup `hash-messaging` crate for input/output messages**
- **messaging: add basic support for schemas**
- **hashc: use correct `tracy` feature name**
- **driver: implement `--schema` option**
- **driver: implement `--configure` flag for configuring the compiler**

This is a huge step towards #1038


compiler/hash-messaging/src/lib.rs
```diff
@@ -1 +1,48 @@
 //! Defines compiler messages that are passed in and out of the compiler.
+
+pub mod listener;
+pub mod stream;
+
+use hash_pipeline::settings::CompilerSettings;
+use hash_reporting::report::Report;
```
kontheocharis, line 7: We need to figure out how to not expose compiler internals here...

feds01: Yep, we definitely need to decide what to keep internal or external!

kontheocharis (approved):

PR #1070, opened 2024-06-07
json messaging format

feds01:
- **pipeline: move `metrics` implementation out of `hash-driver`**
- **logging: add JSON output format**
- **driver: use new `Metrics` type for profiling**
- **messaging: use `CompilerOutputMessage` for `log::info!`s**
- **driver: add configuration for messaging format**


kontheocharis (approved):
this is cool!

PR #1074, opened 2024-07-14
Various metrics comms fixes

feds01:
- **meta: ensure `zstd` is added to link path**
- **settings: ensure `messaging_format` is propagated from `--configure`**
- **driver: disallow `--configure` unknown fields**
- **settings: default serde serialization/deserialization to use kebab-case for enums**

### DIFF
--- a/compiler/hash-pipeline/src/settings.rs
+++ b/compiler/hash-pipeline/src/settings.rs
@@ -402,7 +402,7 @@ impl Default for CompilerSettings {
     Serialize,
     JsonSchema,
 )]
-#[serde(crate = "self::serde")]
+#[serde(crate = "self::serde", rename_all = "kebab-case")]
 pub enum OptimisationLevel {
     /// Run the compiler using the debug optimisation level. This will
     /// disable most optimisations that the compiler would otherwise do.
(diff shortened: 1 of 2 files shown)


.cargo/config.toml
```diff
@@ -1,6 +1,7 @@
 [build]
 rustdoc = "rustdoc"
 rustdocflags = ["--show-type-layout", "-Z", "unstable-options"]
+rustflags = ["-L", "/opt/homebrew/lib/"]
```
kontheocharis, line 4: Surely this shouldn't exist here?

feds01: Ugh yes it really shouldn’t, but for some reason `ld` on mac is refusing to find `zstd` installed on my machine. This is the only way I found to fix it. This will be ignored on systems that this doesn’t exist since its only meant to “indicate” where to look for this libraries.

kontheocharis: True, it won't really hurt others I guess

feds01: I would love to remove this dumb hack, I will leave a `##Hack`

feds01: This now magically works - I think I had a borked installation locally. Removed!

kontheocharis (approved):

PR #1083, opened 2024-12-08
Fix several parser bugs

feds01:
## Description

This PR changes a few things:

- The parser's handling of singular expressions, now we consider single expressions as those that must be physically connected.

- Avoid emitting redundant parenthesis warnings on deref expressions, as
  parentheses are required to disambiguate what expression is being dereferenced.

- Fix a panic that would occur when trying to parse malformed implicit arguments.

To elaborate on single expression handling,
i.e. ones that don't have any operators between them. Previously, the
parser would eagerly continue parsing a singular expression until it
matched the legal structure of an expression. Whilst this was legal,
and handled accordingly at later phases of compilation, it certainly
created odd (and unexpected) behaviour in the parser.

Now, we consider singular expressions as a collection of tokens that
must be "physically" connected. For example, before hand we consider:

```rs
foo
()

// Or

foo
 ()
```

as a call to the function `foo`, however, now we consider them as
separate expressions. Now, calling foo can only be done with:
```rs
foo()
```

## Commits

- **parser: fix panic when trying to parse malformed implicit arguments**
- **parser: fix improper "redundant parenthesis" warning on deref expressions**
- **Add @@Todo on `Token::pretty_print`**
- **reporting: Improve `note_on_span*` functionality with `SpanGuard`**
- **parser: consider physically separated chunks as separate singular expressions**

### DIFF
--- a/compiler/hash-source/src/location.rs
+++ b/compiler/hash-source/src/location.rs
@@ -85,6 +85,20 @@ impl ByteRange {
     pub fn into_location(self, source_id: SourceId) -> Span {
         Span::new(self, source_id)
     }
+
+    /// Check if one span intersects with another one.
+    ///
+    /// We check whether the `current` token touches the other token, if it
+    /// does, then we consider it to be intersecting.
+    /// ```
+    /// 
+    ///  (current token)
+    ///                 (other token)
+    /// ```
+    #[inline(always)]
+    pub fn is_right_before(&self, other: Self) -> bool {
+        self.start() == (other.end() + 1)
+    }
 }
 
 impl Default for ByteRange {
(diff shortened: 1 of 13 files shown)


compiler/hash-source/src/location.rs
```diff
@@ -85,6 +85,21 @@ impl ByteRange {
     pub fn into_location(self, source_id: SourceId) -> Span {
         Span::new(self, source_id)
     }
+
+    /// Check if one span intersects with another one.
```
kontheocharis, line 89: maybe a better name is `touches` since `intersects` implies they can overlap

compiler/hash-source/src/location.rs
```diff
@@ -85,6 +85,21 @@ impl ByteRange {
...
+    ///
+    /// We check whether the `current` token at least touches
```
kontheocharis, line 91:
```suggestion
    /// We check whether the `current` token touches
```

kontheocharis (approved):

compiler/hash-source/src/location.rs
kontheocharis: Also the ordering matters, maybe a better name is actually `a.is_right_after(b)`?

kontheocharis (approved):

PR #1091, opened 2025-04-21
Implement `size_of` directive

feds01:
This PR implements the `#size_of` directive, which allows the caller to inspect
and use the size of a type within the program.

This means that we can now do things like:

```rust

foo := () => {
    ty_size := #size_of u32

    // `ty_size` is now a constant expression that can be used in the program.
    // `ty_size` has a type of `usize`
    // `ty_size` is a compile-time constant of 4.
}
```

The implementation itself simply captures all of these calls and expands them during TIR to then indicate that they should be in fact expanded at `lowering` or evaluation.

There are several restrictions to where the directive can be applied to:

- `#size_of` can only be called on types & expressions that look like types.


kontheocharis (approved):
nice and clean

